PR #4, opened 2025-03-21
Add Brillig support

benjaminwilson:
Brillig is used by the ACVM for solving for the ACIR witness values (the values that are available in `witness_stack`), and this is its only purpose.  Thus adding support for Noir programs that generate Brillig instructions required only mapping all the ACIR witness values to the R1CS witness values (previously we were mapping over only the ACIR _inputs_, which excluded the intermediate ACIR witnesses introduced by Brillig).

I tested this out with the following program (dissected [here](https://www.notion.so/worldcoin/Conditionals-in-ACIR-Brillig-1a68614bdf8c806f95e5f58d2dfcaf43)):

`main.rs`:
```
fn main(a: Field, b: Field) {
	if a == 1 {
		assert(b == 2);
	}
}
```

`Nargo.toml`:
```
a = "1"
b = "2"
```

### DIFF
--- a/noir-r1cs/src/main.rs
+++ b/noir-r1cs/src/main.rs
@@ -97,7 +97,7 @@ fn noir(args: NoirCmd) -> AnyResult<()> {
     // Create the R1CS relation
     let mut r1cs = R1CS::new();
     r1cs.add_circuit(main);
-    r1cs.pretty_print();
+    print!("{}", r1cs);
 
     info!(
         "R1CS: {} witnesses, {} constraints",
--- a/noir-r1cs/src/main.rs
+++ b/noir-r1cs/src/main.rs
@@ -10,7 +10,7 @@ use {
     argh::FromArgs,
     noirc_artifacts::program::ProgramArtifact,
     rand::Rng,
-    std::{fs::File, iter::zip, path::PathBuf, vec},
+    std::{fs::File, path::PathBuf, vec},
     tracing::{info, level_filters::LevelFilter},
     tracing_subscriber::{self, fmt::format::FmtSpan, EnvFilter},
     utils::{file_io::deserialize_witness_stack, PrintAbi},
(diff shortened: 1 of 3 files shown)


noir-r1cs/src/sparse_matrix.rs
```diff
@@ -43,6 +43,32 @@ impl<F> SparseMatrix<F> {
         assert!(col < self.cols, "column index out of bounds");
         self.entries.insert((row, col), value);
     }
+
+    /// Return a dense representation of the matrix.
+    fn to_dense(&self) -> Vec<Vec<F>>
```
recmo, line 48: `Vec<Vec<_>>` is a bad representation for a dense matrix because it requires `n` allocations and needs to follow a pointer for each access. Instead simply allocate a single `Vec` of size `m * n` and compute the indicies (or use a Matrix abstraction to do this for you).
I don't think we'll use dense representation at all much, but if we do, there are existing libraries that provide these efficient abstractions for you.

noir-r1cs/src/sparse_matrix.rs
```diff
@@ -43,6 +43,32 @@ impl<F> SparseMatrix<F> {
...
+    {
+        let mut result = vec![vec![self.default.clone(); self.cols]; self.rows];
+        for ((i, j), value) in self.entries.iter() {
+            result[*i][*j] = value.clone();
+        }
+        result
+    }
+
+    /// Pretty print a dense representation of the matrix.
+    pub fn pretty_print(&self)
```
recmo: In Rust `pretty_print` is commonly implemented using the `Debug` or `Display` trait. This will make it work better with the rest of the debug tooling (e.g. you would be able to `dbg!` print a matrix).

noir-r1cs/src/main.rs
```diff
@@ -85,65 +87,30 @@ fn noir(args: NoirCmd) -> AnyResult<()> {
```
recmo, line 87: Remember the client needs to produce the witness all by itself, we can not provide it as an external file to read.

But for `v0` I think this is acceptable. Witness generation is a big topic we will address soon. Until then, we can re-use Noir's built in witness generator.

recmo (approved):

noir-r1cs/src/sparse_matrix.rs
benjaminwilson: Sure thing, updated comment to indicate its a helper for the Display implementation.

noir-r1cs/src/main.rs
benjaminwilson: Sounds good, I'll put integrating the ACVM solver into our pipeline on the v0 todo list.

PR #5, opened 2025-03-27
Support for ROM using LogUp

benjaminwilson:
Here is a first draft of implementing support for ROM using logUp.  Also separates the compiler and the solver.  WitnessBuilders (recipes for solving for individual witness values) are used in place of the earlier approach.  Transcript is mocked.

Proper tests are still pending.  But the following works, for example, demonstrating dynamic reads:

```
fn main(mem: [Field; 6], addr0: Field, addr1: Field, addr2: Field) -> pub Field {
	mem[addr0] * mem[addr1] + mem[addr2]
}
```

```
mem = [3, 7, 6, 4, 9, 2]
addr0 = 1
addr1 = 4
addr2 = 0
```

To be done:
+ Support range checks with the same code (should be easy).
+ Tests.
+ Add support for predicates (waiting to hear back from Kev on that).


benjaminwilson:
As suggested offline by @vishady721, it seems to me that the static read support can be removed.  As already observed, most static reads can be replaced with the memory witness it is referencing.  And even in more complex cases (involving writes, which oc can't occur with ROM) where that isn't possible, the Noir compiler chooses to express the static read as a dynamic read i.e. use a witness value instead of a constant.  Here's an example:

```
fn main(mut mem: [Field; 3], addr0: Field, addr1: Field) -> pub Field {
	let z = mem[addr0];
	mem[addr0] = mem[addr1];
	mem[addr1] = z;
	mem[0]
}
```

which produces the ACIR:
```
INIT (id: 0, len: 3)
MEM (id: 0, read at: x3, value: x6)
MEM (id: 0, read at: x4, value: x7)
MEM (id: 0, write x7 at: x3)
MEM (id: 0, write x6 at: x4)
---> EXPR [ (-1, _8) 0 ]
---> MEM (id: 0, read at: x8, value: x9)
EXPR [ (1, _5) (-1, _9) 0 ]
```

recmo:
Heads up, we are going to merge this work into `main` as follows:

* [ ] Resolve conflicts by renaming `noir-r1cs` from this PR to `noir-r1cs-logup`.
* [ ] Merge `main` into this branch to make it up to date and resolve conflicts. (This is standard procedure for any PR)
* [ ] Make sure it doesn't break build and tests on crates currently in `main`. This might require solving dependency version issues (e.g. rand). We should take the easiest solutions, e.g. give `noir-r1cs-logup` non-workspace versions.
* [ ] Merge PR to `main`. No review required.

From here on any further work will be in separate PRs with minimal scope. Further work includes copying over features from `noir-r1cs-logup` to `noir-r1cs`, with the eventual goal of removing `noir-r1cs-logup`.

vishady721:
^ above done in https://github.com/worldfnd/ProveKit/pull/50

PR #8, opened 2025-04-01
Add block-multiplier crate

Quarky93:
This PR introduces a new crate called block-multiplier. The crate provides a primitive for efficient modular multiplication, designed to serve as a low-level building block for higher-level cryptographic operations such as the Number-Theoretic Transform (NTT) and the Skyscraper hash.

This is work done by the working group:
Yuval Domb --- @yuvalingo
Xander van der Goot --- @xrvdg 
Koh Wei Jie --- @weijiekoh
Tony Wu --- @Quarky93 

**Motivation**

The core idea behind block-multiplier is to maximize throughput of modular multiplications by exploiting the parallelism between the scalar integer and floating-point SIMD pipelines available on modern ARM CPUs. Specifically, this design targets:
- 3 modular multiplications per block:
  - 1 via the scalar integer pipeline
  - 2 via the floating-point SIMD pipeline
- On optimized platforms (like the Cortex-A76 in the Raspberry Pi 5), the goal is for the latency of a block (3 multiplies) to match that of a single scalar multiplication.

**Platform Notes**
- Raspberry Pi 5 (Cortex-A76): Block size of 3 seem to be optimal.
- Apple Silicon (M1, M3): Slight tuning differences may be required.

**Algorithms**
The techniques implemented draw from the following works:
- Yuval Domb @yuvalingo --- [multi-precision modular reduction scheme with only n^2 + 1 digit multiplications](https://hackmd.io/@Ingonyama/Barret-Montgomery)
- Niall Emmart @nemmart --- [Modular Exponentiation Using Double Precision Floating Point](https://ieeexplore.ieee.org/document/8464792)

These works were implemented by Xander van der Goot @xrvdg in the [modmulzoo](https://github.com/xrvdg/modmulzoo) repo.

This is an initial PR serves as a starting point for us to work directly on the ProveKit repo and **not yet ready for merge**. There are several pending to-dos:
- Add basic documentation and usage examples
- Resolve performance regression due to changing r=260 to r=256 (affect pipeline interleaving and instruction scheduling)
- Finely integrate addv component in the scalar multiplier path
- Remove dependencies except for criterion

NOTES:
- This crate inlines and unrolls most of the code in an attempt to have the compiler better interleave the output instructions.
- The instruction interleaving is very sensitive, as such we keep all the code in one file, to avoid compiler funny business.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,12 @@ Verify the Noir Proof:
 cargo run --release --bin noir-r1cs verify ./noir-proof-scheme.nps ./noir-proof.np
 ```
 
+Generate inputs for Gnark circuit:
+
+```sh
+cargo run --release --bin noir-r1cs generate-gnark-inputs ./noir-proof-scheme.nps ./noir-proof.np
+```
+
 Recursively verify in a Gnark proof (reads the proof from `../ProveKit/prover/proof`):
 
 ```sh
--- a/noir-r1cs/src/cli/cmd/generate_gnark_inputs.rs
+++ b/noir-r1cs/src/cli/cmd/generate_gnark_inputs.rs
@@ -1,0 +1,63 @@
+use {
+    crate::Command,
+    anyhow::{Context, Result},
+    argh::FromArgs,
+    ark_serialize::CanonicalSerialize,
+    noir_r1cs::{
+        create_io_pattern, read, write_gnark_parameters_to_file, NoirProof, NoirProofScheme,
+    },
+    std::{fs::File, io::Write, path::PathBuf},
+    tracing::{info, instrument},
+};
+
+/// Generate input compatible with gnark.
+#[derive(FromArgs, PartialEq, Debug)]
+#[argh(subcommand, name = "generate-gnark-inputs")]
+pub struct Args {
+    /// path to the compiled Noir program
+    #[argh(positional)]
+    scheme_path: PathBuf,
+
+    /// path to the proof file
+    #[argh(positional)]
+    proof_path: PathBuf,
+}
+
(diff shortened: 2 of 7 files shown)


Quarky93:
Latest commit implements subtract 2P on the output. This ensures that if the input is less than R - 2P, then the result will also be less than R - 2P.

block-multiplier/src/constants.rs
```diff
@@ -0,0 +1,141 @@
...
+    0x060C89CE5C263,
+];
+
+pub const F52_P: [f64; 5] = [
+    0x1F593F0000001_u64 as f64,
+    0x4879B9709143E_u64 as f64,
+    0x181585D2833E8_u64 as f64,
+    0xA029B85045B68_u64 as f64,
+    0x030644E72E131_u64 as f64,
+];
```
xrvdg: This constant is not being used.

block-multiplier/src/rtz.rs
```diff
@@ -0,0 +1,208 @@
...
+/// "round toward zero" and restores the previous mode when dropped.
+///
+/// # Safety
+///
+/// This type is not Send because FPCR is a per-core / per OS-thread register. The PhantomData<*mut ()> ensures this.
+/// Only one instance can exist per thread at a time, enforced by the FPCR_OWNED thread-local.
+///
+/// # Design Notes
+///
+/// This type
```
xrvdg: Leftover documentation

xrvdg (approved):

Quarky93:
Will create a new PR instead

PR #13, opened 2025-04-07
Add serde

recmo:


recmo:
Blocked on https://github.com/WizardOfMenlo/whir/pull/168

PR #18, opened 2025-04-12
Blackbox range function (WIP)

vishady721:


noir-r1cs/src/compiler.rs
```diff
@@ -16,6 +19,10 @@ use {
     },
 };
 
+const NUM_WITNESS_THRESHOLD_FOR_LOOKUP_TABLE: usize = 5;
+const NUM_BITS_THRESHOLD_FOR_DIGITAL_DECOMP: u32 = 32;
```
benjaminwilson: Feels like maybe 32 might be too large for `NUM_BITS_THRESHOLD_FOR_DIGITAL_DECOMP`?

noir-r1cs/src/compiler.rs
```diff
@@ -16,6 +19,10 @@ use {
...
+pub const LOG_BASE_DECOMPOSITION: u32 = 32;
```
benjaminwilson: ditto?

noir-r1cs/src/compiler.rs
```diff
@@ -51,6 +58,12 @@ impl R1CS {
 
         // Read-only memory blocks (used for building the memory lookup constraints at the end)
         let mut memory_blocks: BTreeMap<usize, ReadOnlyMemoryBlock> = BTreeMap::new();
+        // Range blocks to map the number of bits threshold to the vector of values that
+        // are meant to be constrained within that range.
+        let mut range_blocks: BTreeMap<u32, Vec<usize>> = BTreeMap::new();
```
benjaminwilson, line 63: As you mentioned, looks like we need to support also RANGE(T) for T not a power of two, as well, so will need to use T as a key to this map (not logT)

noir-r1cs/src/compiler.rs
```diff
@@ -51,6 +58,12 @@ impl R1CS {
...
+        // Same as above, but for number of bits that are above the [NUM_BITS_THRESHOLD_FOR_DIGITAL_DECOMP].
+        // Separated so that we can separate the witness values into digits to do smaller range checks.
+        let mut range_blocks_outside_threshold: BTreeMap<u32, &Vec<usize>> = BTreeMap::new();
```
benjaminwilson: Nit: we don't actually populate this guy until later on, immediately before we use it - I think even perhaps it's redundant, since could just inline that case with the others in the if block starting line 257

noir-r1cs/src/compiler.rs
```diff
@@ -151,9 +164,36 @@ impl R1CS {
                 }
 
                 // These are calls to built-in functions, for this we need to create.
```
benjaminwilson, line 165: missing some words here somehow :)

noir-r1cs/src/compiler.rs
```diff
@@ -151,9 +164,36 @@ impl R1CS {
...
-                    println!("BlackBoxFuncCall")
-                }
+                Opcode::BlackBoxFuncCall(black_box_func_call) => match black_box_func_call {
+                    BlackBoxFuncCall::RANGE {
+                        input: function_input,
+                    } => {
+                        let input = function_input.input();
+                        let num_bits = function_input.num_bits();
+                        let input_wb = match input {
+                            ConstantOrWitnessEnum::Constant(value) => {
```
benjaminwilson: I doesn't feel like Constant should be possible here, since if it were a constant, it'd be known already by the verifier whether it was in range or not. Should Constant be instead an (already existing) R1CSWitness (so no builder required)?  (bc that is a case that seems strangely absent)

noir-r1cs/src/compiler.rs
```diff
@@ -151,9 +164,36 @@ impl R1CS {
...
+                        };
+                        let input_witness = r1cs.add_witness(input_wb);
+                        // Add the entry into the range blocks
+                        range_blocks
+                            .entry(num_bits)
+                            .or_default()
+                            .push(input_witness);
+                        // Keep track of the different ranges we are checking in order
+                        // to track and update multiplicities.
+                        range_blocks.keys().for_each(|key| {
```
benjaminwilson: Doesn't feel like the right place to do this - currently this is happening for every RANGE opcode instance?

noir-r1cs/src/compiler.rs
```diff
@@ -151,9 +164,36 @@ impl R1CS {
...
+                            r1cs.solver.range_checks.push(*key);
+                        });
+                    }
+                    _ => {
+                        println!("Other black box function: {:?}", black_box_func_call);
```
benjaminwilson, line 188: Let's panic or error?

noir-r1cs/src/compiler.rs
```diff
@@ -357,6 +463,119 @@ impl R1CS {
         );
         inverse
     }
+
+    /// Helper function which computes all the terms of the summation for
+    /// each side (LHS and RHS) of the log-derivative multiset check.
+    ///
+    /// Checks that both sums (LHS and RHS) are equal at the end.
+    fn add_logup_summations(&mut self, num_bits: u32, values_to_lookup: &[usize]) {
+        // Sample the Shwartz-Zippel challenge for the log derivative
```
benjaminwilson: Schwartz

noir-r1cs/src/compiler.rs
```diff
@@ -357,6 +463,119 @@ impl R1CS {
...
+                &[
+                    (FieldElement::one(), current_product_witness),
+                    (FieldElement::from(index).neg(), self.solver.witness_one()),
+                ],
+                &[(FieldElement::one(), next_product_witness)],
+            );
+            current_product_witness = next_product_witness;
+        });
+
+        self.matrices.add_constraint(
```
benjaminwilson, line 611: Don't think this constraint is correct?  Shouldn't it be current_product_witness == 0?

noir-r1cs/src/solver.rs
```diff
@@ -29,7 +30,23 @@ pub enum WitnessBuilder {
...
     /// Fields are (sz_challenge, (index_coeff, index), rs_challenge, value).
-    LogUpDenominator(usize, (FieldElement, usize), usize, usize),
+    IndexedLogUpDenominator(usize, (FieldElement, usize), usize, usize),
+    /// For solving for the denominator of a lookup (non-indexed).
+    /// Field are (sz_challenge, (value_coeff, value)).
+    LogUpDenominator(usize, (FieldElement, usize)),
+    /// Products with linear operations on the witness indices.
+    /// Fields are Product((index, a, b), (index, c, d)) such that
+    /// we wish to compute (ax + b) * (cx + d).
+    ProductLinearOperation(
```
benjaminwilson, line 41: I wonder if witness * (another_witness + const) would be sufficiently general? (it's faster to solve for)

benjaminwilson (commented):
looks great!!  btw if you pull on fiat-shamir there are some tests that would serve as a useful template!

PR #19, opened 2025-04-15
Complex M31 field extension and NTT

weijiekoh:
This draft PR adds a `cm31_ntt` member to ProveKit.

As the readme describes, `cm31_ntt` is an implementation of the complex M31 field extension, as well as the NTT algorithm for polynomials with CM31 coefficients.

There is still work to be done to optimise the NTT implementation, so this PR is still a draft.

### DIFF
--- a/cm31_ntt/src/cm31.rs
+++ b/cm31_ntt/src/cm31.rs
@@ -1,0 +1,519 @@
+/// Complex M31 field arithmetic.
+
+use crate::rm31::{ RF, P };
+use std::ops::{ Add, AddAssign, Sub, SubAssign, Neg, Mul, MulAssign };
+use core::fmt::Display;
+use std::convert::{ From, Into };
+use num_traits::Zero;
+use num_traits::identities::One;
+use num_traits::pow::Pow;
+use rand::distributions::{Distribution, Standard};
+use rand::Rng;
+
+#[derive(Copy, Clone, Debug)]
+pub struct CF {
+    pub(crate) a: RF, // the real part
+    pub(crate) b: RF, // the imaginary part
+}
+
+pub const W_8: CF = CF::new(0x00008000, 0x00008000);
+pub const W_8_NEG_1: CF = CF::new(0x7fff7fff, 0x7fff7fff);
+
+pub const W_4: CF = CF::new(0x00000000, 0x00000001);
+pub const W_4_NEG_1: CF = CF::new(0x7fffffff, 0x7ffffffe);
+
+/// Returns the 2nd to n-th roots of unity (inclusive).
+pub fn gen_roots_of_unity(n: usize) -> Vec<CF> {
+    assert!(n > 1);
+    let w2 = CF::root_of_unity_2();
+    let mut w = w2;
+    let mut res = vec![w2];
+    for _ in 2..n+1 {
+        w = w.try_sqrt().unwrap();
+        res.push(w);
+    }
+    res
+}
+
+impl CF {
+    pub const fn new(real: u32, imag: u32) -> CF {
+        CF { a: RF::new(real), b: RF::new(imag) }
+    }
+
+    pub const fn real(self) -> RF {
+        self.a
+    }
+
+    pub const fn imag(self) -> RF {
+        self.b
+    }
+
+    pub fn mul_by_f(self, f: RF) -> CF {
+        CF { 
+            a: f * self.real(),
+            b: f * self.imag(),
+        }
+    }
+
+    pub fn try_inverse(&self) -> Option<Self> {
+        if self.a.val == 0 && self.b.val == 0 {
+            return None;
+        }
+
+        let a2b2 = (self.a * self.a + self.b * self.b).reduce();
+        if a2b2.is_zero() {
+            return None;
+        }
+
+        let a2b2_inv = a2b2.try_inverse().unwrap().reduce();
+        debug_assert!((a2b2 * a2b2_inv).reduce() == RF::new(1));
+
+        let neg_b = self.b.neg();
+        let a_neg_b = CF { a: self.a, b: neg_b };
+
+        let result = a_neg_b.mul_by_f(a2b2_inv);
+        Some(result)
+    }
+
+    pub fn reduce(self) -> CF {
+        CF { a: self.a.reduce(), b: self.b.reduce() }
+    }
+
+    pub fn root_of_unity_2() -> CF {
+        CF::new(0x7ffffffe, 0)
+    }
+
+    /// Returns the 4th root of unity. Since there are 2 options for this value, select the one you
+    /// want using the input `i`.
+    /// The 4th root of unity is (0, +-1)
+    /// The options denoted by `i` are:
+    /// 0. ( 0,  v)
+    /// 1. ( 0, -v)
+    pub fn root_of_unity_4(i: u32) -> Result<CF, String> {
+        assert!(i < 2);
+        if i == 0 {
+            return Ok(CF::new(0, 1));
+        }
+        if i == 1 {
+            return Ok(CF::new(0, P - 1));
+        }
+        panic!("i must be 0 or 1");
+    }
+
+    /// Returns the 8th root of unity. Since there are 4 options for this value, select the one you
+    /// want using the input `i`.
+    /// The 8th root of unity is (+-2^15, +-2^15)
+    /// Let v = 2^15
+    /// The options denoted by `i` are:
+    /// 0. ( v,  v)
+    /// 1. ( v, -v)
+    /// 2. (-v,  v)
+    /// 3. (-v, -v)
+    pub fn root_of_unity_8(i: u32) -> Result<CF, String> {
+        assert!(i < 4);
+        let v = 2u32.pow(15);
+        let neg_v = P - v;
+        // i = 0: (v, v)
+        // i = 1: (v, -v)
+        // i = 2: (-v, v)
+        // i = 3: (-v, -v)
+        if i == 0 {
+            return Ok(CF::new(v, v));
+        }
+        if i == 1 {
+            return Ok(CF::new(v, neg_v));
+        }
+        if i == 2 {
+            return Ok(CF::new(neg_v, v));
+        }
+        if i == 3 {
+            return Ok(CF::new(neg_v, neg_v));
+        }
+        panic!("i must be 0, 1, 2 or 3");
+    }
+
+    #[inline]
+    pub fn mul_neg_1(self) -> Self {
+        let c = self.a.neg();
+        let d = self.b.neg();
+        CF { a: c, b: d }
+    }
+
+    #[inline]
+    pub fn mul_j(self) -> Self {
+        CF { a: self.b.neg(), b: self.a }
+    }
+
+    /// Attempts to compute a square root of a complex element in CF.
+    pub fn try_sqrt(self) -> Option<CF> {
+        if self.is_zero() {
+            return Some(CF::zero());
+        }
+
+        let two = RF::new(2);
+        // 2 is invertible in RF; unwrap is safe since P ≠ 2.
+        let two_inv = two.try_inverse().unwrap();
+        let a = self.a;
+        let b = self.b;
+        // Compute r = sqrt(a^2 + b^2) in RF.
+        let norm = (a * a + b * b).reduce();
+        let r = norm.try_sqrt()?;
+        
+        // Candidate branch 1: try x = sqrt((a + r)/2).
+        let candidate_x2 = ((a + r) * two_inv).reduce();
+        if let Some(x) = candidate_x2.try_sqrt() {
+            // If x ≠ 0 then we can recover y as b/(2x).
+            if !x.is_zero() {
+                let x_inv = x.try_inverse().unwrap();
+                let y = (b * two_inv * x_inv).reduce();
+                let candidate = CF { a: x, b: y }.reduce();
+                if candidate * candidate == self {
+                    return Some(candidate);
+                }
+            }
+        }
+        
+        // Candidate branch 2: try y = sqrt((r - a)/2).
+        let candidate_y2 = ((r - a) * two_inv).reduce();
+        if let Some(y) = candidate_y2.try_sqrt() {
+            if !y.is_zero() {
+                let y_inv = y.try_inverse().unwrap();
+                let x = (b * two_inv * y_inv).reduce();
+                let candidate = CF { a: x, b: y }.reduce();
+                if candidate * candidate == self {
+                    return Some(candidate);
+                }
+            }
+        }
+
+        None
+    }
+}
+
+impl Zero for CF {
+    #[inline]
+    fn zero() -> CF {
+        CF::new(0, 0)
+    }
+    #[inline]
+    fn is_zero(&self) -> bool {
+        self.a.is_zero() && self.b.is_zero()
+    }
+}
+
+impl One for CF {
+    #[inline]
+    fn one() -> CF {
+        CF::new(1, 0)
+    }
+}
+
+impl Into<CF> for u32 {
+    #[inline]
+    /// Converts a u32 into a CF where the real part is the specified u32, and the imaginary part
+    /// is 0
+    fn into(self) -> CF {
+        CF::new(self, 0)
+    }
+}
+
+impl Into<CF> for (u32, u32) {
+    #[inline]
+    fn into(self) -> CF {
+        CF { a: RF::new(self.0), b: RF::new(self.1) }
+    }
+}
+
+impl From<CF> for (u32, u32) {
+    #[inline]
+    fn from(f: CF) -> (u32, u32) {
+        (f.a.reduce().val as u32, f.b.reduce().val as u32)
+    }
+}
+
+impl Add for CF {
+    type Output = Self;
+
+    #[inline]
+    fn add(self, rhs: Self) -> Self::Output {
+        let a = self.a;
+        let b = self.b;
+        let c = rhs.a;
+        let d = rhs.b;
+
+        CF { a: a + c, b: b + d }
+    }
+}
+
+impl AddAssign for CF {
+    #[inline]
+    fn add_assign(&mut self, rhs: Self) {
+        *self = *self + rhs;
+    }
+}
+
+impl Sub for CF {
+    type Output = Self;
+
+    #[inline]
+    fn sub(self, rhs: Self) -> Self::Output {
+        let a = self.a;
+        let b = self.b;
+        let c = rhs.a;
+        let d = rhs.b;
+
+        CF { a: a - c, b: b - d }
+    }
+}
+
+impl SubAssign for CF {
+    #[inline]
+    fn sub_assign(&mut self, rhs: Self) {
+        *self = *self - rhs;
+    }
+}
+
+impl Mul for CF {
+    type Output = Self;
+
+    #[inline]
+    fn mul(self, rhs: Self) -> Self::Output {
+        // (a, b) * (c, d) = (ac - bd, ad + bc)
+        // This implementation uses Karatsuba:
+        // (ac - bd, (a + b)(c + d) - ac - bd)
+        let a = self.a;
+        let b = self.b;
+        let c = rhs.a;
+        let d = rhs.b;
+
+        let ac = a * c;
+        let bd = b * d;
+        let real = ac - bd;
+        let imag = ((a + b) * (c + d) - ac) - bd;
+
+        CF { a: real, b: imag }
+    }
+}
+
+impl MulAssign for CF {
+    #[inline]
+    fn mul_assign(&mut self, rhs: Self) {
+        *self = *self * rhs;
+    }
+}
+
+impl Neg for CF {
+    type Output = Self;
+
+    #[inline]
+    fn neg(self) -> Self::Output {
+        CF { a: -self.a, b: -self.b }
+    }
+}
+
+impl PartialEq for CF {
+    #[inline]
+    fn eq(&self, other: &Self) -> bool {
+        self.a == other.a && self.b == other.b
+    }
+}
+
+impl Eq for CF {}
+
+impl Display for CF {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{} + {}i", self.a, self.b)
+    }
+}
+
+impl Pow<usize> for CF {
+    type Output = CF;
+
+    #[inline]
+    fn pow(self, exp: usize) -> Self::Output {
+        let mut result = CF::one();
+        let mut base = self;
+        let mut exp = exp;
+        while exp > 0 {
+            if exp % 2 == 1 {
+                result *= base;
+            }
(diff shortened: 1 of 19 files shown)


cm31_ntt/src/cm31.rs
```diff
@@ -0,0 +1,537 @@
...
+    /// 1. ( 0, -v)
+    pub fn root_of_unity_4(i: u32) -> Result<CF, String> {
+        assert!(i < 2);
+        if i == 0 {
+            return Ok(CF::new(0, 1));
+        }
+        if i == 1 {
+            return Ok(CF::new(0, P - 1));
+        }
+        panic!("i must be 0 or 1");
```
recmo: A match statement is more canonically rust.

```rust
    pub fn root_of_unity_4(i: u32) -> Result<CF, String> {
        match i {
          0 => CF::new(0, 1),
          1 => CF::new(0, P - 1)
          _ => panic!("i must be 0 or 1")
      }
    }
```

cm31_ntt/src/cm31.rs
```diff
@@ -0,0 +1,537 @@
...
+        CF::new(0x7ffffffe, 0)
+    }
+
+    /// Returns the 4th root of unity. Since there are 2 options for this value, select the one you
+    /// want using the input `i`.
+    /// The 4th root of unity is (0, +-1)
+    /// The options denoted by `i` are:
+    /// 0. ( 0,  v)
+    /// 1. ( 0, -v)
+    pub fn root_of_unity_4(i: u32) -> Result<CF, String> {
```
recmo: This returns `Result` but handles all errors by panicking.

cm31_ntt/src/cm31.rs
recmo: This panic is unreachable (because of the earlier `assert`). So `unreachable!` is more appropriate (but match statement even better).

cm31_ntt/src/cm31.rs
recmo: I recommend using the `anyhow` crate over plain `String` errors, which is what we do elsewhere in this codebase.

cm31_ntt/src/ntt_utils.rs
```diff
@@ -0,0 +1,260 @@
...
+    let b4 = a4 + a6;
+    let b5 = a4 - a6;
+    let b6 = a5 + a7_j;
+    let b7 = a5 - a7_j;
+
+    // Column 3
+    let b5_j = b5.mul_j();
+    let b7_j = b7.mul_j();
+    let b6_w8 = b6 * W_8;
+    let b7_j_w8 = b7_j * W_8;
```
recmo: I'm surprised we are not exploiting the special structure here. Does the compiler turn this into bitshifts?

cm31_ntt/src/ntt_utils.rs
```diff
@@ -0,0 +1,260 @@
...
+/// Helper function to get the n-th root of unity in CF.
+pub fn get_root_of_unity(n: usize) -> CF {
+    assert!(n.is_power_of_two(), "n must be a power of 2");
+    let roots_of_unity = gen_roots_of_unity((n as f64).log2() as usize);
+    roots_of_unity[roots_of_unity.len() - 1]
+}
+
+/// A radix-8 NTT butterfly.
+#[inline]
+pub fn ntt_block_8(
```
recmo, line 38: @xrvdg We will want to aggressively optimize this function.

cm31_ntt/src/ntt_utils.rs
```diff
@@ -0,0 +1,260 @@
...
+    let n = f.len();
+    let wn = get_root_of_unity(n);
+    let mut res = vec![CF::zero(); n];
+    
+    // Parallelize the outer loop
+    res.par_iter_mut().enumerate().for_each(|(i, res_i)| {
+        // Each thread computes one output element
+        for j in 0..n {
+            *res_i += f[j] * wn.pow(i * j);
+        }
```
recmo, line 124: This is very slow, and makes the tests run slower than they need too. A quick improvement would be to:
```rust
let mut wni = wn.pow(i);
let mut wnij = wnij;
*res_i = f[0];
for j in 1..(n-1) {
  *res_i += f[j] * wnij;
  *wnij *= wni;
}
*res_i += f[n-1] * wnij;
```

cm31_ntt/src/rm31.rs
```diff
@@ -0,0 +1,531 @@
...
+use rand::Rng;
+
+pub const P: u32 = 0x7fffffff;
+pub const P_64: u64 = 0x7fffffff;
+pub const P3: u64 = 0x17ffffffd;
+pub const MASK: u64 = 0xffffffff;
+
+// The redundant form of the M31 field. It consists of 31 lower bits (x_l) and the rest are higher
+// bits (x_h).
+// x = x_h * 2^32 + x_l
```
recmo: Should be `x_h * 2^31 + x_l` ?

cm31_ntt/src/rm31.rs
```diff
@@ -0,0 +1,531 @@
...
+
+    #[inline]
+    pub fn reduce(self) -> RF {
+        RF { val: reduce(self.val) }
+    }
+
+    #[inline]
+    fn square(&self) -> Self {
+        // From https://github.com/Plonky3/Plonky3/blob/main/field/src/field.rs
+        self.clone() * self.clone()
```
recmo: `RF` implements copy, so we don't need to `.clone()` it:

```rust
fn square(self) -> Self { self * self }
```

cm31_ntt/src/ntt.rs
```diff
@@ -0,0 +1,1233 @@
...
+}
+
+/// Performs a radix-8 NTT on the polynomial f.
+/// This is unoptimised as it does not use any precomputed twiddles.
+/// It is also slower than ntt_r8_hybrid_p() as it allocates Vecs at each level of recursion, and doing so at
+/// the lower levels (to handle 8^1 to 8^5) incurs significant memory allocation overhead.
+/// @param f The coefficients of the polynomial to be transformed.
+/// @param w The n-th root of unity, where n is the length of f.
+/// @return The transformed polynomial in evaluation form.
+pub fn ntt_r8_vec(f: &Vec<CF>, w: CF) -> Vec<CF> {
```
recmo: Generally you never want to pass `&Vec<_>` as an argument and instead pass a slice `&[_]`.

recmo (approved):
LGTM, but could use some help making it more canonically Rust (which also helps performance).
@Dzejkop Can you help here?

cm31_ntt/src/rm31.rs
```diff
@@ -0,0 +1,531 @@
...
+            
+            e_y = (e_y + e_x) % P;
+
+            assert_eq!(reduced.val, e_y);
+        }
+    }
+
+    #[test]
+    fn test_sqrt() {
+        let valid_test_cases = [
```
Dzejkop: Consider a crate like https://crates.io/crates/test-case. The code would look like

```rust
use test_case::test_case;

#[test_case(0)]
#[test_case(1)]
#[test_case(16)]
#[test_case(3 => panics)]
fn test_sqrt(t: u32) {
    let t = RF::new(t);
    let s = t.try_sqrt().unwrap();
    let s2 = s.square().reduce();
    assert_eq!(s2, t);
}
```

the benefit here is that `test-case` generates a unique test per annotation.  So if the case for `t == 1` fails, you'll still see output for `t == 2`

cm31_ntt/src/rm31.rs
```diff
@@ -0,0 +1,531 @@
...
+            if candidate < threshold {
+                return RF { val: candidate % P };
+            }
+        }
+    }
+}
+
+impl Display for RF {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "{}", self.val)
```
Dzejkop: This ignores the formatting params, which makes a difference e.g. when printing with padding. Consider changing to

```suggestion
        self.val.fmt(f)
```

cm31_ntt/src/ntt_utils.rs
```diff
@@ -0,0 +1,260 @@
+use crate::cm31::{ CF, gen_roots_of_unity, W_8 };
```
Dzejkop: formatting?

weijiekoh:
Thanks for the feedback and suggestions! I'll incorporate them into a new commit.

cm31_ntt/src/ntt.rs
```diff
@@ -0,0 +1,1233 @@
...
+
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+pub struct PrecomputedTwiddles {
+    pub small: Vec<CF>,  // for the base NTT of size NTT_BLOCK_SIZE_FOR_CACHE
+    pub full: Vec<CF>,   // for the higher levels of recursion in the radix-8 context
+    pub stride: Vec<CF>, // for stride 2 or stride 4
+}
+
+pub fn precompute_twiddles(n: usize) -> Result<PrecomputedTwiddles, String> {
+    assert!(is_power_of(n as u32, 2), "n must be a power of 2");
```
Dzejkop: You can use `eyre::Result<PrecomputedTwiddles>` or `anyhow::Result<PrecomputedTwiddles>` instead of `Result<..., String>`. This will make it easier to nail down the errors as both these types can link to source.

With that, this assert can be replaced with something like `anyhow::ensure!(is_power_of(n as u32, 2), "n must be a power of 2");`

cm31_ntt/src/ntt.rs
```diff
@@ -0,0 +1,1233 @@
...
+        let precomp_s4 = precomp_s4(n);
+        return Ok(PrecomputedTwiddles {
+            small: precomp_small,
+            full: precomp_full,
+            stride: precomp_s4,
+        });
+    }
+    
+
+    Err("n must be a power of 2, or of the form 8^k * 2 or 8^k * 4".to_string())
```
Dzejkop: more on a personal note but if this was a chain of `if .. else if ...` then you could do away with the return statements and it'd be a little more type safe (in the sense that the if blocks don't **have** to return a value).

so, e.g.:
```suggestion
    if is_power_of(n as u32, 8) {
        let (precomp_small, precomp_full) = precomp_for_ntt_r8_hybrid_p(n);
        Ok(PrecomputedTwiddles {
            small: precomp_small,
            full: precomp_full,
            stride: vec![],
        })
    }
    // Case where n is of the form 8^k * 2
    else if is_power_of((n / 2) as u32, 8) {
        let (precomp_small, precomp_full) = precomp_for_ntt_r8_hybrid_p(n / 2);
        let precomp_s2 = precomp_s2(n);
        Ok(PrecomputedTwiddles {
            small: precomp_small,
            full: precomp_full,
            stride: precomp_s2,
        })
    }
    // Case where n is of the form 8^k * 4
    else if is_power_of((n / 4) as u32, 8) {
        let (precomp_small, precomp_full) = precomp_for_ntt_r8_hybrid_p(n / 4);
        let precomp_s4 = precomp_s4(n);
        Ok(PrecomputedTwiddles {
            small: precomp_small,
            full: precomp_full,
            stride: precomp_s4,
        })
    } else {
        Err("n must be a power of 2, or of the form 8^k * 2 or 8^k * 4".to_string())
    }
```

cm31_ntt/src/ntt.rs
```diff
@@ -0,0 +1,1233 @@
...
+/// the lower levels (to handle 8^1 to 8^5) incurs significant memory allocation overhead.
+/// @param f The coefficients of the polynomial to be transformed.
+/// @param w The n-th root of unity, where n is the length of f.
+/// @return The transformed polynomial in evaluation form.
+pub fn ntt_r8_vec(f: &Vec<CF>, w: CF) -> Vec<CF> {
+    let n = f.len();
+    debug_assert!(n >= 8, "n must be at least 8");
+    debug_assert!(is_power_of(n as u32, 8), "n must be a power of 8");
+
+    fn do_ntt(f: &Vec<CF>, w: CF) -> Vec<CF> {
```
Dzejkop: ~why the nested fn? Seems like we could just inline the body here~

cm31_ntt/src/ntt.rs
Dzejkop:
```suggestion
pub fn ntt_r8_vec(f: &[CF], w: CF) -> Vec<CF> {
```

cm31_ntt/src/ntt.rs
```diff
@@ -0,0 +1,1233 @@
...
+        let n = f.len();
+        if n == 1 {
+            return f.clone();
+        }
+
+        let m = n / 8;
+        let w_pow_8 = w.pow(8);
+
+        // Partition f into 8 subarrays.
+        let mut a = vec![vec![CF::zero(); m]; 8];
```
Dzejkop: ~it'd be more performant to "flatten" this into `vec![CF::zero(); m * 8]` and then index into it like `a[i + j * 8]`~

cm31_ntt/src/ntt.rs
```diff
@@ -0,0 +1,1233 @@
...
+}
+
+/// Performs a radix-2 NTT on the polynomial f.
+/// This is unoptimised as it does not use any precomputed twiddles.
+/// Like ntt_r8_vec, it is also slower than ntt_r8_s2_hybrid_p() as it allocates Vecs at each level
+/// of recursion.
+/// @param f The coefficients of the polynomial to be transformed.
+/// @param w The n-th root of unity, where n is the length of f.
+/// @return The transformed polynomial in evaluation form.
+pub fn ntt_radix_2(f: &Vec<CF>, w: CF) -> Vec<CF> {
```
Dzejkop:
```suggestion
pub fn ntt_radix_2(f: &[CF], w: CF) -> Vec<CF> {
```

cm31_ntt/src/ntt.rs
Dzejkop: This could be an array of vecs though
```suggestion
        let mut a = [vec![CF::zero(); m]; 8];
```

cm31_ntt/src/ntt.rs
```diff
@@ -0,0 +1,1233 @@
...
+                for j in 0..8 {
+                    a[j][i] = *f.get_unchecked(i_8 + j);
+                }
+            }
+
+            // Recurse
+            let mut ntt_a = Vec::with_capacity(8);
+            for i in 0..8 {
+                ntt_a.push(do_ntt(&a[i], w_pow_8));
+            }
```
Dzejkop: if this is perf critical we could play around with `MaybeUninit` but otherwise an array of empty vecs could do the trick

```suggestion
            let mut ntt_a = [Vec::new(); 8];
            for i in 0..8 {
                ntt_a[i] = do_ntt(&a[i], w_pow_8);
            }
```

cm31_ntt/src/ntt.rs
```diff
@@ -0,0 +1,1233 @@
...
+/// of recursion.
+/// @param f The coefficients of the polynomial to be transformed.
+/// @param w The n-th root of unity, where n is the length of f.
+/// @return The transformed polynomial in evaluation form.
+pub fn ntt_radix_2(f: &Vec<CF>, w: CF) -> Vec<CF> {
+    let n = f.len();
+    assert!(n >= 2, "n must be at least 2");
+    assert!(n.is_power_of_two(), "n must be a power of 2");
+
+    fn do_ntt(f: &Vec<CF>, w: CF) -> Vec<CF> {
```
Dzejkop:
```suggestion
    fn do_ntt(f: &[CF], w: CF) -> Vec<CF> {
```

cm31_ntt/src/ntt.rs
```diff
@@ -0,0 +1,1233 @@
...
+                let f5 = *ntt_a.get_unchecked(5).get_unchecked(k);
+                let f6 = *ntt_a.get_unchecked(6).get_unchecked(k);
+                let f7 = *ntt_a.get_unchecked(7).get_unchecked(k);
+
+                let butterfly = ntt_block_8(
+                    f0, f1, f2, f3, f4, f5, f6, f7,
+                    wt, wt2, wt3, wt4, wt5, wt6, wt7
+                );
+
+                *res.get_unchecked_mut(k) = butterfly.0;
```
Dzejkop, line 161: k is in range `0..m` where `m = n / 8`, res is `vec![CF::zero(); n];`. If n is smaller than zero, or not divisible by 8 you can run into undefined behaviour here with index overrun.

The asserts at the start only safeguard against it in debug. You need to either replace the debug asserts with regular asserts or make this function fallible.

weijiekoh: Thanks for spotting this!

Added `assert!(n > 0 && n % 8 == 0); ` in commit aaddb593 to prevent this issue.

cm31_ntt/src/ntt.rs
```diff
@@ -0,0 +1,1289 @@
...
+        assert!(n > 0 && n % 8 == 0);
+
+        let m = n / 8;
+        let w_pow_8 = w.pow(8);
+
+        // Partition f into 8 subarrays.
+        let mut a = [(); 8].map(|_| vec![CF::zero(); m]);
+
+        let mut res = vec![CF::zero(); n];
+        unsafe {
```
Dzejkop, line 119: Let's add safety comments to the unsafe blocks as per https://std-dev-guide.rust-lang.org/policy/safety-comments.html#inside-safe-elements

cm31_ntt/src/ntt.rs
```diff
@@ -0,0 +1,1289 @@
...
+/// This is unoptimised as it does not use any precomputed twiddles.
+/// It is also slower than ntt_r8_hybrid_p() as it allocates Vecs at each level of recursion, and doing so at
+/// the lower levels (to handle 8^1 to 8^5) incurs significant memory allocation overhead.
+/// @param f The coefficients of the polynomial to be transformed.
+/// @param w The n-th root of unity, where n is the length of f.
+/// @return The transformed polynomial in evaluation form.
+pub fn ntt_r8_vec(f: &[CF], w: CF) -> Vec<CF> {
+    let n = f.len();
+    debug_assert!(n >= 8, "n must be at least 8");
+    debug_assert!(is_power_of(n as u32, 8), "n must be a power of 8");
```
Dzejkop, line 100:
```suggestion
    assert!(n >= 8, "n must be at least 8");
    assert!(is_power_of(n as u32, 8), "n must be a power of 8");
```

I'm always in favour of Result though, but I'm not sure how perf critical this is

cm31_ntt/src/ntt.rs
Dzejkop: I'd just like to suggest making this and other functions fallible if we can handle the perf impact of error handling - we sometimes deal with unexpected panics deep in various crypto crates and it's always a pain to track down.

Also I think if you changed the `debug_asserts` to regular asserts this would've also resolved the undefined behaviour danger - why a new assert?

cm31_ntt/src/ntt.rs
```diff
@@ -0,0 +1,1289 @@
...
+}
+
+pub fn ntt(
+    f: &[CF],
+    precomp: &PrecomputedTwiddles,
+) -> Result<Vec<CF>> {
+    let n = f.len();
+    let mut scratch = vec![CF::zero(); NTT_BLOCK_SIZE_FOR_CACHE];
+    debug_assert!(is_power_of(n as u32, 2), "n must be a power of 2");
+    debug_assert!(n >= 8, "n must be at least 8");
```
Dzejkop: what do you think about replacing these with `anyhow::ensure` since the function is already fallible?

Dzejkop: and building on top of that, make other `ntt_...` functions fallible and bubble up the errors

cm31_ntt/src/ntt.rs
Dzejkop: This relates to https://github.com/worldfnd/ProveKit/pull/19/files#r2072454803

I also noticed that some functions have `debug_asserts` and some have `asserts` in the beginning - let's make a this consistent across all functions

cm31_ntt/src/ntt.rs
```diff
@@ -0,0 +1,1289 @@
...
+
+        res
+    }
+
+    do_ntt(f, w)
+}
+
+/// Precompute twiddles for use in the ntt_r8_vec_p() function.
+pub fn precomp_for_ntt_r8_vec_p(n: usize, w: CF) -> Vec<CF> {
+    let radix: usize = 8;
```
Dzejkop:
```suggestion
    const RADIX: usize = 8;
```

cm31_ntt/src/ntt.rs
```diff
@@ -0,0 +1,1289 @@
...
+    assert!(n.is_power_of_two(), "n must be a power of 2");
+    assert!(radix.is_power_of_two(), "radix must be a power of 2");
```
Dzejkop: this is always true since `radix` is constant
```suggestion
```

cm31_ntt/src/ntt.rs
```diff
@@ -0,0 +1,1289 @@
...
+        current_n /= radix;
+        current_w = next_w;
+    }
+    
+    twiddles
+}
+
+pub fn ntt_r8_vec_p(
+    f: &Vec<CF>,
+    twiddles: &Vec<CF>,
```
Dzejkop:
```suggestion
    f: &[CF],
    twiddles: &[CF],
```

cm31_ntt/src/ntt.rs
```diff
@@ -0,0 +1,1289 @@
...
+        for _ in 0..depth {
+            off += 1 + 7 * len;
+            len /= 8;
+        }
+        off
+    }
+
+    fn do_ntt_precomputed(
+        f: &Vec<CF>,
+        twiddles: &Vec<CF>,
```
Dzejkop:
```suggestion
        f: &[CF],
        twiddles: &[CF],
```

cm31_ntt/src/ntt_utils.rs
weijiekoh: Resolved in commit 496dbce0, though benchmarks don't show a significant improvement.

cm31_ntt/src/ntt.rs
weijiekoh: I love that suggestion. It would make things more consistent. I've made these changes in commit 5b2a8d9c

cm31_ntt/src/ntt.rs
weijiekoh: Resolved by using `ensure!` consistently.

cm31_ntt/src/rm31.rs
weijiekoh: `test_case` is really helpful! I've incorporated it into the tests.

cm31_ntt/src/ntt.rs
weijiekoh: Resolved in commit 22a45aa.

weijiekoh:
Thank you @Dzejkop ! Your suggestions are super helpful. Things look much cleaner and consistent now!

recmo:
Thank you @Dzejkop  and @weijiekoh. This looks good to merge (we can always follow up with further PRs).

PR #20, opened 2025-04-15
Calculate from transcript

batmendbar:
This draft PR removes r, alpha, last_sumcheck_val, and statement from WhirR1CSProof. Now, verifier computes them on their own. whir_query_answer_sums remains in the WhirR1CSProof as the verifier cannot compute them from the transcript on their own.

### DIFF
--- a/noir-r1cs/src/whir_r1cs.rs
+++ b/noir-r1cs/src/whir_r1cs.rs
@@ -33,7 +33,7 @@ use {
             domainsep::WhirDomainSeparator,
             parameters::WhirConfig as GenericWhirConfig,
             prover::Prover,
-            statement::{Statement, StatementVerifier as GenericStatementVerifier, Weights},
+            statement::{Statement, StatementVerifier as GenericStatementVerifier, VerifierWeights, Weights},
             verifier::Verifier,
             WhirProof as GenericWhirProof,
         },
@@ -76,8 +76,6 @@ pub struct WhirR1CSProof {
     // TODO: Derive from transcript
     #[serde(with = "serde_ark")]
     whir_query_answer_sums: [FieldElement; 3],
-    // TODO: Derive from scheme and transcript
-    // statement: Statement<FieldElement>,
 }
 
 impl WhirR1CSScheme {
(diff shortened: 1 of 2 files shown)


noir-r1cs/src/whir_r1cs.rs
```diff
@@ -333,28 +319,31 @@ pub fn run_whir_pcs_prover(
...
     let _ = arthur.fill_challenge_scalars(&mut r);
 
     let mut saved_val_for_sumcheck_equality_assertion = FieldElement::zero();
 
-    for _i in 0..m_0 {
+    let mut alpha = vec![FieldElement::zero(); m_0];
+
+    for i in 0..m_0 {
         let mut hhat_i = vec![FieldElement::zero(); 4];
         let mut alpha_i = vec![FieldElement::zero(); 1];
```
recmo: `Vec` should be avoided as much as possible for performance reasons. These can be plain arrays.

recmo (changes requested):
LGTM, should be a big improvement. @Dzejkop can you help out with perf and style review?

Dzejkop (approved):

noir-r1cs/src/whir_r1cs.rs
```diff
@@ -333,28 +319,31 @@ pub fn run_whir_pcs_prover(
...
-    (proof, merlin, sums, statement)
+    (proof, merlin, sums)
 }
 
 #[instrument(skip_all)]
 pub fn run_sumcheck_verifier(
     arthur: &mut VerifierState<SkyscraperSponge, FieldElement>,
     m_0: usize,
-) -> Result<()> {
+) -> Result<(Vec<FieldElement>, Vec<FieldElement>, FieldElement)> {
```
Dzejkop: Consider a struct for the return type

noir-r1cs/src/whir_r1cs.rs
batmendbar: Did you mean these should be plain arrays?
```
let mut hhat_i = vec![FieldElement::zero(); 4];
let mut alpha_i = vec![FieldElement::zero(); 1];
```

recmo (approved):

PR #37, opened 2025-05-06
Refactor to generalize over modes and archs

recmo:


fp-rounding/src/arch/aarch64.rs
```diff
@@ -0,0 +1,59 @@
+#![cfg(target_arch = "aarch64")]
+//! Floating point rounding mode control for aarch64 architecture.
+//!
+//! See <https://developer.arm.com/documentation/ddi0595/2021-06/AArch64-Registers/FPCR--Floating-point-Control-Register>
+use {crate::RoundingMode, core::arch::asm};
+
+/// Layout of the floating point control register (FPCR).
+const SHIFT: u32 = 21;
```
xrvdg: FPCR is numbered as [63:0] and the rounding mode [23:22]. The shift need to be 22 instead of 21, because 0b11 is in [1:0].

PR #45, opened 2025-05-12
Add codspeed

recmo:


recmo:
I added a Raspberry Pi 5 as an action runner for benchmarks.

recmo:
Not all features are supported: https://github.com/CodSpeedHQ/codspeed-rust/issues/96

PR #46, opened 2025-05-12
Benchmark prover

recmo:


recmo:
The PoW bits benchmark is too noisy. Maybe replace with deterministic challenges.

PR #122, opened 2025-06-03
Execute the program to generate the witness

Dzejkop:


noir-r1cs/benches/bench.rs
```diff
@@ -22,29 +22,15 @@ fn prove_poseidon_1000(bencher: Bencher) {
...
+
+    let program_path = crate_dir.join("target/basic.json");
+    let witness_path = crate_dir.join("Prover.toml");
+
+    let witness_map = execute_program_witness(program_path, witness_path).unwrap();
 
-    let witness_file_path = &PathBuf::from("../noir-examples/poseidon-rounds/target/basic.gz");
-    let mut witness_stack = deserialize_witness_stack(witness_file_path).unwrap();
-    let witness_map: WitnessMap<NoirFieldElement> = witness_stack.pop().unwrap().witness;
     bencher.bench(|| black_box(&scheme).prove(black_box(&witness_map)));
```
recmo: A prove should start with a `.nps` and `.toml` file and end with a `.np` file. Nothing more nothing less. Here file reading and witness computation are excluded. It's fine to have such a benchmark (doing only partial proof) but we also need a benchmark that does an end-to-end proof.

noir-r1cs/benches/bench.rs
```diff
@@ -22,29 +22,15 @@ fn prove_poseidon_1000(bencher: Bencher) {
...
-        .expect("Running nargo execute");
-    if !status.success() {
-        panic!("Failed to run nargo execute");
-    }
+    compile_workspace(crate_dir).expect("Compiling workspace");
+
+    let program_path = crate_dir.join("target/basic.json");
+    let witness_path = crate_dir.join("Prover.toml");
+
+    let witness_map = execute_program_witness(program_path, witness_path).unwrap();
```
recmo: Witness computation is part of the proof computation.

noir-r1cs/src/cli/cmd/prove.rs
```diff
@@ -12,7 +12,11 @@ use {
 #[derive(FromArgs, PartialEq, Eq, Debug)]
 #[argh(subcommand, name = "prove")]
 pub struct Args {
-    /// path to the compiled Noir program
+    /// path to the compile Noir program
+    #[argh(positional)]
+    program_path: PathBuf,
```
recmo: The inputs to the prover are a `.nps` file and a `.toml` file. Nothing more, nothing less. Please add things to NoirProofScheme as necessary.

noir-r1cs/src/cli/cmd/prove.rs
```diff
@@ -47,10 +51,7 @@ impl Command for Args {
         let (constraints, witnesses) = scheme.size();
         info!(constraints, witnesses, "Read Noir proof scheme");
 
-        // Read the input toml
-        let witness_file_path = &PathBuf::from(&self.witness_path);
-        let mut witness_stack = deserialize_witness_stack(witness_file_path).unwrap();
-        let witness_map: WitnessMap<NoirFieldElement> = witness_stack.pop().unwrap().witness;
+        let witness_map = execute_program_witness(&self.program_path, &self.witness_path)?;
```
recmo: Witness computation should be part of `shceme.prove`, which should take only public and private inputs (e.g. as a toml string).

noir-tools/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+}
+
+pub fn execute_program_witness(
+    program_path: impl AsRef<Path>,
+    witness_path: impl AsRef<Path>,
+) -> Result<WitnessMap<FieldElement>> {
+    let program_file = std::fs::File::open(program_path.as_ref())?;
+    let program: ProgramArtifact =
+        serde_json::from_reader(program_file).context("Reading program")?;
+    let program: CompiledProgram = program.into();
```
recmo: We can add `CompiledProgram` to the `NoirProofScheme` struct or one of it sub-structs (I think there's one dedicated to witness generation)

noir-tools/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+    let compile_options = CompileOptions::default();
+
+    compile_workspace_full(&workspace, &compile_options, None)?;
+
+    Ok(workspace)
+}
+
+pub fn execute_program_witness(
+    program_path: impl AsRef<Path>,
+    witness_path: impl AsRef<Path>,
```
recmo: We should not handle files inside the library.

noir-tools/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+        root_path:    None,
+        package_name: None,
+    }
+    .build();
+
+    let mut exec_results = execute(
+        &program,
+        &solver,
+        &mut foreign_call_executor,
+        witness_path.as_ref(),
```
recmo: We should inline `execute` to remove the use of a files.

Cargo.toml
```diff
@@ -113,6 +114,13 @@ noirc_abi = { git = "https://github.com/noir-lang/noir", rev = "nightly-2025-05-
 noirc_artifacts = { git = "https://github.com/noir-lang/noir", rev = "nightly-2025-05-28" }
 noirc_printable_type = { git = "https://github.com/noir-lang/noir", rev = "nightly-2025-05-28" }
 bn254_blackbox_solver = { git = "https://github.com/noir-lang/noir", rev = "nightly-2025-05-28" }
+nargo = { git = "https://github.com/noir-lang/noir", rev = "nightly-2025-05-28" }
+nargo_toml = { git = "https://github.com/noir-lang/noir", rev = "nightly-2025-05-28" }
+noirc_driver = { git = "https://github.com/noir-lang/noir", rev = "nightly-2025-05-28" }
+noirc_errors = { git = "https://github.com/noir-lang/noir", rev = "nightly-2025-05-28" }
+fm = { git = "https://github.com/noir-lang/noir", rev = "nightly-2025-05-28" }
+nargo_cli = { git = "https://github.com/noir-lang/noir", rev = "nightly-2025-05-28" }
```
recmo, line 117: I wasn't expecting the command line specific stuff to be linked in. Fine for now, but we probably want to trim this before audit.

noir-r1cs/src/cli/cmd/prove.rs
```diff
@@ -47,14 +46,12 @@ impl Command for Args {
         let (constraints, witnesses) = scheme.size();
         info!(constraints, witnesses, "Read Noir proof scheme");
 
-        // Read the input toml
-        let witness_file_path = &PathBuf::from(&self.witness_path);
-        let mut witness_stack = deserialize_witness_stack(witness_file_path).unwrap();
-        let witness_map: WitnessMap<NoirFieldElement> = witness_stack.pop().unwrap().witness;
+        // Read witness data
+        let input_map = scheme.read_witness(&self.witness_path)?;
```
recmo: `witness_path` should be named back to `input_path`.

noir-r1cs/src/cli/cmd/prove.rs
```diff
@@ -47,14 +46,12 @@ impl Command for Args {
         let (constraints, witnesses) = scheme.size();
         info!(constraints, witnesses, "Read Noir proof scheme");
 
-        // Read the input toml
-        let witness_file_path = &PathBuf::from(&self.witness_path);
-        let mut witness_stack = deserialize_witness_stack(witness_file_path).unwrap();
-        let witness_map: WitnessMap<NoirFieldElement> = witness_stack.pop().unwrap().witness;
+        // Read witness data
```
recmo: `// Read the input toml `

noir-r1cs/src/noir_proof_scheme.rs
```diff
@@ -16,8 +20,9 @@ use {
 };
 
 /// A scheme for proving a Noir program.
-#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct NoirProofScheme {
+    pub program:           Program<NoirFieldElement>,
     pub r1cs:              R1CS,
     pub witness_builders:  Vec<WitnessBuilder>,
     pub witness_generator: NoirWitnessGenerator,
```
recmo, line 28: `witness_builders`, `witness_generator` and now also a `generate_witness` fn entirely separate from these? This is needs to be cleaned up soon, but maybe not in this PR.

noir-r1cs/src/noir_proof_scheme.rs
```diff
@@ -94,18 +100,54 @@ impl NoirProofScheme {
...
 
+    pub fn read_witness(&self, prover_toml: impl AsRef<Path>) -> Result<InputMap> {
+        let (input_map, _expected_return) =
+            read_inputs_from_file(prover_toml.as_ref(), self.witness_generator.abi())?;
+
+        Ok(input_map)
+    }
+
+    pub fn generate_witness(&self, input_map: &InputMap) -> Result<WitnessMap<NoirFieldElement>> {
+        let span = span!(Level::INFO, "generate_witness").entered();
```
recmo: Question, why use this over the `instrument` macro?

noir-r1cs/src/noir_proof_scheme.rs
```diff
@@ -94,18 +100,54 @@ impl NoirProofScheme {
...
+        let initial_witness = self.witness_generator.abi().encode(input_map, None)?;
+
+        let mut witness_stack = nargo::ops::execute_program(
+            &self.program,
+            initial_witness,
+            &solver,
+            &mut foreign_call_executor,
+        )?;
+
+        drop(span);
```
recmo: Redundant, it would drop around here anyway as the scope ends (unless you specifically want to exclude these last couple operations like `.pop()`)

noir-r1cs/src/noir_witness.rs
```diff
@@ -22,6 +22,7 @@ pub struct NoirWitnessGenerator {
     // Note: Abi uses an [internally tagged] enum format in Serde, which is not compatible
     // with some schemaless formats like Postcard.
     // [internally-tagged]: https://serde.rs/enum-representations.html
+    // TODO: serializes the ABI as a json string. Something like CBOR might be better.
```
recmo, line 25: Agree. Now I remember it all again. This is a pretty ugly hack. Fortunately the Abi object is small so it has little impact.

noir-r1cs/tests/compiler.rs
```diff
@@ -17,32 +13,10 @@ struct NargoTomlPackage {
...
-    if !status.success() {
-        panic!("Failed to execute the test case");
-    }
-}
-
 fn test_compiler(test_case_path: impl AsRef<Path>) {
     let test_case_path = test_case_path.as_ref();
 
-    run_nargo(test_case_path);
+    compile_workspace(test_case_path).expect("Compiling workspace");
```
recmo, line 19: To clarify: It's of course fine to run `nargo` to do some stuff for end-to-end tests. And might even be preferred if it reduces dependency bloat or complexity.

noir-r1cs/Cargo.toml
```diff
@@ -40,8 +44,13 @@ acir.workspace = true
 acir_field.workspace = true
 noirc_artifacts.workspace = true
 noirc_abi.workspace = true
+nargo.workspace = true
+nargo_toml.workspace = true
+noirc_driver.workspace = true
+nargo_cli.workspace = true
```
recmo: I don't think all of these are used?

recmo (approved):

Cargo.toml
Dzejkop: This is mostly for the `compile_workspace` funciton so we can avoid calling the external `nargo` binary in tests & benchmarks

noir-r1cs/src/noir_proof_scheme.rs
Dzejkop: I think I just copied it other, from what was here before, fixing now

noir-r1cs/src/noir_witness.rs
Dzejkop: Yeah, I'm leaving the TODO - I don't think it's worth sinking time into it now

noir-r1cs/tests/compiler.rs
Dzejkop: We can make the nargo cli deps dev-dependencies. This approach is a little easier to setup for tests (no external dependency) and ensures that there's no version mismatch between the `nargo` binary and the dependencies.

It also gives us a little more flexibility as it's e.g. possible to make tests independent by overriding the workspace compilation target directory

noir-r1cs/Cargo.toml
Dzejkop: yeah, cleaning them up

noir-r1cs/tests/compiler.rs
recmo: Fully agree with this! Dev dependency are excluded from bloat/audit concerns.

recmo (approved):

PR #123, opened 2025-06-03
Build on GH runners for RP5 benchmarks

Dzejkop:


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,63 @@
...
+  pull_request:
+
+env:
+  DIVAN_MAX_TIME: 30
+  CODSPEED_RUNNER_MODE: walltime
+
+jobs:
+  build:
+    name: Build benchmark binaries
+    runs-on: ubuntu-22.04-arm
```
recmo, line 17: It's better to be explicit about what the compilation target is. If you compile this machine, which processors will the resulting executable run on? Which fancy instructions does it use? It may crash on a different processor.
Minor to this is optimization: microarchitecture (pipelines etc.) differ between processors and this can be quite a substantial difference in performance. We already see this in a ~20% performance difference between two skyscraper algorithms, with the fastest one being different on an M3 than on the RPi5!

recmo (approved):

.github/workflows/benchmark.yml
recmo: The `simd_mul` regression in codspeed is actually exactly what you'd expect due to microachitectural mismatch. That could be the explanation.

But it's also not a huge concern: a big part of the reason to use assembly is to avoid depending on the compiler in this regard.

Dzejkop: I aggree, but `cargo codspeed` doesn't support setting the target triple right now. I'll add an issue for that, but let's leave it they way it is for now

Dzejkop: And it's a bit more involved to build the benchmarks without it (wrangling the benchmark binaries + handling the walltime/instrumentation modes)

PR #125, opened 2025-06-16
Parallel sparse multiplication

xrvdg:


noir-r1cs/src/sparse_matrix.rs
```diff
@@ -175,10 +202,44 @@ impl Mul<HydratedSparseMatrix<'_>> for &[FieldElement] {
             rhs.matrix.num_rows,
             "Vector length does not match number of rows."
         );
+
+        let intermediate_multiplication =
+            LockFreeArray::new(vec![(0, FieldElement::zero()); rhs.matrix.num_entries()]);
```
recmo: It's unfortunate that it requires allocating a temporary vector larger than the output.

noir-r1cs/src/sparse_matrix.rs
```diff
@@ -175,10 +202,44 @@ impl Mul<HydratedSparseMatrix<'_>> for &[FieldElement] {
...
+        // Use a lock-free array to prevent constant resizing when collecting the
+        // iterator as the size is not known to Rayon. Collecting without a
+        // preallocating the intermediate vector is >15% slower
+        // Other options that have been explored
+        // - An IndexedParallelIterator on the values of the sparse matrix also wasn't
+        //   an option as it requires random access which we can't provide as we
+        //   wouldn't know the row a value belongs to. That's why the rows drive the
+        //   iterator below.
+        // - Acquiring a mutex per column in the result was too expensive (even with
+        //   parking_lot)
```
recmo: What I had in mind was splitting the work over the mutable output vector, and creating a `iter_col` method on the matrix. (Which is substantially more complicated than `iter_row`, but should be doable).
Can we explore this option?

noir-r1cs/src/sparse_matrix.rs
```diff
@@ -156,16 +159,40 @@ impl Mul<&[FieldElement]> for HydratedSparseMatrix<'_> {
...
-            result[i] += value * rhs[j];
-        }
+
+        (0..self.matrix.num_rows)
+            .into_par_iter()
+            .map(|i| {
+                self.iter_row(i)
+                    .fold(FieldElement::zero(), |sum, (j, value)| sum + value * rhs[j])
+            })
+            .collect_into_vec(&mut result);
```
recmo, line 170: `collect_into_vec` clears the vec, so there's no need to initialize it. You can just allocate with `Vec::with_capacity`

recmo: Alternatively, it should also be possible to do this with something like:

`result.par_iter_mut().enumerate().for_each(|(i, x)| *x = self.iter_row(i)... );`

noir-r1cs/src/sparse_matrix.rs
```diff
@@ -175,10 +182,28 @@ impl Mul<HydratedSparseMatrix<'_>> for &[FieldElement] {
...
             "Vector length does not match number of rows."
         );
+
+        let num_threads = rayon::current_num_threads();
         let mut result = vec![FieldElement::zero(); rhs.matrix.num_cols];
-        for ((i, j), value) in rhs.iter() {
-            result[j] += value * self[i];
-        }
+
+        let chunk_size = result.len().div_ceil(num_threads);
```
recmo, line 189: If the perf delta is small (e.g. <5%) I think I prefer a solution that let's rayon decide the chunk size. I don't like hard dependencies on the number of available threads. The number of threads does not say much, as we might be doing other work in parallel.

Instead we should pick the workload size such that it amortizes the overhead, while still allowing parallelization for large problem sizes. To approximate this I like to pick 'whatever subproblem fits in L1 cache' as problem size. And this in turn is approximated with the `workload_size::<F>(result.len())` function.

noir-r1cs/src/sparse_matrix.rs
```diff
@@ -175,10 +182,28 @@ impl Mul<HydratedSparseMatrix<'_>> for &[FieldElement] {
...
+
+        // In microbenchmarks par_iter_mut.chunks outperforms par_chunks_mut slightly.
```
recmo, line 191: If it is a small difference this might be a compiler quirk. I'd stick with `par_chunks_mut` for simplicity and maintainability.

If `.par_iter_mut().chunks(..)` is faster than this is basically a bug somewhere as the former provides the libraries/compiler with strictly more information. This bug is likely to be fixed at some point.

noir-r1cs/src/sparse_matrix.rs
```diff
@@ -175,10 +182,28 @@ impl Mul<HydratedSparseMatrix<'_>> for &[FieldElement] {
...
+        result
+            .par_iter_mut()
+            .chunks(chunk_size)
+            .enumerate()
```
recmo, line 195: If you do `.enumerate().par_chunks_mut(size)` you should already get the `base` value, which simplifies the code.

noir-r1cs/src/sparse_matrix.rs
```diff
@@ -175,10 +182,28 @@ impl Mul<HydratedSparseMatrix<'_>> for &[FieldElement] {
...
+            .for_each(|(chunk_number, mut chunk)| {
+                let base = chunk_number * chunk_size;
+                let col_range = base..base + chunk_size;
```
recmo, line 198: `let col_range = base..base + chunk.len()`. 

Otherwise `col_range` will be out of bounds when `chunk_size` doesn't divide `result.len()`. You are protected here from this bug because `col` will only be in-range, but better to do it right.

recmo (approved):
LGTM after comments addressed.

noir-r1cs/src/sparse_matrix.rs
recmo: Ah, that won't work because you can't chunk the resulting tuples. Nevermind!

noir-r1cs/src/sparse_matrix.rs
recmo: `let base = chunk_number * chunk_size` should still be correct as only the last chunk is not exactly `chunk_size` (but please confirm with rayon docs, and maybe leave a comment explaining correctness.)

PR #132, opened 2025-07-21
Use zk-whir variant instead of the leaky one

veljkovranic:
Opening this instead of this work:
https://github.com/worldfnd/ProveKit/pull/32

It is relying on an unmerged PR in WHIR: https://github.com/WizardOfMenlo/whir/pull/188

The goal of this PR is to rely on WHIRs batching option to perform polynomial commitment in the ZK fashion.

The crux of it is, instead of performing WHIR rounds on the witness polynomial, you mask and blind its values with a huge random polynomial and do the batch commit on [masked, blinding] pair of polynomials.

Theoretical explanation can be found here: https://github.com/worldfnd/ProveKit/blob/main/sage/fri-and-friends/Zero%20Knowledge%20for%20WHIR.md


gnark-whir/mt.go
```diff
@@ -185,17 +282,110 @@ func verify_circuit(
...
+		LeafSiblingHashes:          totalLeafSiblingHashes,
+		LeafIndexes:                totalLeafIndexes,
+		ContainerAuthPaths:         containerTotalAuthPath,
+		ContainerLeaves:            containerTotalLeaves,
+		ContainerLeafSiblingHashes: containerTotalLeafSiblingHashes,
+		ContainerLeafIndexes:       containerTotalLeafIndexes,
+	}
+}
+
+// TODO: refactor this function
```
wzmuda, line 304: TODO: refactor this whole project 😅

noir-r1cs/src/whir_r1cs.rs
```diff
@@ -83,27 +86,56 @@ impl WhirR1CSScheme {
...
+        &self,
+        witness_polynomial_evals: EvaluationsList<FieldElement>,
+        merlin: &mut ProverState<SkyscraperSponge, FieldElement>,
+    ) -> (
+        Witness<FieldElement, SkyscraperMerkleConfig>,
+        EvaluationsList<FieldElement>,
+        EvaluationsList<FieldElement>,
+    ) {
+        let mask = generate_mask(witness_polynomial_evals.evals().len());
+        let masked_polynomial = create_masked_polynomial(&witness_polynomial_evals, &mask);
```
batmendbar, line 121: Very minor: IMO, `generate_mask` and `create_masked_polynomia` can be combined to just one function called `mask_polynomial(&whitness_polynomial_evals)`.

noir-r1cs/src/whir_r1cs.rs
```diff
@@ -314,13 +365,55 @@ pub fn run_whir_pcs_prover(
...
     });
+    let prover = Prover(params.clone());
+    prover
+        .prove(&mut merlin, statement, witness)
+        .expect("WHIR prover failed to generate a proof");
+    (merlin, sums)
+}
+
+#[instrument(skip_all)]
+pub fn run_zk_whir_pcs_prover(
```
batmendbar, line 376: This seems to replace `run_whir_pcs_prover`, making it a dead code. Should we remove it?

batmendbar (changes requested):
There are some changes that might conflict with the Spark implementation, but I can handle  them when merging Spark to this. I left some comments on small issues. Otherwise, looks good.

PR #133, opened 2025-07-21
Updating ZK document

yswami-tfh:
Updating ZK document with details about sampling randomness. It's sufficient to sample randomness similar to ZK done in Libra. This should improve both time and memory requirements for adding ZK.


recmo (approved):

wf-orbie:
Superseded by #488, which restores the actual unzipped UTF-8 Markdown document on an in-repository branch.

PR #141, opened 2025-08-14
Remove SPARK and add back direct matrix evaluation

batmendbar:


ashpect:
@veljkovranic Just curious, was there an unresolved blocker that required commenting out the compiler tests to keep CI passing ?

PR #142, opened 2025-08-16
Fix: Update `noir-r1cs` with Skyscraper v2

shreyas-londhe:
This PR:
- Refactors the code by moving `block_multiplier`, `block_multiplier_codegen`, `fp_rounding` and `hla` into `/skyscraper` and moves `/skyscraper` to `/skyscraper/core`.
- Updates `/noir-r1cs/skyscraper` to be compatible with Skyscraper v2.

Impact:
- No behavior change expected; Merkle, sponge, and proofs continue to work via externalized Skyscraper
- Clear separation of concerns: generic performance code in `skyscraper/`, noir-r1cs keeps only protocol glue
- Easier maintenance and reuse of `skyscraper/` components across projects

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -24,31 +24,9 @@ jobs:
           cache-base: main
           cache-target: release
           bins: cargo-codspeed
-      - name: Find packages with benchmarks
-        id: find-benchmarks
-        run: |
-          # Find all Cargo.toml files that contain [[bench]] sections
-          BENCH_PACKAGES=""
-          for toml in $(find . -name "Cargo.toml" -not -path "./target/*"); do
-            if grep -q "^\[\[bench\]\]" "$toml"; then
-              # Extract package name from the Cargo.toml
-              pkg_name=$(grep -m1 "^name = " "$toml" | sed 's/name = "\(.*\)"/\1/')
-              if [ -n "$pkg_name" ]; then
-                BENCH_PACKAGES="$BENCH_PACKAGES -p $pkg_name"
-              fi
-            fi
-          done
-          echo "packages=$BENCH_PACKAGES" >> $GITHUB_OUTPUT
-          echo "Found benchmark packages:$BENCH_PACKAGES"
       - name: Build the benchmark target(s)
         timeout-minutes: 30
-        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/benchmark.yml
```diff
@@ -24,9 +24,31 @@ jobs:
           cache-base: main
           cache-target: release
           bins: cargo-codspeed
+      - name: Find packages with benchmarks
```
Bisht13: Is manually discovery of benches required here? This should already be happening by `cargo codspeed build`

noir-r1cs/src/skyscraper/pow.rs
Bisht13, line 1: This should be in `skyscraper/core/src/pow.rs`

noir-r1cs/src/skyscraper/sponge.rs
```diff
@@ -0,0 +1,60 @@
+use {
+    crate::FieldElement,
+    ark_bn254::Fr,
+    ark_ff::{BigInt, PrimeField},
+    spongefish::duplex_sponge::{DuplexSponge, Permutation},
+    zeroize::Zeroize,
+};
+
+fn to_fr(x: FieldElement) -> Fr {
```
Bisht13, line 9: Fr and FieldElement are same?

Bisht13 (changes requested):

noir-r1cs/src/skyscraper/pow.rs
shreyas-londhe: This implements `PowStrategy` required in Spongefish for the transcript. The reason for having this is `noir-r1cs` is that it is related to the prover and not the Skyscraper hash function so it does not make sense for it be in the `skyscraper/core`.

noir-r1cs/src/skyscraper/sponge.rs
shreyas-londhe: Internally yes, but whir has it's own type named `Field256`, because of which we need this function.

.github/workflows/benchmark.yml
shreyas-londhe: It was not working earlier because the `cm31_ntt` crate was not declared in the root `Cargo.toml`. I reverted this change.

PR #147, opened 2025-08-21
EVM Implementation of Skyscraper

recmo:
@philsippl asked how Skyscraper would perform onchain. Turns out it's about 10x better than Posseidon. This is useful for  Merkle trees that are efficient to handle on-chain and in circuit.


recmo:
@xrvdg You spend some time optimizing Skyscraper. Do you have any ideas we could try here?

skyscraper-evm/src/Skyscraper.sol
```diff
@@ -0,0 +1,219 @@
...
+
+    function sss(
+        uint256 l,
+        uint256 r,
+        uint256 rc_a,
+        uint256 rc_b
+    ) internal pure returns (uint256, uint256) {
+        unchecked {
+            r = rc_a + addmod(mulmod(mulmod(l, l, P), SIGMA_INV, P), r, P);
+            l = rc_b + addmod(mulmod(mulmod(r, r, P), SIGMA_INV, P), l, P);
```
xrvdg, line 146: I think these `addmod` are safe to be replaced by regular adds. After `mulmod` there is `4.29p` space left before a u256 overflows. Afterwards it directly gets fed into the `mulmod` in `sss_reduce_l` which in turn reduces it.

skyscraper-evm/src/Skyscraper.sol
```diff
@@ -0,0 +1,219 @@
...
+        uint256 rc_b
+    ) internal pure returns (uint256, uint256) {
+        unchecked {
+            r = rc_a + addmod(mulmod(mulmod(l, l, P), SIGMA_INV, P), r, P);
+        }
+        l = addmod(
+            rc_b,
+            addmod(mulmod(mulmod(r, r, P), SIGMA_INV, P), l, P),
+            P
+        );
```
xrvdg, line 164: Same here, only the outer `addmod` should be required.

skyscraper-evm/src/Skyscraper.sol
```diff
@@ -0,0 +1,219 @@
...
+
+    function sss_reduce_l(
+        uint256 l,
+        uint256 r,
+        uint256 rc_a,
+        uint256 rc_b
+    ) internal pure returns (uint256, uint256) {
+        unchecked {
+            r = rc_a + addmod(mulmod(mulmod(l, l, P), SIGMA_INV, P), r, P);
+        }
```
xrvdg, line 159: Based on this line it looks like `r` doesn't have to be reduced for `bb` so it should be fine to drop the `addmod` here. Same reasoning as above.

xrvdg (commented):
In the other skyscraper implementation we tuned the output range of the montgomery multiplication (modmul) to only reduce once the addition is done. Looks like we can do the opposite here: modmul does a full reduction and that leaves space within the `u256` for the additions.

recmo:
Closing as pursuing Skyscraper no longer seems useful.

PR #148, opened 2025-08-22
Add SkyscraperV2 Noir implementation

recmo:
Switching to SkyscraperV2 for onchain Merkle trees makes sense (see #147). To realize this, we also need an efficient Noir implementation of SkyscraperV2.

Additionally, we need this for self-recursion.

The current implementation suffers performance in two areas:

* `Field::to_le_bytes` does a bytewise comparisson of the result to make sure it represents the canonical representation of the field element. While this check is necessary, it seems in efficient to do it this way.
* `SBOX[bytes[i] as u32]` introduces bounds checks. These are completely unnecessary as bytes is by construction constraint to 0..255.

### DIFF
--- a/skyscraper-noir-bench/Nargo.toml
+++ b/skyscraper-noir-bench/Nargo.toml
@@ -1,0 +1,7 @@
+[package]
+name = "skyscraper_noir_bench"
+type = "bin"
+authors = [""]
+
+[dependencies]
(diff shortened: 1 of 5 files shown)


recmo:
Closing as pursuing Skyscraper no longer seems useful.

PR #174, opened 2025-09-01
Memory access optimised NTT for BN254

xrvdg:
The main addition in this PR is the BN254 NTT which is 1.5x to 2x times faster and uses half the space for the roots vector compared to the implementation in WHIR for the range 2ˆ20 to 2ˆ24. The performance increase is mainly due not fetching roots from memory during the hot loop. By storing the roots in reverse bit order the access pattern for the roots is linear.

The NTT is split into three phases all using the same general strategy (DIT NR). In the last phase the problem size fits within cache and in a future PR it will switch to an optimised strategy. 

Other additions in this PR:
- simple proptest valuetree and strategy to generate valid NTTs for use in tests. 
- Unoptimised inverse NTT

Follow ups to this PRs will deal with:
- Winograd kernels for cached-sized problems
- Incorporate the optimised block multiplier instead of ark-ff
- further reduce the size of the roots table
- caching engine like in WHIR
- optimise out omega^0 = 1 multiplications

### Measurements 

On macOS the improvement is 2.25x. When looking at the CPU resource graph there seem to be some unnecessary idles. Moving to a custom schedular instead of Rayon might allow for some extra performance improvment. 

On Raspberry Pi there is a 1.5x-2x improvement. Caveat both WHIR and this ran into thermal throttling on my Raspberry Pi as it doesn't have any heatsink or any cooling.


ntt/src/proptest.rs
```diff
@@ -0,0 +1,79 @@
...
+
+impl<T: Strategy> Strategy for NTTStrategy<T> {
+    type Tree = NTTValueTree<T::Tree>;
+
+    type Value = NTT<T::Value>;
+
+    fn new_tree(
+        &self,
+        runner: &mut prop::test_runner::TestRunner,
+    ) -> prop::strategy::NewTree<Self> {
```
recmo: Why do we need a decicated Strategy implementation for this? I would expect something like

```rust
    (0..10)
        .prop_map(|k| 1 << k)
        .prop_flat_map(|len| vec(elem.clone(), len))
```

to suffice.

ntt/src/proptest.rs
```diff
@@ -0,0 +1,79 @@
...
+
+    fn current(&self) -> Self::Value {
+        let vec = self.vec.0[..self.len]
+            .iter()
+            .map(|element| element.current())
+            .collect();
+        NTT::new(vec).unwrap()
+    }
+
+    // Simplifies only the structure and not the values.
```
recmo: It's worthwhile to reduce the values too. In my experience proptests often ends up with some 0/1 vector that triggers the bug, which makes debugging easier.

ntt/src/ntt.rs
```diff
@@ -0,0 +1,274 @@
...
+
+    if n <= 1 {
+        return;
+    }
+
+    // Each unique twiddle factor within a stage is a group.
+    let mut pairs_in_group = n / 2;
+    let mut num_of_groups = 1;
+
+    // For small NTTs go directly to the NTT optimized for cache size
```
recmo: Nitpick: I expect the comment to be written from the perspective of the branch being taken, i.e. "For large NTTs we first do ..."

PR #182, opened 2025-09-10
feat: Parallelise `generate_random_multilinear_polynomial`

Bisht13:


veljkovranic (approved):

provekit/common/src/utils/zk_utils.rs
```diff
@@ -11,12 +14,23 @@ pub fn create_masked_polynomial(
...
-    let mut rng = ark_std::rand::thread_rng();
-    let mut elements = Vec::with_capacity(1 << num_vars);
-
-    for _ in 0..(1 << num_vars) {
-        elements.push(FieldElement::rand(&mut rng));
+    let num_elements = 1 << num_vars;
+    let mut elements = Vec::with_capacity(num_elements);
+    unsafe {
+        elements.set_len(num_elements);
     }
```
xrvdg, line 21: I believe `set_len` needs to be called after initialisation/`for_each`.  
>[Safety](https://doc.rust-lang.org/std/vec/struct.Vec.html#safety-4)
new_len must be less than or equal to [capacity()](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.capacity).
The elements at old_len..new_len must be initialized.

A more idiomatic way to do this is with: [spare_capacity_mut](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.spare_capacity_mut)

PR #185, opened 2025-09-11
Fix: Recursive Verification

shreyas-londhe:
This PR is dependent on https://github.com/WizardOfMenlo/whir/pull/195. It fixes recursive-verification by making the transcript constant length.


veljkovranic (approved):
lgtm, we should check what is happening with e2e though

PR #186, opened 2025-09-11
perf(common): Parallelize polynomial generation via spare_capacity

Bisht13:
Addressing https://github.com/worldfnd/provekit/pull/182#discussion_r2338677588

### DIFF
--- a/provekit/common/src/utils/zk_utils.rs
+++ b/provekit/common/src/utils/zk_utils.rs
@@ -16,21 +16,24 @@ pub fn create_masked_polynomial(
 pub fn generate_random_multilinear_polynomial(num_vars: usize) -> Vec<FieldElement> {
     let num_elements = 1 << num_vars;
     let mut elements = Vec::with_capacity(num_elements);
-    unsafe {
-        elements.set_len(num_elements);
-    }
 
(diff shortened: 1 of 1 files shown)


provekit/common/src/utils/zk_utils.rs
```diff
@@ -16,21 +16,24 @@ pub fn create_masked_polynomial(
...
     const CHUNK_SIZE: usize = 32;
 
-    // Fill the pre-allocated vector in parallel using chunked approach
-    // Each thread gets its own RNG instance and processes a chunk of elements
-    elements.par_chunks_mut(CHUNK_SIZE).for_each(|chunk| {
+    // Get access to the uninitialized memory
+    let spare = elements.spare_capacity_mut();
+
+    // Fill the uninitialized memory in parallel using chunked approach
+    spare.par_chunks_mut(CHUNK_SIZE).for_each(|chunk| {
```
xrvdg, line 27: [`map_init`](https://docs.rs/rayon/latest/rayon/iter/trait.ParallelIterator.html#method.map_init) is Rayon's way of dealing with this pattern. It removes the need for manually defining a chunk size and let Rayon's planner handle it.

PR #196, opened 2025-09-22
feat: Implement layered witness solving with batch inversion optimization

shreyas-londhe:
## Description
Optimizes witness solving by grouping operations into layers and batching field inversions. Reduces N expensive field inversions to 1 inversion + 3N multiplications using [Montgomery's trick](https://medium.com/eryxcoop/montgomerys-trick-for-batch-galois-field-inversion-9b6d0f399da2).

## Implementation
1. **Dependency Analysis**: Build DAG from witness read/write relationships
2. **Layered Scheduling**: Use topological sort to group non-inverse operations into segments
3. **Deferred Inversion**: Collect inverse operations and batch them using Montgomery's trick
4. **Execution**: Alternate between PRE segments (non-inverse ops) and inverse batches

## Key Components
- LayerScheduler: Frontier-based algorithm to compute optimal execution layers
- LayeredWitnessBuilders: Execution plan with PRE segments + inverse batches
- batch_inverse_montgomery: Mathematical optimization for field inversions
- solve_witness_vec: New execution engine that processes layered plans

## Files Changed
- Added `layer_scheduler.rs` - Core scheduling algorithm
- Updated `witness_builder.rs` - Dependency analysis utilities
- Updated `r1cs.rs` - Layered execution engine
- Updated `utils/mod.rs` - Montgomery batch inversion
- Updated `noir_proof_scheme.rs` - Integration into compilation

### DIFF
--- a/provekit/common/src/witness/layer_scheduler.rs
+++ b/provekit/common/src/witness/layer_scheduler.rs
@@ -1,0 +1,388 @@
+use {
+    super::WitnessBuilder, crate::witness::{ConstantOrR1CSWitness, ConstantTerm, ProductLinearTerm, SumTerm, WitnessCoefficient, BINOP_ATOMIC_BITS}, std::{
+        collections::{BTreeMap, BTreeSet, VecDeque},
+        mem,
+    }
+};
+
+/// Layered plan for serial batched-inversion solving.
+///
+/// Semantics:
+/// - `pre_builders` contains only non-`Inverse` builders, concatenated
+///   segment-by-segment.
+/// - For segment i, the PRE slice is `pre_builders[pre_segment_starts[i] ..
+///   pre_segment_starts[i+1] or end]`.
+/// - Immediately after PRE[i], batch-invert all denominators listed in
+///   `inverse_batches[i]`.
+/// - Each `inverse_batches[i]` contains only `WitnessBuilder::Inverse(out,
+///   denom)` elements.
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize, PartialEq)]
+pub struct LayeredWitnessBuilders {
+    pub pre_builders:       Vec<WitnessBuilder>,
+    pub pre_segment_starts: Vec<usize>,
+    pub inverse_batches:    Vec<Vec<WitnessBuilder>>,
+}
+
+impl LayeredWitnessBuilders {
+    pub fn layers_len(&self) -> usize {
+        self.pre_segment_starts.len()
+    }
+
+    pub fn layer_range(&self, seg_idx: usize) -> (usize, usize) {
+        let start = self.pre_segment_starts[seg_idx];
+        let end = self
+            .pre_segment_starts
+            .get(seg_idx + 1)
+            .copied()
+            .unwrap_or(self.pre_builders.len());
+        (start, end)
+    }
+}
+
+/// Internal state manager for the layer scheduling algorithm.
+///
+/// Encapsulates the complex state tracking needed for frontier-based
+/// topological sorting with deferred inverse batching.
+pub struct LayerScheduler<'a> {
+    witness_builders: &'a [WitnessBuilder],
+    adjacency_list:   Vec<Vec<usize>>,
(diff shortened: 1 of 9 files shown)


provekit/common/src/witness/layer_scheduler.rs
```diff
@@ -0,0 +1,374 @@
...
+        ConstantOrR1CSWitness, ConstantTerm, ProductLinearTerm, SumTerm, WitnessCoefficient,
+        BINOP_ATOMIC_BITS,
+    },
+    std::{
+        collections::{HashMap, HashSet, VecDeque},
+        mem,
+    },
+};
+
+/// Layered plan for serial batched-inversion solving.
```
Bisht13: Layering is not only for batched inverse solving, need to update the comment

provekit/common/src/witness/layer_scheduler.rs
```diff
@@ -0,0 +1,374 @@
...
+///
+/// Optimized single-array design:
+/// - `builders` contains ALL builders in execution order: [pre1, pre2, ...,
+///   inv1, inv2, ..., pre3, ...]
+/// - For layer i: PRE slice is `builders[pre_starts[i] .. inverse_starts[i]]`
```
Bisht13: What does PRE stand for, it's not clear here

provekit/common/src/witness/layer_scheduler.rs
```diff
@@ -0,0 +1,374 @@
...
+/// - For layer i: INVERSE slice is `builders[inverse_starts[i] ..
+///   pre_starts[i+1] or end]`
+/// - Perfect cache locality: execution order = memory order
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize, PartialEq)]
+pub struct LayeredWitnessBuilders {
+    pub builders:       Vec<WitnessBuilder>,
```
Bisht13: Why not just only use Layer<Vec<SubLayer>> instead of these 3 members where Layer is Vec\<T\> and SubLayer is (enum, Vec\<WitnessBuilder\>)? Also, where is the layer logic in the struct, like which witness builders fall under which layer?

provekit/common/src/witness/layer_scheduler.rs
```diff
@@ -0,0 +1,374 @@
...
+        // Restore deferred nodes
+        self.frontier = deferred;
+
+        // Return true if we made progress
+        self.frontier.len() < initial_frontier_size
+    }
+
+    #[inline]
+    fn can_process_now(&self, node_idx: usize) -> bool {
+        // Quick check: does this node read any pending inverse outputs?
```
Bisht13: I don't think so this check is needed since inverse outputs are never read.

shreyas-londhe: This check is needed because there are witness builders which consume the outputs of the `Inverse` Witness Builders. Specifically for LogUp where inverses are multiplied with Multiplicities and later Sum'ed for the final logup-check. Reference - https://github.com/worldfnd/provekit/blob/main/provekit/r1cs-compiler/src/binops.rs#L159-L175

provekit/common/src/witness/layer_scheduler.rs
shreyas-londhe: This can be done and would be more readable, but the goal was to have a single re-ordered array of witness builders with just layer boundaries marked. The comments can be improved and I'll update them.

provekit/common/src/witness/layer_scheduler.rs
shreyas-londhe: Makes sense, I'll update the comment.

provekit/common/src/witness/layer_scheduler.rs
Bisht13: We want to have layering for all witness builder without treating inverses specially. All this parallelisation would actually show difference. At the moment, this PR caters specifically to inverses, which it shouldn't.

Bisht13 (approved):

PR #198, opened 2025-09-22
feat(ffi): add provekit-ffi

Bisht13:


tooling/provekit-ffi/src/utils.rs
```diff
@@ -0,0 +1,19 @@
...
+    anyhow::Result,
+    std::{ffi::CStr, os::raw::c_char},
+};
+
+/// Internal helper to convert C string to Rust string
+///
+/// # Safety
+///
+/// The caller must ensure that `ptr` is a valid null-terminated C string.
+pub unsafe fn c_str_to_str(ptr: *const c_char) -> Result<&'static str, PKError> {
```
yash25198: Returns 'static lifetime but the underlying C string has no such guarantee if the caller frees the string while Rust code holds the reference.

tooling/provekit-ffi/src/ffi.rs
```diff
@@ -0,0 +1,171 @@
...
+/// # Safety
+///
+/// The caller must ensure that:
+/// - The buffer was allocated by a ProveKit FFI function
+/// - The buffer is not used after calling this function
+/// - This function is called exactly once for each allocated buffer
+#[no_mangle]
+pub unsafe extern "C" fn pk_free_buf(buf: PKBuf) {
+    if !buf.ptr.is_null() && buf.len > 0 {
+        drop(Vec::from_raw_parts(buf.ptr, buf.len, buf.len));
```
yash25198: If the original Vec had capacity > len, we're reconstructing with wrong capacity. 
example :
```
let mut v: Vec<u8> = Vec::with_capacity(100);  // capacity = 100
v.push(1);
v.push(2);
v.push(3);
// Now: len = 3, capacity = 100
```
and then in types.rs
```
pub fn from_vec(mut v: Vec<u8>) -> Self {
    let ptr = v.as_mut_ptr();
    let len = v.len();          // len = 3
    std::mem::forget(v);        // capacity (100) 
    Self { ptr, len }           // PKBuf { ptr, len: 3 } 
}
```
and then `pk_free_buf` reconstructs with wrong capacity
```
drop(Vec::from_raw_parts(buf.ptr, buf.len, buf.len));
// len=3    cap=3 (was 100)
```
the allocator allocated 100 bytes but only 3 bytes were freed

tooling/provekit-ffi/src/ffi.rs
```diff
@@ -0,0 +1,171 @@
...
+    out_buf: *mut PKBuf,
+) -> c_int {
+    if out_buf.is_null() {
+        return PKError::InvalidInput.into();
+    }
+
+    catch_panic(PKError::ProofError.into(), || {
+        let out_buf = match out_buf.as_mut() {
+            Some(buf) => buf,
+            None => return PKError::InvalidInput.into(),
```
yash25198: unreachable condition

tooling/provekit-ffi/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+# Noir language
+acir.workspace = true
+noirc_abi.workspace = true
+
+# 3rd party
+anyhow.workspace = true
+serde.workspace = true
+serde_json.workspace = true
+postcard.workspace = true
+tracing.workspace = true
```
yash25198: unused dependency

PR #200, opened 2025-09-23
fix:  automate bench files creation

ashpect:
- Remove hardcoded posiedon rounds file (np and nps)
- fix ci to do dynamic benching hence removing need to locally remove and push new .np .nps file everytime
- fix bench paths
- add -r1cs flag for generate_gnark


.github/workflows/benchmark.yml
```diff
@@ -10,28 +10,60 @@ on:
...
       - name: Upload benchmark binaries
         uses: actions/upload-artifact@v4
         with:
           name: benchmark-binaries
           path: target/codspeed
+      
+      - name: Compile and execute Nargo circuit
+        working-directory: noir-examples/poseidon-rounds
+        run: |
+          nargo compile --skip-brillig-constraints-check --force
```
shreyas-londhe, line 47: todo: please run `nargo execute` as well.

.github/workflows/benchmark.yml
```diff
@@ -68,13 +106,20 @@ jobs:
...
         run: chmod -R +x target/codspeed
 
-      - name: Verify codspeed installation
+      # - name: Verify codspeed installation
+      #   run: |
+      #     echo "cargo-codspeed location:"
+      #     which cargo-codspeed 
+      #     echo "codspeed-post-bench location:"
+      #     which codspeed-post-bench
+
```
shreyas-londhe: question: Why is this commented out?

tooling/provekit-bench/benches/poseidon-1000.toml
```diff
@@ -1,10 +0,0 @@
-plains = [1, 2]
-
-# Value for 100 rounds
-# result = '0x0b50a50fda9459e4eda4b0e9382fcbb5d00c6b387bd6b79b9efb3ff95c2a32b6'
-
-# Value for 1000 rounds
-result = '0x0b5b6e4be36ffe606c2f640d972634fcf030bd0346eadcec12d0e3b95bf7f523'
-
-# Value for 10000 rounds
-# result = '0x235d6e083cc444e44782af67ba6be6e9b358e24a52188aea85c5643e3aab9125'
```
shreyas-londhe, line 10: todo: This file is needed for `nargo execute` and shouldn't be deleted.

.github/workflows/benchmark.yml
ashpect: I had added this for debugging in an earlier pr for codspeed debugging. I don't think this is relevant anymore and it's better to comment out than remove than debugging.

tooling/provekit-bench/benches/poseidon-1000.toml
ashpect: Since this is the same file as in noir-examples/posiedon-rounds/Prover.toml, it is better to be read from that path for consistency

.github/workflows/benchmark.yml
ashpect: The benches run internal performance benchmarks and do not run comparative benchmarking in the ci. Hence, running execute does not seem of any use.
Is there any reason why to run it ? I will edit the name to compile only if its not needed.

shreyas-londhe: Nargo execute is run to check if the circuit constraints are satisfied before we move on to next steps. So it's good to run execute here.

.github/workflows/benchmark.yml
shreyas-londhe: Anything not relevant should be removed, you can have this locally to debug.

PR #203, opened 2025-09-24
Feat: Add BlackBox::Sha256Compression

shreyas-londhe:
## Description
Adds complete SHA256 compression blackbox function support to `NoirToR1CSCompiler`, enabling efficient SHA256 hashing in R1CS circuits with significant performance improvements through optimized binary operation batching.

## Key Innovation: Global Binary Operation Optimization
Implements novel approach to decouple binary operation value computation from constraint generation:
- **Collection Phase**: `add_binop()` allocates witnesses, adds `WitnessBuilder`s for immediate value computation, and collects operations
- **Constraint Phase**: `add_binop_constraints()` processes all collected AND/XOR operations in a single batch with shared lookup tables
- **Benefits**: Eliminates redundant digital decompositions and lookup table setups, dramatically reducing memory usage and constraint count

## Performance Results
Benchmarked on `complete_age_check` example:

| Metric | Main Branch | SHA256 Blackbox | Improvement |
|--------|-------------|-----------------|-------------|
| Opcodes | 590,297 | 84,204 | **86% reduction** |
| Proving Time | 22s | ~9s | **59% faster** |
| Peak Memory | 10.4GB | 5.25GB | **50% reduction** |
| Constraints | 3,604,748 | 2,503,439 | **31% reduction** |
| Witnesses | 5,074,284 | 3,578,996 | **29% reduction** |

### Core Implementation
- **`sha256_compression.rs`**: New module implementing SHA256 compression algorithm
  - `add_sha256_compression()`: Main entry point processing 16 input words + 8 hash values → 8 output words
  - 64-round compression loop using SHA256 constants and message schedule

### Binary Operation Batching
- **`binops.rs`**: Refactored for collection-based processing
  - `add_binop()`: Collects operations while computing values immediately
  - `add_binop_constraints()`: Processes all collected operations with shared digital decomposition and lookup tables

### WitnessBuilder Extensions
- **`WitnessBuilder::U32Addition`**: Computes `result = (a + b) % 2^32`, `carry = (a + b) / 2^32`
- **`WitnessBuilder::And`** / **`WitnessBuilder::Xor`**: Bitwise operations for 32-bit values
- Added corresponding solver logic in `prover/src/witness/witness_builder.rs`

### Integration
- **`noir_to_r1cs.rs`**: Added SHA256 compression blackbox function handling with batched binary operation processing

### DIFF
--- a/provekit/r1cs-compiler/src/sha256_compression.rs
+++ b/provekit/r1cs-compiler/src/sha256_compression.rs
@@ -460,7 +460,7 @@ pub(crate) fn add_maj(
 }
 
 /// SHA256 message schedule expansion: expand 16 u32 words to 64 u32 words
-/// W[i] = σ₁(W[i-2]) + W[i-7] + σ₀(W[i-15]) + W[i-16] for i = 16..64
+/// W\[i\] = σ₁(W\[i-2\]) + W\[i-7\] + σ₀(W\[i-15\]) + W\[i-16\] for i = 16..64
 pub(crate) fn add_message_schedule_expansion(
     r1cs_compiler: &mut NoirToR1CSCompiler,
     xor_ops: &mut Vec<(ConstantOrR1CSWitness, ConstantOrR1CSWitness, usize)>,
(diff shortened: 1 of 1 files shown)


provekit/r1cs-compiler/src/binops.rs
```diff
@@ -20,11 +20,39 @@ pub enum BinOp {
     Xor,
 }
 
+/// Allocate a witness for a binary operation result, add the appropriate
+/// WitnessBuilder for value computation, and collect the operation for later
+/// constraint generation. Returns the witness index of the result.
+pub(crate) fn add_binop(
```
veljkovranic, line 26: Maybe we can just say here that you need to call add_binop_constrains on every collected_ops Vec that you pass into this method, if we end up needing to add more Binary ops than we did so far

veljkovranic (approved):
Left a minor comment, otherwise great work!
Really awesome PR :)

PR #205, opened 2025-09-29
Add input generation to FFI.

piohei:
Move provekit-input-gen to tooling and allow to use it over FFI.


Bisht13:
We should not move passport input gen under tooling and also should add the functionality behind a `passport` flag in FFI.

piohei:
Should I stay in playground? I thought that other people/teams will use it as a library provided with provekit. Where do you think it should be?

PR #213, opened 2025-10-10
feat: poseidon2 blackbox impl in Noir

rose2221:
### Poseidon2 Permutation Blackbox 

This PR implements `BlackBox::Poseidon2Permutation` in the R1CS backend for all supported widths t ∈ {2, 3, 4, 8, 12, 16}, wiring the opcode to a full Poseidon2 permutation.

```
t ∈ {2, 3, 4, 8, 12, 16}
```

The permutation is executed using the standard Poseidon2 round schedule:

```
4 full rounds
+ Rp partial rounds
+ 4 full rounds
```

All MDS operations (external and internal) are represented as **linear forms**, so they do not emit any constraints during the permutation itself. Only the S-box applications `x → x⁵` produce non-linear R1CS constraints.

For each active S-box, the `linear_form_pow5` helper emits exactly **3 constraints**. Final outputs are materialised into `t` output witnesses using one linear constraint per lane.

After the gadget completes, Noir emits **t additional AssertZero linear constraints** to wire the blackbox outputs into the ACIR result.

#### **Constraint Summary (general t)**

```
Poseidon2 non-linear constraints = 24·t + 3·Rp
Final materialisation     =  t
Noir AssertZero wiring    =  t
-------------------------------------------------
Total R1CS                = 26t + 3·Rp
```

#### **Example (t = 4)**

For `t=4` and `Rp=56`:

```
Poseidon2 non-linear constraints = 24·4 + 3·56 = 264
Final materialisation = 4
Noir AssertZero       = 4
----------------------------------------
Total R1CS constraints = 272

### DIFF
--- a/provekit/r1cs-compiler/src/poseidon2/permutation.rs
+++ b/provekit/r1cs-compiler/src/poseidon2/permutation.rs
@@ -197,7 +197,7 @@ impl LinearForm {
         }
 
         // Shared owner: clone, canonicalize, then re-wrap.
-        let mut owned: Vec<(FieldElement, usize)> = self.terms.as_ref().clone();
+        let mut owned: Vec<(FieldElement, usize)> = self.terms.to_vec();
         collapse_terms_inplace(&mut owned, threshold);
         self.terms = Rc::new(owned);
     }

### CODE REVIEW

File: provekit/r1cs-compiler/src/lib.rs [hunk 1]
```diff
@@ -3,6 +3,8 @@ mod digits;
 mod memory;
 mod noir_proof_scheme;
 mod noir_to_r1cs;
+pub(crate) mod poseidon2_constants;
+pub(crate) mod poseidon2_permutation;
```
shreyas-londhe: todo: We should create a module for blackbox implementations.

provekit/r1cs-compiler/src/poseidon2/constants.rs
```diff
@@ -0,0 +1,950 @@
...
+
+pub fn amount_partial_rounds(t: u32) -> u32 {
+    if t <= 4 {
+        56
+    } else {
+        57
+    }
+}
+
+pub fn load_diag(t: u32) -> Vec<F> {
```
shreyas-londhe, line 62: issue: The file contains hundreds of hardcoded cryptographic constants with no documentation about their source, generation method, or verification. This is a critical security risk — incorrect constants would completely break the security of Poseidon2 while still producing valid-looking proofs.

suggestion: Add documentation about their source, generation method, or verification. Also add a test that verifies at least a few constants against known test vectors from the reference implementation.

provekit/r1cs-compiler/src/poseidon2_constants.rs
```diff
@@ -0,0 +1,950 @@
...
+    let s = s.strip_prefix("0x").unwrap_or(s);
+    let mut bytes = hex_to_bytes(s); // trivial hex -> Vec<u8>
+                                     // left-pad to field length
+    let L = ((F::MODULUS_BIT_SIZE as usize) + 7) / 8;
+    assert!(bytes.len() <= L, "constant too large for field");
+    if bytes.len() < L {
+        let mut pad = vec![0u8; L - bytes.len()];
+        pad.extend_from_slice(&bytes);
+        bytes = pad;
+    }
```
shreyas-londhe:
```suggestion
    let field_byte_len = ((F::MODULUS_BIT_SIZE as usize) + 7) / 8;
    assert!(bytes.len() <= field_byte_len, "constant too large for field");
    if bytes.len() < field_byte_len {
        let mut pad = vec![0u8; field_byte_len - bytes.len()];
        pad.extend_from_slice(&bytes);
        bytes = pad;
    }
```

provekit/r1cs-compiler/src/poseidon2_permutation.rs
```diff
@@ -0,0 +1,303 @@
...
+    ark_ff::Field,
+    ark_std::One,
+    provekit_common::{
+        witness::{ConstantOrR1CSWitness, SumTerm, WitnessBuilder},
+        FieldElement,
+    },
+};
+
+// y = a + c  (c is a field constant)
+fn add_add_const(cs: &mut NoirToR1CSCompiler, a: usize, c: FieldElement) -> usize {
```
shreyas-londhe: todo: We should follow consistent naming. in `Sha256Blackbox` we are using complete names -- `r1cs_compiler` which should be followed.

provekit/r1cs-compiler/src/poseidon2_permutation.rs
```diff
@@ -0,0 +1,303 @@
...
+                            block_out[i][j],
+                            col_acc[j],
+                        ]);
+                }
+            }
+            out
+        }
+        _ => panic!("unsupported t for external MDS"),
+    }
+}
```
shreyas-londhe: issue: The external MDS implementations for t=2,3,4 are optimized (reducing multiplications), while t=8,12,16 use a block-based approach. 

suggestion: Add documentation explaining:
- Why these specific optimizations
- How they relate to the Poseidon2 paper
- Whether constraint counts match expectations

provekit/r1cs-compiler/src/poseidon2_permutation.rs
```diff
@@ -0,0 +1,303 @@
...
+                            cs.witness_one(),
+                        )]));
+                        // constrain y = c
+                        cs.r1cs.add_constraint(
+                            &[(FieldElement::ONE, y)],
+                            &[(FieldElement::ONE, cs.witness_one())],
+                            &[(*c, cs.witness_one())],
+                        );
+                        y
+                    }
```
shreyas-londhe: issue: Every constant input immediately becomes a witness + constraint (lines 230-244). This is wasteful because:
- Constant operations (add/mul) can be computed at compile time
- Constant × Witness requires only linear constraints, not multiplication constraints
- MDS matrix operations often propagate many constants through the computation

suggestion: `state` can be refactored as `Vec<ConstantOrWitness>` and when constants are used in constraints, they should be used via a linear constraint and not a separate witness. Example -- https://github.com/worldfnd/provekit/blob/main/provekit/r1cs-compiler/src/sha256_compression.rs#L110-L115

provekit/r1cs-compiler/src/poseidon2_permutation.rs
```diff
@@ -0,0 +1,303 @@
...
+                }
+            })
+            .collect();
+
+        // Initial linear layer
+        state = external_mds_t(cs, &state);
+
+        // First 4 full rounds
+        for r in 0..4 {
```
shreyas-londhe: todo: Please document why the number of rounds has been fixed to 4 by referring to reference implementation or official Poseidon2 paper.

provekit/r1cs-compiler/src/noir_to_r1cs.rs
```diff
@@ -436,6 +466,7 @@ impl NoirToR1CSCompiler {
         // For the AND and XOR operations, add the appropriate constraints.
         add_binop_constraints(self, BinOp::And, and_ops);
         add_binop_constraints(self, BinOp::Xor, xor_ops);
+        add_poseidon2_permutation(self, poseidon2_ops);
```
shreyas-londhe, line 470:
```suggestion
        // For the Poseidon2 permutation operation.
        add_poseidon2_permutation(self, poseidon2_ops);
```

shreyas-londhe (changes requested):

provekit/r1cs-compiler/src/lib.rs
rose2221: done!

provekit/r1cs-compiler/src/poseidon2/constants.rs
rose2221: added inline documentation describing the source and generation method for the constants, also includes tests to cross-check them with the TaceoLabs reference constants.

provekit/r1cs-compiler/src/poseidon2_permutation.rs
rose2221: done!

provekit/r1cs-compiler/src/poseidon2_permutation.rs
rose2221: done!

provekit/r1cs-compiler/src/poseidon2_permutation.rs
rose2221: done!

provekit/r1cs-compiler/src/poseidon2_permutation.rs
rose2221: i've added the comments for the same!

provekit/r1cs-compiler/src/poseidon2/permutation.rs
```diff
@@ -0,0 +1,442 @@
...
+    let mut c0 = FieldElement::from(0u64);
+    for (c, v) in coeffs.iter().copied().zip(vs.iter()) {
+        match *v {
+            ConstantOrR1CSWitness::Witness(w) => lc.push((c, w)),
+            ConstantOrR1CSWitness::Constant(k) => c0 += c * k,
+        }
+    }
+
+    let y = cs.num_witnesses();
+    cs.r1cs.add_witnesses(1);
```
shreyas-londhe: issue: Manual witness addition should be avoided. Because of this extra witnesses are added which are not constrained and increase witness size. 

Ref: https://github.com/worldfnd/provekit/pull/227

provekit/r1cs-compiler/src/poseidon2/permutation.rs
```diff
@@ -0,0 +1,442 @@
...
+            FieldElement::ONE,
+            y,
+        )]);
+
+    ConstantOrR1CSWitness::Witness(y)
+}
+
+// Multiply two values: linearize const*w, use product gate for w*w.
+fn add_mul(
+    cs: &mut NoirToR1CSCompiler,
```
shreyas-londhe: todo: `R1CSCompiler` is still named as `cs` which is misleading, it should be name `r1cs_compiler`

provekit/r1cs-compiler/src/poseidon2/permutation.rs
```diff
@@ -0,0 +1,442 @@
...
+        Some(k),
+        cs.witness_one(),
+    )]));
+    cs.r1cs.add_constraint(
+        &[(FieldElement::ONE, y)],
+        &[(FieldElement::ONE, cs.witness_one())],
+        &[(k, cs.witness_one())],
+    );
+    y
+}
```
shreyas-londhe: issue: If `k` is a constant, it can be directly embedded in R1CS instead of assigning it to a witness. 

**BAD - Constant as Witness:**
```rust
// Wastes a witness slot
let five_witness = r1cs_compiler.num_witnesses();
r1cs_compiler.add_witness_builder(WitnessBuilder::Constant(five_witness, FieldElement::from(5)));

// Constraint: x + w[five_witness] = y  (where w[five_witness] = 5)
r1cs.add_constraint(
    &[(FieldElement::ONE, x), (FieldElement::ONE, five_witness)],
    &[(FieldElement::ONE, witness_one)],
    &[(FieldElement::ONE, y)],
);
// Result: (w[x] + w[five_witness]) * 1 = w[y]
```

**GOOD - Constant Embedded:**
```rust
// No extra witness needed
// Constraint: x + 5 = y  (5 embedded as coefficient)
r1cs.add_constraint(
    &[(FieldElement::ONE, x), (FieldElement::from(5), witness_one)],
    &[(FieldElement::ONE, witness_one)],
    &[(FieldElement::ONE, y)],
);
// Result: (w[x] + 5*1) * 1 = w[y]
```

Both compute `x + 5 = y`, but the good version uses one less witness by exploiting that `witness_one` always equals 1, so `5 * witness_one = 5`.

provekit/r1cs-compiler/src/poseidon2/permutation.rs
rose2221: it's been updated.

provekit/r1cs-compiler/src/poseidon2/permutation.rs
rose2221: done.

provekit/r1cs-compiler/src/poseidon2/permutation.rs
rose2221: Removed `materialize_const` entirely, MDS functions now return `ConstantOrR1CSWitness` to avoid premature witness materialization.

provekit/r1cs-compiler/src/poseidon2/permutation.rs
```diff
@@ -0,0 +1,757 @@
...
+    sum_terms: &mut Vec<SumTerm>,
+    constraint_terms: &mut Vec<(FieldElement, usize)>,
+) {
+    let t = forms.len();
+    assert_eq!(t, rc.len());
+
+    out.clear();
+    out.reserve(t);
+
+    for (i, mut form) in forms.iter_mut().enumerate() {
```
shreyas-londhe: **todo**: There is no need to have `form` to be mutable as you are already creating a mutable iterator.

provekit/r1cs-compiler/src/poseidon2/permutation.rs
```diff
@@ -0,0 +1,757 @@
...
+        }
+
+        // Unique owner: canonicalize in-place.
+        if let Some(vec_mut) = Rc::get_mut(&mut self.terms) {
+            collapse_terms_inplace(vec_mut, threshold);
+            return;
+        }
+
+        // Shared owner: clone, canonicalize, then re-wrap.
+        let mut owned: Vec<(FieldElement, usize)> = self.terms.as_ref().clone();
```
shreyas-londhe:
```suggestion
        let mut owned: Vec<(FieldElement, usize)> = self.terms.to_vec();
```

provekit/r1cs-compiler/src/poseidon2/permutation.rs
```diff
@@ -0,0 +1,757 @@
...
+
+/// external MDS2 as linear forms
+fn mds2_block_forms(s: &[ConstantOrR1CSWitness]) -> Vec<LinearForm> {
+    let f1 = FieldElement::ONE;
+    let f2 = FieldElement::from(2u64);
+
+    let o0 = compute_linear_form(&[f2, f1], s);
+    let o1 = compute_linear_form(&[f1, f2], s);
+    vec![o0, o1]
+}
```
shreyas-londhe: **todo**: If the return is always going to be an array of fixed length, we should return a slice instead of an array here. 

```suggestion
/// external MDS2 as linear forms
fn mds2_block_forms(s: &[ConstantOrR1CSWitness]) -> [LinearForm; 2] {
    let f1 = FieldElement::ONE;
    let f2 = FieldElement::from(2u64);

    let o0 = compute_linear_form(&[f2, f1], s);
    let o1 = compute_linear_form(&[f1, f2], s);
    vec![o0, o1]
}
```

provekit/r1cs-compiler/src/poseidon2/permutation.rs
```diff
@@ -0,0 +1,757 @@
...
+            r1cs_compiler,
+            &form,
+            sum_terms,
+            constraint_terms,
+        ));
+    }
+}
+
+/// external MDS2 as linear forms
+fn mds2_block_forms(s: &[ConstantOrR1CSWitness]) -> Vec<LinearForm> {
```
shreyas-londhe: **todo**: It is best practice to return the most specific type in a function. Here it is `[LinearForm; 2]`. Please make the change in other places where similar pattern is observed.

provekit/r1cs-compiler/src/poseidon2/permutation.rs
rose2221: Agreed, removed the unnecessary mut.

provekit/r1cs-compiler/src/poseidon2/permutation.rs
rose2221: Done, updated fixed-size MDS helpers to return [LinearForm; N].

PR #215, opened 2025-10-13
pub_inputs support

ashpect:
Support for public_witness opening

### DIFF
--- a/recursive-verifier/app/circuit/circuit.go
+++ b/recursive-verifier/app/circuit/circuit.go
@@ -389,7 +389,6 @@ func verifyCircuit(
 		Values: make([]frontend.Variable, len(publicInputs.Values)),
 	}
 
-
 	circuit := Circuit{
 		IO:                                      []byte(cfg.IOPattern),
 		Transcript:                              contTranscript,
--- a/provekit/r1cs-compiler/src/whir_r1cs.rs
+++ b/provekit/r1cs-compiler/src/whir_r1cs.rs
@@ -17,13 +17,23 @@ const MIN_WHIR_NUM_VARIABLES: usize = 12;
 const MIN_SUMCHECK_NUM_VARIABLES: usize = 1;
 
 pub trait WhirR1CSSchemeBuilder {
-    fn new_for_r1cs(r1cs: &R1CS, w1_size: usize, num_challenges: usize, has_public_inputs: bool) -> Self;
+    fn new_for_r1cs(
+        r1cs: &R1CS,
+        w1_size: usize,
+        num_challenges: usize,
+        has_public_inputs: bool,
+    ) -> Self;
 
     fn new_whir_config_for_size(num_variables: usize, batch_size: usize) -> WhirConfig;
 }
 
 impl WhirR1CSSchemeBuilder for WhirR1CSScheme {
-    fn new_for_r1cs(r1cs: &R1CS, w1_size: usize, num_challenges: usize, has_public_inputs: bool) -> Self {
+    fn new_for_r1cs(
+        r1cs: &R1CS,
+        w1_size: usize,
+        num_challenges: usize,
+        has_public_inputs: bool,
+    ) -> Self {
         let total_witnesses = r1cs.num_witnesses();
         assert!(
             w1_size <= total_witnesses,


provekit/prover/src/r1cs.rs
```diff
@@ -18,8 +18,12 @@ pub trait R1CSSolver {
         &self,
         plan: LayeredWitnessBuilders,
         acir_map: WitnessMap<NoirElement>,
+        acir_public_inputs: &[u32],
         transcript: &mut ProverState<SkyscraperSponge, FieldElement>,
-    ) -> Vec<Option<FieldElement>>;
+    ) -> (
+        Vec<Option<FieldElement>>,
+        std::collections::HashMap<u32, usize>,
```
shreyas-londhe: todo: Import `HashMap`.

provekit/prover/src/whir_r1cs.rs
```diff
@@ -78,19 +89,29 @@ impl WhirR1CSProver for WhirR1CSScheme {
...
         // Compute weights from R1CS instance
         let alphas = calculate_external_row_of_r1cs_matrices(alpha, r1cs);
-        let (statement, f_sums, g_sums) = create_combined_statement_over_two_polynomials::<3>(
+        let (mut statement, f_sums, g_sums) = create_combined_statement_over_two_polynomials::<3>(
             self.m,
             &commitment_to_witness,
-            masked_polynomial,
-            random_polynomial,
+            masked_polynomial.clone(),
+            random_polynomial.clone(),
```
shreyas-londhe: todo: Please use ref here.

provekit/common/src/utils/sumcheck.rs
```diff
@@ -114,6 +114,10 @@ pub trait SumcheckIOPattern {
     fn add_rand(self, num_rand: usize) -> Self;
 
     fn add_zk_sumcheck_polynomials(self, num_vars: usize) -> Self;
+
+    /// Prover sends the hash of the public inputs
+    /// Verifier sends randomness to construct weights
+    fn add_public_inputs(self) -> Self;
```
shreyas-londhe: todo: This should not be under `SumcheckIOPattern`. Create a separate trait for it for clarity.

provekit/common/Cargo.toml
```diff
@@ -38,6 +38,7 @@ ruint.workspace = true
 serde.workspace = true
 serde_json.workspace = true
 tracing.workspace = true
+sha2.workspace = true
```
Bisht13: The `sha2` dependency is added but appears to be unused in the Rust code.

recursive-verifier/app/circuit/mtUtilities.go
```diff
@@ -112,3 +112,24 @@ func rlcBatchedLeaves(api frontend.API, leaves [][]frontend.Variable, foldSize i
 	}
 	return collapsed
 }
+
+// hashPublicInputs computes the hash of public inputs as field elements sequentially
+func hashPublicInputs(sc *skyscraper.Skyscraper, publicInputs PublicInputs) (frontend.Variable, error) {
```
Bisht13, line 117: The following must produce identical results:
- Rust: PublicInputs::hash() using SkyscraperCRH::evaluate()
- Go: hashPublicInputs() using sc.CompressV2()

Add a cross-language test with known test vectors.

recursive-verifier/app/utilities/utilities.go
```diff
@@ -58,6 +62,77 @@ func IndexOf(_ *big.Int, inputs []*big.Int, outputs []*big.Int) error {
 	return nil
 }
 
+// HashPublicInputsHint is a hint function that computes SHA-256 hash of public inputs
+// matching the Rust PublicInputs::hash() implementation.
+// It takes public input values, converts them to BigInt, extracts limbs, hashes them,
+// and returns the hash as a field element.
+func HashPublicInputsHint(_ *big.Int, inputs []*big.Int, outputs []*big.Int) error {
```
Bisht13: This SHA-256 based hint function appears to be UNUSED.  The circuit uses `hashPublicInputs()` in mtUtilities.go which uses Skyscraper, not SHA-256.

recursive-verifier/app/circuit/circuit.go
```diff
@@ -125,12 +170,15 @@ func (circuit *Circuit) Define(api frontend.API) error {
 		bz = api.Add(circuit.WitnessClaimedEvaluations[0][1], circuit.WitnessClaimedEvaluations[1][1])
 		cz = api.Add(circuit.WitnessClaimedEvaluations[0][2], circuit.WitnessClaimedEvaluations[1][2])
 	} else {
+		log.Println("Single Mode")
```
Bisht13: Remove debug log.

recursive-verifier/app/circuit/circuit.go
```diff
@@ -268,6 +385,13 @@ func verifyCircuit(
 		witnessBlindingEvals = [][]frontend.Variable{gSums}
 	}
 
+	// Empty container while circuit creation
+	publicInputsContainer := PublicInputs{
+		Values: make([]frontend.Variable, len(publicInputs.Values)),
+	}
+
+	log.Println("publicInputs", publicInputs)
```
Bisht13: Remove debug log.

provekit/common/src/witness/scheduling/splitter.rs
```diff
@@ -165,9 +182,53 @@ impl<'a> WitnessSplitter<'a> {
...
+        let mut public_input_builder_indices = Vec::new();
+        let mut rest_indices = Vec::new();
+
+        // Sanity Check: Make sure all public inputs and WITNESS_ONE_IDX are in
+        // w1_indices.
+        // Convert to HashSet for O(1) lookups since we're checking many times
+        let w1_indices_set = w1_indices.iter().copied().collect::<HashSet<_>>();
+        for &idx in acir_public_inputs_indices_set.iter() {
+            if !w1_indices_set.contains(&(idx as usize)) {
+                panic!("Public input {} is not in w1_indices", idx);
```
Bisht13: Panicking on validation errors is not idiomatic Rust. The function signature should return Result<(Vec<usize>, Vec<usize>), SplitError> instead.

recursive-verifier/app/circuit/circuit.go
```diff
@@ -396,10 +528,15 @@ func verifyCircuit(
...
 		MatrixB:                                 matrixB,
 		MatrixC:                                 matrixC,
+		PublicInputs:                            publicInputs,
 	}
 
 	witness, _ := frontend.NewWitness(&assignment, ecc.BN254.ScalarField())
-	publicWitness, _ := witness.Public()
+	publicWitness, err := witness.Public()
+	if err != nil {
+		log.Printf("Failed witess,Public(): %v", err)
```
Bisht13: Typo: "witness"

recursive-verifier/app/circuit/circuit.go
```diff
@@ -150,23 +198,74 @@ func (circuit *Circuit) Define(api frontend.API) error {
...
+
 		// Single mode - existing logic
 		matrixExtensionEvals := evaluateR1CSMatrixExtension(api, circuit, spartanSumcheckRand, whirFoldingRandomness)
 		for i := 0; i < 3; i++ {
-			api.AssertIsEqual(matrixExtensionEvals[i], circuit.WitnessLinearStatementEvaluations[i])
+			api.AssertIsEqual(matrixExtensionEvals[i], circuit.WitnessLinearStatementEvaluations[offset+i])
 		}
 	}
 
+	// Geomteric weights for public inputs
```
Bisht13: Typo: "Geometric"

provekit/common/src/witness/scheduling/splitter.rs
```diff
@@ -165,9 +182,53 @@ impl<'a> WitnessSplitter<'a> {
...
+            }
+        }
+
+        // Separate into: 0, public inputs, and rest
+        for builder_idx in w1_indices {
+            if builder_idx == 0 {
+                continue; // Will add 0 first
+            } else if let WitnessBuilder::Acir(_, acir_idx) = &self.witness_builders[builder_idx] {
+                if acir_public_inputs_indices_set.contains(&(*acir_idx as u32)) {
+                    public_input_builder_indices.push(builder_idx);
```
Bisht13, line 234: Public input builder indices are not sorted
```rs
public_input_builder_indices.push(builder_idx);
```
The public_input_builder_indices vector collects indices in iteration order, which depends on the order of w1_indices. For deterministic behavior, sort by ACIR index:
```
public_input_builder_indices.sort_by_key(|&idx| {
    if let WitnessBuilder::Acir(_, acir_idx) = &self.witness_builders[idx] {
        *acir_idx
    } else {
        usize::MAX
    }
});
```

provekit/common/src/witness/mod.rs
```diff
@@ -41,3 +45,43 @@ impl ConstantOrR1CSWitness {
...
+        self.0.len()
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+
+    pub fn hash(&self) -> FieldElement {
+        match self.0.len() {
+            0 => FieldElement::from(0u64),
```
Bisht13, line 71: Why does empty return 0 instead of a domain-separated hash? This should be documented or use SkyscraperCRH::evaluate(&(), &[]) for consistency.

provekit/common/src/witness/mod.rs
```diff
@@ -41,3 +45,43 @@ impl ConstantOrR1CSWitness {
...
+            1 => {
+                // For single element, hash it with zero to ensure it gets properly hashed
```
Bisht13, line 73: The comment explains what but not why. Is this to match Skyscraper's compression function expecting pairs? Document the cryptographic rationale.

recursive-verifier/app/circuit/circuit.go
```diff
@@ -150,23 +198,74 @@ func (circuit *Circuit) Define(api frontend.API) error {
...
+			circuit.WHIRParamsWitness.MVParamsNumberOfVariables, publicWeightsChallenge[0],
+		)
+
+		api.AssertIsEqual(publicWeightEval, circuit.WitnessLinearStatementEvaluations[0])
+	}
+
 	return nil
 }
 
+func computePublicWeightEvaluation(
```
Bisht13, line 234: This computes all 2^m eq polynomials which is exponential. Should replace with `geometric_till` algorithm.

provekit/common/src/whir_r1cs.rs
```diff
@@ -34,10 +34,11 @@ impl WhirR1CSScheme {
         if self.num_challenges > 0 {
             // Compute total constraints: OOD + statement
             // OOD: 2 witnesses × committment_ood_samples each
-            // Statement: 2 statements × 3 constraints each = 6
+            // Statement: statement_1 has 3 constraints + 1 public weights constraint = 4,
+            // statement_2 has 3 constraints = 3, total = 7
             let num_witnesses = 2;
             let num_ood_constraints = num_witnesses * self.whir_witness.committment_ood_samples;
-            let num_statement_constraints = 6; // 2 statements × 3 constraints
+            let num_statement_constraints = 7;
```
Bisht13: This is hardcoded and assumes:
- Batch mode: 3 constraints for statement1 + 1 public weights + 3 for statement2 = 7
- But what if public inputs are empty in batch mode?
The Go verifier calculates this dynamically:
```go
if !cfg.PublicInputs.IsEmpty() {
    witnessLinearStatementEvalsSize = 7
} else {
    witnessLinearStatementEvalsSize = 6
}
```

provekit/prover/src/lib.rs
```diff
@@ -112,14 +113,30 @@ impl Prove for Prover {
         self.r1cs
             .test_witness_satisfaction(&witness.iter().map(|w| w.unwrap()).collect::<Vec<_>>())
             .context("While verifying R1CS instance")?;
+
+        // Gather public inputs from witness
```
Bisht13, line 117: No validation that the public inputs extracted from witness match what the circuit expects. If `acir_public_inputs` indices are wrong, proof will be valid but verify wrong values.

provekit/verifier/src/whir_r1cs.rs
```diff
@@ -147,6 +198,20 @@ fn prepare_statement_for_witness_verifier<const N: usize>(
...
 }
 
+fn update_statement_for_witness_verifier(
+    m: usize,
+    statement_verifier: &mut Statement<FieldElement>,
+    parsed_commitment: &ParsedCommitment<FieldElement, FieldElement>,
+    whir_public_weights_query_answer: (FieldElement, FieldElement),
+) {
+    let (public_f_sum, public_g_sum) = whir_public_weights_query_answer;
+    let public_weight = Weights::linear(EvaluationsList::new(vec![FieldElement::zero(); 1 << m]));
```
Bisht13, line 195: The verifier creates dummy zero-weight vectors. This is correct because the verifier doesn't know the actual weights (only the prover does), but:
1. This should be commented explaining why zeros are OK
2. Verify that the WHIR protocol handles this correctly (the actual weighted sum comes from the hint)

provekit/verifier/src/whir_r1cs.rs
```diff
@@ -79,6 +79,32 @@ impl WhirR1CSVerifier for WhirR1CSScheme {
...
+                );
+
+                let mut public_weights_vector_random_buf = [FieldElement::zero()];
+                arthur.fill_challenge_scalars(&mut public_weights_vector_random_buf)?;
+
+                let whir_public_weights_query_answer: (FieldElement, FieldElement) = arthur
+                    .hint()
+                    .context("failed to read WHIR public weights query answer")?;
+
+                if !public_inputs.is_empty() {
```
Bisht13, line 99: When public inputs are empty, the hint `whir_public_weights_query_answer` is still read from the transcript but never validated. A malicious prover could put garbage here and it would be ignored.

provekit/prover/src/whir_r1cs.rs
```diff
@@ -159,16 +161,36 @@ impl WhirR1CSProver for WhirR1CSScheme {
...
-                commitment.random_polynomial,
+                &commitment.masked_polynomial,
+                &commitment.random_polynomial,
                 &alphas,
             );
 
             merlin.hint::<(Vec<FieldElement>, Vec<FieldElement>)>(&(f_sums, g_sums))?;
 
+            // VERIFY the size given by self.m
+            let public_weight = get_public_weights(public_inputs, &mut merlin, self.m);
```
Bisht13: Duplicate Logic Between Single/Batch Paths. Same pattern in the verifier, the public input hash verification and statement extension logic is copy-pasted between the two branches.

recursive-verifier/app/circuit/mtUtilities.go
ashpect: Ref in : https://github.com/worldfnd/provekit/issues/257

provekit/common/src/witness/scheduling/splitter.rs
ashpect: This is intentional to avoid compromised prover creating same proof for (2,1) and (1,2) type public inputs set

provekit/common/src/witness/mod.rs
ashpect: Skyscraper doesn't define hash for empty values, hence on empty hash is defined by us as 0 for pub inputs

provekit/common/src/witness/mod.rs
ashpect: Skyscraper cannot hash single element, it needs pairs, hence for a single public input, we hash it with a standard 0. 
Although a case arises (0,0) and (0) as public inputs would produce same hash but their geometric weights will be different hence this case doesn't create similar proofs.

recursive-verifier/app/circuit/circuit.go
ashpect: Ref : https://github.com/worldfnd/provekit/issues/258

provekit/prover/src/lib.rs
ashpect: acir_public_inputs come directly from noir program's abi and hence are a reflection of circuit. A malicious user compromising `program` artifact to change `public_inputs_indices` is equivalent to changing the circuit itself.

provekit/verifier/src/whir_r1cs.rs
ashpect: This is fine, as there are weights which are `deferred` in evaluation from whir side and whir doesn't actually evaluate them to verify. They are verified separately at the end, both in gnark and naive verifier, ref: https://github.com/worldfnd/provekit/pull/256

provekit/verifier/src/whir_r1cs.rs
ashpect: It doesn't break protocol. And keeping the io transcript dynamic for a single hint would end up creating multiple if else branches in the codebase which is not optimal. It's a tradeoff which is good.

PR #217, opened 2025-10-14
Fix logging, tracing and tracy

recmo:
Before, the default features disabled all logging. To enable even basic logging you'd had to enable `profilng` which also enabled a very expensive Tracy stack trace for each allocation, blowing up runtime by 4x. This changes:

* Logging is always enabled.
* Use the re-exported crates from `tracing_tracy` to avoid version mismatch.
* Make allocator counters an optional feature, but default since it is low overhead.
* Make Tracy an optional dependency.
* Make Tracy allocator tracing a runtime option, with configurable depth.

### DIFF
--- a/tooling/cli/src/cmd/mod.rs
+++ b/tooling/cli/src/cmd/mod.rs
@@ -21,10 +21,16 @@ pub struct Args {
     #[argh(switch)]
     pub tracy: bool,
 
-    /// enable tracy allocation tracking with provided stack depth
+    /// enable Tracy allocation tracking with provided stack depth, or 0 to
+    /// trace allocations without stack traces.
     #[cfg(feature = "tracy")]
     #[argh(option)]
     pub tracy_allocations: Option<usize>,
+
+    /// keep the process alive after completion to allow tracy to collect data
+    #[cfg(feature = "tracy")]
+    #[argh(switch)]
+    pub tracy_keepalive: bool,
 }
 
 #[derive(FromArgs, PartialEq, Debug)]
--- a/tooling/cli/src/main.rs
+++ b/tooling/cli/src/main.rs
@@ -7,7 +7,7 @@ mod span_stats;
 #[cfg(feature = "profiling-allocator")]
 use crate::profiling_alloc::ProfilingAllocator;
 #[cfg(feature = "tracy")]
-use tracing::info;
+use tracing::{info, warn};
 use {
     self::cmd::Command,
     anyhow::Result,
@@ -33,7 +33,14 @@ fn main() -> Result<()> {
             } else {
                 info!("Tracy profiling enabled (without allocation tracking).");
             }
-        };
+        } else {
+            if args.tracy_allocations.is_some() {
+                warn!("--tracy-allocations specified without --tracy, ignoring.");
+            }
+            if args.tracy_keepalive {
+                warn!("--tracy-keepalive specified without --tracy, ignoring.");
+            }
+        }
         subscriber.with(args.tracy.then(tracing_tracy::TracyLayer::default))
     };
 
@@ -42,21 +49,13 @@ fn main() -> Result<()> {
     // Run CLI command
     let res = args.run();
 
-    // SAFETY: It is safe to call `___tracy_shutdown_profiler()` here because all
-    // profiling work is complete, and no further Tracy API calls will be made
-    // after this point. This function must be called to properly shut down the
-    // Tracy profiler before program exit.
-    //
-    // @recmo: This is not safe. tracing_tracy may still
-    // be doing work in other threads. We should properly join all threads and
-    // ensure all work is complete before calling this function.
     #[cfg(feature = "tracy")]
-    if args.tracy {
-        info!("Shutting down Tracy client.");
-        ALLOCATOR.disable_tracy();
-        unsafe {
-            tracing_tracy::client::sys::___tracy_shutdown_profiler();
-        }
+    if args.tracy_keepalive {
+        use std::io::{stderr, stdin, stdout, Write};
+        eprintln!("Tracy keepalive enabled, press Enter to exit.");
+        stdout().flush()?;
+        stderr().flush()?;
+        stdin().read_line(&mut String::new())?;
     }
 
     res
--- a/tooling/cli/src/profiling_alloc.rs
+++ b/tooling/cli/src/profiling_alloc.rs
@@ -66,11 +66,6 @@ impl ProfilingAllocator {
         self.tracy_depth.store(depth, Ordering::SeqCst);
     }
 
-    #[cfg(feature = "tracy")]
-    pub fn disable_tracy(&self) {
-        self.tracy_enabled.store(false, Ordering::SeqCst);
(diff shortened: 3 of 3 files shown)


tooling/cli/src/profiling_alloc.rs
```diff
@@ -0,0 +1,172 @@
...
+            .current
+            .fetch_add(size, Ordering::SeqCst)
+            .wrapping_add(size);
+        self.max.fetch_max(current, Ordering::SeqCst);
+        self.count.fetch_add(1, Ordering::SeqCst);
+        self.tracy_alloc(size, ptr);
+        ptr
+    }
+
+    unsafe fn realloc(&self, ptr: *mut u8, old_layout: Layout, new_size: usize) -> *mut u8 {
```
Bisht13, line 149: We should be reporting the deallocation as well.

```suggestion
    unsafe fn realloc(&self, ptr: *mut u8, old_layout: Layout, new_size: usize) -> *mut u8 {
    self.tracy_dealloc(ptr);
```

tooling/cli/src/profiling_alloc.rs
```diff
@@ -0,0 +1,172 @@
...
+#[allow(unsafe_code)]
+unsafe impl GlobalAlloc for ProfilingAllocator {
+    unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
+        let ptr = SystemAlloc.alloc(layout);
+        let size = layout.size();
+        let current = self
+            .current
+            .fetch_add(size, Ordering::SeqCst)
+            .wrapping_add(size);
+        self.max.fetch_max(current, Ordering::SeqCst);
```
Bisht13, line 124: Nit, between `fetch_add` and `fetch_max`, another thread could update `max`, causing the local current calculation to be stale, we can avoid this by using `fetch_update`

```suggestion
        self.max.fetch_update(Ordering::SeqCst, Ordering::SeqCst, |max_val| {
    let new_current = self.current.load(Ordering::SeqCst);
    (new_current > max_val).then_some(new_current)
}).ok();
```

tooling/cli/src/cmd/mod.rs
```diff
@@ -15,6 +15,16 @@ pub trait Command {
...
+
+    /// enable Tracy profiling
+    #[cfg(feature = "tracy")]
+    #[argh(switch)]
+    pub tracy: bool,
+
+    /// enable tracy allocation tracking with provided stack depth
+    #[cfg(feature = "tracy")]
+    #[argh(option)]
+    pub tracy_allocations: Option<usize>,
```
Bisht13, line 28: UX improvement

```suggestion
    #[argh(option, default = "0")]
    pub tracy_allocations: usize,
```
and in main.rs

```rs
if args.tracy {
    if args.tracy_allocations > 0 {
        info!("Tracy enabled with allocation tracking (depth {}).", args.tracy_allocations);
        ALLOCATOR.enable_tracy(args.tracy_allocations);
    } else {
        info!("Tracy enabled (no allocation tracking).");
    }
}
```

tooling/cli/src/main.rs
```diff
@@ -1,55 +1,62 @@
...
     let res = args.run();
 
     // SAFETY: It is safe to call `___tracy_shutdown_profiler()` here because all
     // profiling work is complete, and no further Tracy API calls will be made
     // after this point. This function must be called to properly shut down the
     // Tracy profiler before program exit.
-    unsafe {
-        tracy_client_sys::___tracy_shutdown_profiler();
+    //
+    // @recmo: This is not safe. tracing_tracy may still
```
Bisht13: I think to handle this we can do one of the two things,
1. Remove it
Pro: No crashes. Con: Lose last ~100ms of trace data.
2. Grace period (~100 ms)
Pro: Usually works. Con: Still has race condition, just smaller window.

tooling/cli/src/profiling_alloc.rs
recmo: I'm following the example of `tracy-client` here:

https://github.com/nagisa/rust_tracy_client/blob/45cd2681f3083f0bef6fdfb117e575c0905bdf8f/tracy-client/src/lib.rs#L292

It made sense to me since the `ptr` uniquely identifies the allocation until it gets freed. So if the client sees a multiple `alloc` events on the same `ptr` it must mean a realloc. (Whether the Tracy client actually does this, I don't know).

recmo: Ok, the Tracy doc has a different opinion then what the `tracy-cleint` crate does:

Screenshot 2025-10-15 at 16 33 51
Screenshot 2025-10-15 at 16 34 20

recmo: nvm, I'm blind. `tracy-client` does emit a free right here:

https://github.com/nagisa/rust_tracy_client/blob/45cd2681f3083f0bef6fdfb117e575c0905bdf8f/tracy-client/src/lib.rs#L287

tooling/cli/src/profiling_alloc.rs
recmo: Thought about this, it should be fine! The `current` value computed here is always accurate at that sequence point. Some thread might compute an earlier or a later value of `current`, but the `max` operation is a join, so doesn't matter if it gets sequenced as 

```
self.max.fetch_max(current_old, Ordering::SeqCst);
self.max.fetch_max(current_new, Ordering::SeqCst);
```

```
self.max.fetch_max(current_new, Ordering::SeqCst);
self.max.fetch_max(current_old, Ordering::SeqCst);
```

The end result is going to be that `max` contains `max(old_max, current_old, current_new)`. Similarly if there are more parallel threads.


The one exception to this may be if a concurrent thread resets the maximum.

tooling/cli/src/cmd/mod.rs
recmo: There's actually a difference between 'off' and 'zero depth'. With zero depth we still collect allocation events, but we do not include stack traces (in fact, there is special code in the Allocator to handle this). With 'off' we do not collect allocation events in Tracy at all.

tooling/cli/src/main.rs
recmo: > Con: Lose last ~100ms of trace data.

It's worse unfortunately. There could be an unbounded backlog of data still to be transmitted. This actually happened all the time before this PR when we did 100-depth stacktraces on each allocation (times 90M allocations is gigabytes of data to transfer over what seems a very poorly implemented IPC bus).

 I checked, but there doesn't seem to be a 'flush' function that keeps the Tracy client alive, but clears the backlog.

recmo: Screenshot 2025-10-15 at 16 54 46
Screenshot 2025-10-15 at 16 55 09

recmo: I guess the best solution here is to manually implement `TRACY_NO_EXIT` with something like "Keeping alive for Tracy connections. Press any key to exit."

Bisht13:
LGTM

Bisht13 (approved):

PR #218, opened 2025-10-14
Upgrade to Noir beta14

recmo:
~~`acir_field` is forked to add a zero-copy serde implementation. It retains compatibility with the original.~~

~~It would be more performant if we stored as little-endian Montgomery reduced form instead, but this is a format breaking change. To support both formats would require a hack, as we can't pass configuration down easily.~~


Bisht13:
Replaced with #349

PR #225, opened 2025-10-30
refactor(cli): enhance circuit_stats with exact R1CS breakdown tracking

Bisht13:
### Summary
Refactored `circuit_stats` into a clean modular structure and instrumented the R1CS compiler to provide **exact** cost breakdowns for batched operations (AND/XOR/RANGE).

### DIFF
--- a/provekit/r1cs-compiler/src/noir_to_r1cs.rs
+++ b/provekit/r1cs-compiler/src/noir_to_r1cs.rs
@@ -312,11 +312,16 @@ impl NoirToR1CSCompiler {
         &mut self,
         circuit: &Circuit<NoirElement>,
     ) -> Result<R1CSBreakdown> {
-        // Memory blocks for building memory lookup constraints at the end
+        // Read-only memory blocks (used for building the memory lookup constraints at
+        // the end)
         let mut memory_blocks: BTreeMap<usize, MemoryBlock> = BTreeMap::new();
-        // Map from range size k to witnesses to be constrained within [0..2^k]
+        // Mapping the log of the range size k to the vector of witness indices that
+        // are to be constrained within the range [0..2^k].
+        // These will be digitally decomposed into smaller ranges, if necessary.
         let mut range_checks: BTreeMap<u32, Vec<usize>> = BTreeMap::new();
-        // (input, input, output) tuples for batched AND/XOR operations
+        // (input, input, output) tuples for AND and XOR operations.
+        // Inputs may be either constants or R1CS witnesses.
+        // Outputs are always R1CS witnesses.
         let mut and_ops = vec![];
         let mut xor_ops = vec![];
         let mut sha256_compression_ops = vec![];
@@ -330,7 +335,7 @@ impl NoirToR1CSCompiler {
         for opcode in &circuit.opcodes {
             match opcode {
                 Opcode::AssertZero(expr) => self.add_acir_assert_zero(expr),
-                // Brillig is only for witness generation and does not produce constraints
+                // Brillig is only for witness generation and does not produce constraints.
                 Opcode::BrilligCall { .. } => {}
                 Opcode::MemoryInit {
                     block_id,
@@ -360,7 +365,10 @@ impl NoirToR1CSCompiler {
                     op,
                     predicate,
                 } => {
+                    // Panic if the predicate is set (according to Noir developers, predicate is
+                    // always None and will soon be removed).
                     assert!(predicate.is_none());
+
                     let block_id = block_id.0 as usize;
                     assert!(
                         memory_blocks.contains_key(&block_id),
@@ -368,6 +376,12 @@ impl NoirToR1CSCompiler {
                         block_id
                     );
                     let block = memory_blocks.get_mut(&block_id).unwrap();
+
+                    // `op.index` is _always_ just a single ACIR witness, not a more complicated
+                    // expression, and not a constant. See [here](https://discord.com/channels/1113924620781883405/1356865341065531446)
+                    // Static reads are hard-wired into the circuit, or instead rendered as a
+                    // dummy dynamic read by introducing a new witness constrained to have the value
+                    // of the static address.
                     let addr = op.index.to_witness().map_or_else(
                         || {
                             unimplemented!(
@@ -378,6 +392,14 @@ impl NoirToR1CSCompiler {
                         |acir_witness| self.fetch_r1cs_witness_index(acir_witness),
                     );
                     let op = if op.operation.is_zero() {
+                        // Create a new (as yet unconstrained) witness `result_of_read` for the
+                        // result of the read; it will be constrained by later memory block
+                        // processing.
+                        // "In read operations, [op.value] corresponds to the witness index at which
+                        // the value from memory will be written." (from the Noir codebase)
+                        // At R1CS solving time, only need to map over the value of the
+                        // corresponding ACIR witness, whose value is already determined by the ACIR
+                        // solver.
                         let result_of_read =
                             self.fetch_r1cs_witness_index(op.value.to_witness().unwrap());
                         MemoryOperation::Load(addr, result_of_read)
@@ -410,6 +432,11 @@ impl NoirToR1CSCompiler {
                             .or_default()
                             .push(input_witness);
                     }
+
+                    // Binary operations:
+                    // The inputs and outputs will have already been solved for by the ACIR solver.
+                    // Collect the R1CS witnesses indices so that we can later constrain them
+                    // appropriately.
                     BlackBoxFuncCall::AND { lhs, rhs, output } => {
                         and_ops.push((
                             self.fetch_constant_or_r1cs_witness(lhs.input()),
@@ -432,10 +459,14 @@ impl NoirToR1CSCompiler {
                         assert_eq!(inputs.len() as u32, *len, "Poseidon2: inputs.len != len");
                         assert_eq!(outputs.len() as u32, *len, "Poseidon2: outputs.len != len");
                         let t = *len;
+
+                        // Only these widths are allowed for Poseidon2
                         assert!(
                             matches!(t, 2 | 3 | 4 | 8 | 12 | 16),
                             "Poseidon2: unsupported width {t}"
                         );
+
+                        // Convert ACIR inputs to (Constant | Witness)
                         let in_wits: Vec<ConstantOrR1CSWitness> = inputs
                             .iter()
                             .map(|inp| self.fetch_constant_or_r1cs_witness(inp.input()))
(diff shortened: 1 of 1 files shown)


tooling/cli/src/cmd/circuit_stats/stats_collector.rs
```diff
@@ -0,0 +1,277 @@
...
+            Circuit, Opcode,
+        },
+        native_types::Expression,
+        FieldElement,
+    },
+    std::collections::{HashMap, HashSet},
+};
+
+/// R1CS complexity constants for black box functions (measured empirically).
+pub(super) const SHA256_COMPRESSION_CONSTRAINTS: usize = 31_264;
```
0xVikasRushi: remove unused first two constants

tooling/cli/src/cmd/circuit_stats/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+#[argh(subcommand, name = "circuit_stats")]
+pub struct Args {
+    /// path to the ACIR circuit file (.json)
+    #[argh(positional)]
+    circuit_path: PathBuf,
+
+    /// path to witness file (reserved for future use)
+    #[argh(positional)]
+    witness_path: PathBuf,
+}
```
0xVikasRushi, line 37: add example command for circuit_stats cli usage

 `cargo run --release --bin provekit-cli circuit_stats <circuits.json> <Prover.toml>`

0xVikasRushi (changes requested):

provekit/r1cs-compiler/src/noir_to_r1cs.rs
```diff
@@ -40,8 +91,6 @@ pub(crate) struct NoirToR1CSCompiler {
     pub initial_memories: BTreeMap<usize, Vec<usize>>,
 }
 
-/// Compile a Noir circuit to a R1CS relation, returning the R1CS and a map from
```
rose2221, line 43: **Nit:** The removed doc comment explained the purpose of noir_to_r1cs and its return values. Any reason it was dropped?

provekit/r1cs-compiler/src/noir_to_r1cs.rs
```diff
@@ -230,30 +293,31 @@ impl NoirToR1CSCompiler {
...
+        Ok(())
+    }
+
+    fn add_circuit_with_breakdown(
+        &mut self,
+        circuit: &Circuit<NoirElement>,
+    ) -> Result<R1CSBreakdown> {
         let mut memory_blocks: BTreeMap<usize, MemoryBlock> = BTreeMap::new();
-        // Mapping the log of the range size k to the vector of witness indices that
-        // are to be constrained within the range [0..2^k].
```
rose2221, line 238: The refactor removed several useful inline comments from `add_circuit`. Could these comments be preserved in `add_circuit_with_breakdown`?

provekit/r1cs-compiler/src/noir_to_r1cs.rs
```diff
@@ -463,15 +512,38 @@ impl NoirToR1CSCompiler {
...
-        // For the AND and XOR operations, add the appropriate constraints.
+        breakdown.sha256_direct_constraints =
+            self.r1cs.num_constraints() - constraints_before_sha256;
+        breakdown.sha256_direct_witnesses = self.num_witnesses() - witnesses_before_sha256;
+        breakdown.sha256_and_ops = and_ops.len() - and_ops_before;
+        breakdown.sha256_xor_ops = xor_ops.len() - xor_ops_before;
+        breakdown.sha256_range_ops =
+            range_checks.values().map(|v| v.len()).sum::<usize>() - range_ops_before;
+
+        let constraints_before = self.r1cs.num_constraints();
```
rose2221: **Nit:** Minor readability thing `constraints_before `/ `witnesses_before` are shadowed here, while earlier code uses more explicit names. Might be worth keeping the naming consistent for readability.

provekit/r1cs-compiler/src/noir_to_r1cs.rs
```diff
@@ -427,34 +455,55 @@ impl NoirToR1CSCompiler {
...
-        memory_blocks.iter().for_each(|(_, block)| {
+        breakdown.assert_zero_constraints = self.r1cs.num_constraints() - constraints_before_assert;
+        breakdown.assert_zero_witnesses = self.num_witnesses() - witnesses_before_assert;
+
+        let constraints_before_rom = self.r1cs.num_constraints();
+        let witnesses_before_rom = self.num_witnesses();
+        let mut constraints_before_ram = 0;
+        let mut witnesses_before_ram = 0;
+
+        for (_, block) in memory_blocks.iter() {
```
rose2221, line 517: This logic seems to rely on ROM blocks appearing before RAM blocks during `memory_blocks.iter()`. Is this guaranteed by how the keys are assigned? Otherwise, interleaving could affect the breakdown.

tooling/cli/src/cmd/circuit_stats/display.rs
```diff
@@ -0,0 +1,461 @@
+use {
+    super::{memory::describe_block_type, stats_collector::CircuitStats},
```
rose2221, line 4: **Nit:** This file doesn’t have a module-level doc comment explaining its purpose.

tooling/cli/src/cmd/circuit_stats/memory.rs
```diff
@@ -0,0 +1,186 @@
...
+    pub fn is_read_only(&self) -> bool {
+        self.writes == 0
+    }
+
+    fn total_ops(&self) -> usize {
+        self.reads + self.writes
+    }
+
+    /// R1CS constraint count for RAM (read-write memory).
+    pub fn ram_constraint_count(&self) -> usize {
```
rose2221, line 54: **Nit:** The R1CS cost formulas  (`ram_constraint_count`, `rom_constraint_count`etc.)  here use magic constants. A brief comment on where these numbers come from (or a reference to the memory-checking logic) would help

tooling/cli/src/cmd/circuit_stats/display.rs
```diff
@@ -0,0 +1,461 @@
...
+        total_batched_constraints, total_batched_witnesses
+    );
+    println!("└{}", SUBSECTION);
+
+    if sha256_count > 0 {
+        let sha256_direct = breakdown.sha256_direct_constraints;
+        let sha256_batched =
+            breakdown.and_constraints + breakdown.xor_constraints + breakdown.range_constraints;
+        let sha256_total_constraints = sha256_direct + sha256_batched;
+        let per_sha256 = sha256_total_constraints / sha256_count;
```
rose2221: The per-compression cost (`sha256_total `/ `sha256_count`) assumes a purely linear SHA256 cost model, but the formula includes a fixed one-time setup cost. This inflates the per-SHA estimate. The same applies to the witness count.

tooling/cli/src/cmd/circuit_stats/display.rs
```diff
@@ -0,0 +1,461 @@
...
+
+        let sha256_direct_w = breakdown.sha256_direct_witnesses;
+        let sha256_batched_w =
```
rose2221: The SHA total cost summary seems to include all batched RANGE constraints. Are we sure `complete_age_check` doesn’t have explicit non-SHA range checks that would get counted here?

provekit/r1cs-compiler/src/noir_to_r1cs.rs
Bisht13: Restored. Added back the doc comment explaining the function's purpose and return values.

provekit/r1cs-compiler/src/noir_to_r1cs.rs
Bisht13: Restored key comments explaining:
- Memory blocks for deferred lookup constraints
- Range check bit-size mapping  
- AND/XOR tuple structure
- Brillig being witness-generation-only

provekit/r1cs-compiler/src/noir_to_r1cs.rs
Bisht13: Separating into two distinct loops, first processing all ROM blocks, then all RAM blocks. This ensures accurate breakdown tracking regardless of block_id ordering in the BTreeMap.

tooling/cli/src/cmd/circuit_stats/memory.rs
Bisht13: Added in function comment.

provekit/r1cs-compiler/src/noir_to_r1cs.rs
Bisht13: Renamed to descriptive names.

tooling/cli/src/cmd/circuit_stats/display.rs
Bisht13: Added and_ops_total, xor_ops_total, and range_ops_total fields to R1CSBreakdown. The SHA256 cost summary now computes proportional attribution:
- sha256_and_constraints = (and_constraints * sha256_and_ops) / and_ops_total
- This correctly attributes only SHA256's share of batched operations, not all of them.

tooling/cli/src/cmd/circuit_stats/display.rs
Bisht13: Yes, there's a one-time lookup table setup cost. The AND/XOR operations use a shared lookup table of size 2^16 (65,536 entries for 8-bit atomic operations). This table is constructed once for all AND ops and once for all XOR ops.

The proportional attribution handles this: SHA256's share of the table cost is proportional to its share of operations (sha256_and_ops / and_ops_total). This means the "per compression" estimate includes SHA256's fair share of the amortized table overhead, which is the correct behavior since the table is shared across all operations.

The estimate will be slightly higher than the marginal cost of adding one more SHA256 compression (since it includes amortized table cost), but it accurately reflects the total cost attributable to SHA256.

rose2221: Good point. The current "per compression" cost is the amortized cost (total / count), which includes SHA256's proportional share of the one-time lookup table setup. This means the per-compression number will vary depending on how many SHA256 calls are in the circuit  it's higher with fewer calls and decreases as calls increase.
While the amortized cost correctly reflects SHA256's total cost attribution, it's less useful for:

- **Benchmarking** - comparing SHA256 costs across different circuits becomes inconsistent

- **Predicting incremental cost** - knowing "what does adding one more SHA256 call cost?"

**Suggestion:** We could add a separate marginal cost field that represents the cost of adding one additional SHA256 call, excluding the one-time table overhead.

Bisht13: Valid point. The marginal cost metric would be useful for benchmarking and predicting incremental costs.

However, this is additive scope, the current implementation correctly shows total cost attribution, which is what the original refactor aimed to provide.

I'd suggest we merge this PR as-is and open a follow-up issue for "Add marginal cost estimates for SHA256/Poseidon operations" to track this enhancement. That way we can properly design how to separate table overhead from per-operation costs. WDYT?

rose2221: Sounds good to me.

rose2221 (approved):

Issue #212, opened 2025-10-10
`ntt/src/ntt`: Replace by parallel alternative to speed up tests


recmo:
*On 2025-10-10 in [`38c724d`](https://github.com/worldfnd/provekit/commit/38c724dced40e87ce32b55778b3a3870dd12e0e1) “Merge pull request #204 from worldfnd/xr/interleaved-ntt”:*

Replace by parallel alternative to speed up tests

```rust

            prop_assert_eq!(original,s);
        }
    }

    // TODO Replace by parallel alternative to speed up tests
    fn transpose<T: Copy>(matrix: &[T], rows: usize, columns: usize) -> Vec<T> {
        assert_eq!(matrix.len(), rows * columns);
        let mut v = Vec::with_capacity(matrix.len());
        for j in 0..columns {
            for i in 0..rows {

```
*From [`ntt/src/ntt.rs:434`](https://github.com/worldfnd/provekit/blob/38c724dced40e87ce32b55778b3a3870dd12e0e1/ntt/src/ntt.rs#L434)*

xrvdg:
So far this hasn't turned out to be a bottleneck.

PR #226, opened 2025-10-30
 add parallel tiled transpose implementation

Vishalkulkarni45:
## Description

This PR closes issue #212  by implementing a parallelized matrix transpose operation.

## Implementation

The solution splits large matrices into smaller tiles and transposes each tile in a separate thread

## Performance

Benchmarks were run on **MacBook M4 Pro** (14-core CPU, 48GB RAM) with n×n matrices where n ranges from 2^6 to 2^12.

photo_2025-10-31_00-28-12

**Results:**
- `transpose`: 86.66ms (slowest)
- `transpose_tiled`: 25.61ms (slowest)
- **Improvement: 70.4% faster** (~3.4x speedup)

[Benchmark code](https://github.com/Vishalkulkarni45/provekit/blob/bench/transpose_tiled/ntt/benches/transpose.rs)

---

Let me know if any changes are needed!

### DIFF
--- a/ntt/src/ntt.rs
+++ b/ntt/src/ntt.rs
@@ -362,6 +362,7 @@ mod tests {
         ark_bn254::Fr,
         ark_ff::BigInt,
         proptest::collection,
+        rayon::iter::{IntoParallelIterator, ParallelIterator},
         std::{
             fmt,
             num::{NonZero, NonZeroUsize},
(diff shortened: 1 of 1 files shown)


ntt/src/ntt.rs
```diff
@@ -431,21 +432,44 @@ mod tests {
...
-            for i in 0..rows {
-                v.push(matrix[i * columns + j]);
-            }
-        }
-        v
+    fn transpose<T: Copy + Send + Sync>(matrix: &[T], rows: usize, columns: usize) -> Vec<T> {
+        let len = rows * columns;
+        assert_eq!(matrix.len(), len);
+
+        const TILE_SIZE: usize = 64;
```
xrvdg, line 439: Tile size should be a function of the cache size and the size of the element. A function like `workload_size<F>` that we use in the codebase helps in formulating an equation.

xrvdg:
Thank you for looking into this.

In our case, it's not just the raw speedup that's important, but also the parallel efficiency of a core. We need to consider both, especially on our low-core-count target devices where low efficiency can hurt overall system throughput.

In this case, the efficiency is approx. 24% (from 3.4x speedup / 14 cores), which is unfortunately below our target (ideally >80%).

Because this approach doesn't meet our efficiency requirements and transpose hasn't been a bottleneck in our tests, I will close this PR and the corresponding issue.

PR #235, opened 2025-11-11
Modular NTT in WHIR v1b

xrvdg:


provekit/r1cs-compiler/src/ntt.rs
```diff
@@ -0,0 +1,35 @@
...
+    ) -> Vec<Fr> {
+        interleaved_rs_encode(interleaved_coeffs, expansion, fold_factor)
+    }
+}
+
+fn interleaved_rs_encode(
+    interleaved_coeffs: &[Fr],
+    expansion: usize,
+    fold_factor: usize,
+) -> Vec<Fr> {
```
Bisht13, line 21: `interleaved_coeffs.len().is_power_of_two()` is not checked.

Bisht13, line 21: `expansion > 0` is not checked.

ntt/src/ntt.rs
```diff
@@ -136,40 +130,39 @@ impl Default for NTTEngine {
...
     // The order of the interleaved NTTs themselves
-    let order = n / *num_of_polys;
+    let order = values.order().0;
 
     // This conditional is here because chunk_size for *chunk_exact_mut can't be 0
     if order <= 1 {
         return;
     }
 
+    let number_of_polyes = n / order;
```
Bisht13: Nit: typo, change to `number_of_polys`

provekit/r1cs-compiler/src/ntt.rs
xrvdg: Added a debug assertion to conform to the implementation in WHIR.

xrvdg: This is done as part of NTT::new at line 29. It handles the more general case and in this context comes down to checking `interleaved_coeffs.len().is_power_of_two()`.

xrvdg:
> Given that we want to use the NTT crate from ProveKit repo in WHIR which in turn is being used in provekit-cli crate, I think the changes in `common/` should be in the NTT crate.

That would make the NTT crate dependent on WHIR which must not happen. I've moved this to `provekit/r1cs-compiler`, which is where it will be used.

PR #236, opened 2025-11-11
RS reverse ordered + custom arkworks

xrvdg:


recmo:
NTT works should move to the WHIR repo.

PR #240, opened 2025-12-04
Feat: Witness splitting for sound challenge generation for LogUp/Spice

shreyas-londhe:
## Description

This PR implements witness-splitting to enable sound challenge generation for LogUp/Spice sub-protocols. A detailed spec of the problem is linked [here](https://hackmd.io/@shreyas-londhe/HkgVaTXCxx).

### Witness Splitting & IOPattern
- Split Witness: Refactored `WhirR1CSProver` to handle `w1` (pre-challenge) and `w2` (post-challenge) witnesses separately.
- Transcript Sync: Updated `IOPattern` to correctly inject `add_logup_challenges` between the two witness commitments.
- Prover Flow: Updated `prove()` to accept two separate commitments (`commitment_1` and `commitment_2`) and correctly slice the witness.

## Verification
Verified full proof generation and verification flow with `provekit-cli`.

### DIFF
--- a/provekit/common/src/witness/scheduling/splitter.rs
+++ b/provekit/common/src/witness/scheduling/splitter.rs
@@ -48,29 +48,29 @@ impl<'a> WitnessSplitter<'a> {
         // Also collect lookup builders (direct challenge consumers)
         let mut mandatory_w2 = challenge_builders.clone();
         let mut lookup_builders = HashSet::new();
-        let mut visited = vec![false; builder_count];
-        let mut stack = VecDeque::new();
+        let mut forward_visited = vec![false; builder_count];
+        let mut forward_stack = VecDeque::new();
 
         for &challenge_idx in &challenge_builders {
-            visited[challenge_idx] = true;
+            forward_visited[challenge_idx] = true;
             // Collect direct consumers as lookup builders
             for &consumer_idx in &self.deps.adjacency_list[challenge_idx] {
                 lookup_builders.insert(consumer_idx);
-                if !visited[consumer_idx] {
-                    visited[consumer_idx] = true;
+                if !forward_visited[consumer_idx] {
+                    forward_visited[consumer_idx] = true;
                     mandatory_w2.insert(consumer_idx);
-                    stack.push_back(consumer_idx);
+                    forward_stack.push_back(consumer_idx);
                 }
             }
         }
 
         // Continue DFS to find all transitive dependents
-        while let Some(current_idx) = stack.pop_front() {
+        while let Some(current_idx) = forward_stack.pop_front() {
             for &consumer_idx in &self.deps.adjacency_list[current_idx] {
-                if !visited[consumer_idx] {
-                    visited[consumer_idx] = true;
+                if !forward_visited[consumer_idx] {
+                    forward_visited[consumer_idx] = true;
                     mandatory_w2.insert(consumer_idx);
-                    stack.push_back(consumer_idx);
+                    forward_stack.push_back(consumer_idx);
                 }
             }
         }
@@ -79,18 +79,18 @@ impl<'a> WitnessSplitter<'a> {
         // (exclude anything in mandatory_w2 to maintain disjoint sets)
         let witness_producer = &self.deps.witness_producer;
         let mut mandatory_w1 = HashSet::new();
-        let mut visited = vec![false; builder_count];
-        let mut stack = VecDeque::new();
+        let mut backward_visited = vec![false; builder_count];
+        let mut backward_stack = VecDeque::new();
 
         for &lookup_idx in &lookup_builders {
-            stack.push_back(lookup_idx);
+            backward_stack.push_back(lookup_idx);
         }
 
-        while let Some(current_idx) = stack.pop_front() {
-            if visited[current_idx] {
+        while let Some(current_idx) = backward_stack.pop_front() {
+            if backward_visited[current_idx] {
                 continue;
             }
-            visited[current_idx] = true;
+            backward_visited[current_idx] = true;
 
             // Only add to w1 if not in mandatory_w2 (maintain disjoint)
             if !mandatory_w2.contains(&current_idx)
@@ -102,8 +102,8 @@ impl<'a> WitnessSplitter<'a> {
 
             for &witness_idx in &self.deps.reads[current_idx] {
                 if let Some(&producer_idx) = witness_producer.get(&witness_idx) {
-                    if !visited[producer_idx] && !mandatory_w2.contains(&producer_idx) {
-                        stack.push_back(producer_idx);
+                    if !backward_visited[producer_idx] && !mandatory_w2.contains(&producer_idx) {
+                        backward_stack.push_back(producer_idx);
                     }
                 }
             }


provekit/common/src/witness/scheduling/remapper.rs
```diff
@@ -0,0 +1,289 @@
...
+    ///
+    /// The map goes from ACIR witness index -> R1CS witness index.
+    /// We need to update the R1CS indices to their new remapped values.
+    pub fn remap_acir_witness_map(&self, map: Vec<Option<NonZeroU32>>) -> Vec<Option<NonZeroU32>> {
+        map.into_iter()
+            .map(|opt_idx| {
+                opt_idx.map(|idx| {
+                    let old_r1cs_idx = idx.get() as usize;
+                    let new_r1cs_idx = self.remap(old_r1cs_idx);
+                    NonZeroU32::new(new_r1cs_idx as u32).expect("Remapped index should be non-zero")
```
Bisht13, line 284: Can `new_r1cs_idx` be 0? If it is then this will panic.

provekit/prover/src/lib.rs
```diff
@@ -73,75 +57,70 @@ impl Prove for Prover {
...
+        self.r1cs.solve_witness_vec(
+            &mut witness,
+            self.split_witness_builders.w1_layers,
+            &acir_witness_idx_to_value_map,
+            &mut merlin,
         );
-        let witness = fill_witness(partial_witness).context("while filling witness")?;
 
-        // Verify witness (redudant with solve)
+        let w1 = witness[..self.whir_for_witness.w1_size]
```
Bisht13, line 75: Can clear witness' processed segment (w1), else this is persisting in memory.

recursive-verifier/app/circuit/circuit_test.go
```diff
@@ -0,0 +1,415 @@
...
+					)
+					if err != nil {
+						return nil, nil, err
+					}
+					stirAnswers = append(stirAnswers, stirAnswersTemporary)
+				}
+
+			case "merkle_proof":
+				var path FullMultiPath[KeccakDigest]
+				_, _ = arkSerialize.CanonicalDeserializeWithMode(
```
Bisht13: Several parsing operations like this are silently ignoring errors

.github/workflows/ci.yml
```diff
@@ -31,9 +31,8 @@ jobs:
       - run: cargo fmt --all --check
       - run: cargo clippy --all-targets --all-features --verbose
       - run: cargo build --all-targets --all-features --verbose
-      - run: cargo test --no-fail-fast --all-targets --all-features --verbose
+      - run: cargo test --no-fail-fast --all-features --verbose --lib --tests --bins
```
Bisht13, line 34: Why are we skipping testing benchmarks?

provekit/common/src/sparse_matrix.rs
```diff
@@ -121,6 +122,42 @@ impl SparseMatrix {
...
+
+            let row_cols = &mut self.col_indices[start..end];
+            let row_vals = &mut self.values[start..end];
+
+            let mut pairs: Vec<_> = row_cols
+                .iter()
+                .zip(row_vals.iter())
+                .map(|(&c, &v)| (c, v))
+                .collect();
+            pairs.sort_unstable_by_key(|(c, _)| *c);
```
Bisht13, line 153: Sequential row sorting after parallel column remapping may become a bottleneck for large matrices. Maybe consider parallel row sorting?

provekit/common/src/witness/scheduling/splitter.rs
```diff
@@ -0,0 +1,169 @@
...
+                if let Some(&producer_idx) = witness_producer.get(&witness_idx) {
+                    if !visited[producer_idx] && !mandatory_w2.contains(&producer_idx) {
+                        stack.push_back(producer_idx);
+                    }
+                }
+            }
+        }
+
+        // Step 5: Identify free builders (not in either mandatory set)
+        let mut free_builders = Vec::new();
```
Bisht13, line 117: What happens if there are not free builders? If free builders are rare, the balancing logic is essentially dead code and could be simplified.

provekit/common/src/witness/scheduling/remapper.rs
shreyas-londhe: No, ACIR witnesses should never be reammped to index 0 as it is reserved for `witness_one`. This was an implicit rule earlier but I pushed a fix which should make this explicit.

.github/workflows/ci.yml
shreyas-londhe: Benchmarks are already run in `benchmark.yml` where the artifacts are built from scratch. Running benchmarks as tests fails because the test does not generate the artifact.

provekit/common/src/witness/scheduling/splitter.rs
shreyas-londhe: Free builders are essentially builders which do not depend on range checks and mem ops. There are quite a few builders which become free builders while splitting. Also if there are no free builders, the loop is skipped, but this case needs to be handled.

provekit/common/src/sparse_matrix.rs
shreyas-londhe: The per-row sort is O(k log k) where k is typically small (<100 elements). Parallelization adds overhead that likely exceeds gains for sparse matrices. Will profile and revisit if this shows up as a hotspot.

provekit/prover/src/lib.rs
shreyas-londhe: `w1` is consumed by `commit` below.

recursive-verifier/app/circuit/common.go
```diff
@@ -121,15 +167,44 @@ func PrepareAndVerifyCircuit(config Config, r1cs R1CS, pk *groth16.ProvingKey, v
...
-	var witnessData = consumeWhirData(config.WHIRConfigWitness, &merklePaths, &stirAnswers)
+	if config.NumChallenges > 0 {
+		// Batch mode: N commitments
+		// Rust emits: N first-round hints, then NRounds hints for batched polynomial
+		var numCommitments int
+		if config.NumChallenges > 0 {
+			numCommitments = 2
+		} else {
+			numCommitments = 1
+		}
```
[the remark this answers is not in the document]
shreyas-londhe: This is not necessary for our use case.

tooling/provekit-bench/Cargo.toml
```diff
@@ -34,4 +34,4 @@ workspace = true
 
 [[bench]]
 name = "bench"
-harness = true
\ No newline at end of file
+harness = false
```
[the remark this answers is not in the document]
shreyas-londhe: This isn't a change — harness = false is required for divan (our benchmark framework) to function. Custom harnesses like divan/criterion need to provide their own main(), which is only possible with harness = false. This is standard configuration, not a breaking change.
If this file previously had harness = true, the benchmarks wouldn't have compiled/run at all with divan.

provekit/prover/src/lib.rs
Bisht13: I meant the original witness vector still holds the w1 data until drop(witness) on line 115, you can `take` instead.

shreyas-londhe: Can't use `.take()` here — when solving w2, the solver may need to read w1 witnesses from the `witness` vector to compute w2 values (w2 witnesses can depend on w1 witnesses). Taking would leave `None` in those slots and break the solve.

Bisht13 (approved):

PR #244, opened 2025-12-17
M31 support

xrvdg:
This PR adds basic support for M31 (addition, subtraction, multiplication and shift left). This implementation differs from Plonky, STWO and Icicle in that it uses the correspondence between Mersenne numbers and one complement representation. This correspondence allows us to get rid of the asymmetry between addition and subtraction and thereby making effective use of 64 bit registers. With this PR intermediate results can accumulate to fill a 62 bit signed integer before needing reduction.


recmo:
This might become relevant again, but needs to develop into more than just and experiment to be mergeable.

recmo (changes requested):
This might become relevant again if we decide to pick op M31, but then:

1) it needs to develop into more than just and experiment to be mergeable,
2) it needs to demonstrate a meaningful improvement over a SotA implementation like Stwo.

PR #246, opened 2025-12-19
feat: add WASM support for browser-based proof generation

Bisht13:
## Summary
- Add wasm32 architecture support to Skyscraper field arithmetic
- Add WASM bindings for Prover and Verifier (`provekit-wasm`)
- Add Node.js and browser demo with noir_js integration

## Changes
- **skyscraper**: wasm32 block multiplier, fp-rounding stubs, SIMD codegen
- **provekit-wasm**: `Prover` and `Verifier` classes with `proveBytes()` API
- **playground/wasm-node-demo**: Setup scripts, browser UI, Node.js CLI demo

## Usage
cd playground/wasm-node-demo
pnpm install
pnpm run setup -- path/to/noir-circuit
pnpm run demo        # Node.js
pnpm run demo:web    # Browser (localhost:8080)

Blocked on #198

### DIFF
--- a/tooling/provekit-wasm/src/format.rs
+++ b/tooling/provekit-wasm/src/format.rs
@@ -59,7 +59,7 @@ fn decompress_impl(data: &[u8]) -> Result<Vec<u8>, String> {
     if data.len() >= 4 && data[..4] == ZSTD_MAGIC {
         let mut decoder = ruzstd::decoding::StreamingDecoder::new(std::io::Cursor::new(data))
             .map_err(|e| format!("Failed to init Zstd decoder: {e}"))?;
-        let hint = decoder.decoder.content_size() as usize;
+        let hint = usize::try_from(decoder.decoder.content_size()).unwrap_or(0);
         let mut out = Vec::with_capacity(hint);
         std::io::Read::read_to_end(&mut decoder, &mut out)
             .map_err(|e| format!("Failed to decompress Zstd data: {e}"))?;
--- a/tooling/provekit-wasm/src/prover.rs
+++ b/tooling/provekit-wasm/src/prover.rs
@@ -70,8 +70,9 @@ impl Prover {
     pub fn get_circuit(&self) -> Result<Box<[u8]>, JsError> {
         let noir_prover = match self.inner_ref()? {
             ProverCore::Noir(p) => p,
-            #[allow(unreachable_patterns)]
-            _ => return Err(JsError::new("Only Noir provers are supported in WASM")),
+            ProverCore::Mavros(_) => {
+                return Err(JsError::new("Only Noir provers are supported in WASM"))
+            }
         };
 
         let program_bytes = Program::<NoirElement>::serialize_program(&noir_prover.program);


tooling/provekit-wasm/src/verifier.rs
```diff
@@ -0,0 +1,60 @@
...
+        } else {
+            serde_json::from_slice(verifier_data)
+                .map_err(|err| JsError::new(&format!("Failed to parse verifier JSON: {err}")))?
+        };
+        Ok(Self { inner })
+    }
+
+    /// Verifies a proof provided as JSON bytes.
+    #[wasm_bindgen(js_name = verifyBytes)]
+    pub fn verify_bytes(&mut self, proof_json: &[u8]) -> Result<(), JsError> {
```
ocdbytes: The verify call here doesn't look correct to me. Here verify function is there and it calls inner verify then after one call it is consumed then If I call verify again on the same proof or some other proof for the same circuit it will error out. The demo bypasses it by creating a new verifier instance after each verifier call. 

the correct way would be to have verify call take a reference instead of a mutable reference and remove the `.take()` call. This should eliminate this behaviour.

playground/wasm-demo/scripts/setup.mjs
```diff
@@ -0,0 +1,534 @@
+#!/usr/bin/env node
```
ocdbytes, line 1: When I run the setup and then serve the server in a fresh clone then I would not be able to test it because the website
  errors out saying that

  console error:
```
  http://localhost:8080/vendor/noirc_abi/noirc_abi_wasm_bg.wasm 404 (Not Found)
```

The vendor/ directory has the JS wrappers committed but the .wasm binaries are gitignored. The setup script never runs npm install or copies these from node_modules/@noir-lang/*/web/ into vendor/. Also @noir-lang/acvm_js is missing from package.json (only pulled in as a transitive dep of noir_js).

The setup script should install and setup any vendor dependency for me by design.

tooling/provekit-wasm/src/verifier.rs
ocdbytes: Current behaviour in browser : 

Image

tooling/provekit-wasm/src/format.rs
```diff
@@ -0,0 +1,90 @@
...
+        return Err(JsError::new(&format!(
+            "Incompatible {label} format: minor version {minor}, expected >= {min_minor}"
+        )));
+    }
+
+    Ok(&data[HEADER_SIZE..])
+}
+
+/// Auto-detect compression (Zstd or XZ) and decompress.
+fn decompress(data: &[u8]) -> Result<Vec<u8>, JsError> {
```
ocdbytes: Vec::new() for decompression means the buffer grows by repeated doubling + copy. In WASM (where linear memory can only grow, never shrink), this peaks at ~2x the decompressed size during the final realloc.

Zstd includes the decompressed size and ruzstd exposes it via decoder.decoder.content_size() after construction. Pre-allocating with Vec::with_capacity(content_size) would eliminate all realloc copies. Worth doing for large circuits where the decompressed .pkp can be 100MB+.

Opt would look something like this : 

```rust
  let mut decoder = ruzstd::decoding::StreamingDecoder::new(std::io::Cursor::new(data))?;
  let hint = decoder.decoder.content_size() as usize;
  let mut out = Vec::with_capacity(hint);
  std::io::Read::read_to_end(&mut decoder, &mut out)?;
```

But right now zstd is not being used for pkp files purely for the file size difference. but zstd should be able to decompress the files faster compared to xz. 

- lzma-rs (pure Rust XZ) is significantly slower to decompress than ruzstd, especially in WASM
- XZ doesn't expose decompressed size in the header, so the decompression buffer can't be pre-allocated it grows by repeated doubling, which is costly in WASM's grow-only linear memory
- .pkp is the largest artifact and the one users wait on in the browser


This is not that much related to this PR but just flagging it. was looking into opt then went further to see why the difference.

noir-examples/noir_sha256/src/main.nr
```diff
@@ -1,16 +1,7 @@
...
-fn main() {
-    let bytes = [
-        253, 15, 121, 52, 86, 78, 18, 209, 88, 183, 177, 111, 131, 205, 215, 147, 232, 192, 101, 55,
-        109, 129, 230, 104, 73, 104, 23, 12, 52, 30, 149, 217,
-    ];
-    let result = [
-        131, 243, 207, 3, 17, 244, 189, 0, 155, 99, 191, 153, 41, 237, 133, 132, 66, 118, 103, 215,
-        34, 23, 130, 78, 18, 170, 104, 180, 29, 124, 242, 128,
-    ];
+fn main() -> pub [u8; 32] {
```
ashpect: Is there a reason to change the functions? 
prev functions provide better tests as i can see

rust-toolchain.toml
```diff
@@ -1,3 +1,4 @@
 [toolchain]
-channel = "nightly-2026-03-04"
-components = ["rustfmt", "rust-analyzer", "clippy"]
+channel = "nightly-2025-10-20"
```
ashpect: Please rebase the branch. This is defaulting to older nightly versions, which isn't compatible rn.

provekit/common/src/prover.rs
```diff
@@ -21,26 +21,15 @@ pub struct NoirProver {
...
-    pub ad_binary:          Vec<u64>,
-    pub constraints_layout: ConstraintsLayout,
-    pub witness_layout:     WitnessLayout,
-    pub hash_config:        HashConfig,
-}
-
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub enum Prover {
     Noir(NoirProver),
+    #[cfg(not(target_arch = "wasm32"))]
```
ashpect: Nitpick but when WASM tries to parse discriminant 1, we'd get deserialization error or silent data corruption. Maybe on WASM, MavrosProver can be a stub type that implements Serialize/Deserialize but errors on use

noir-examples/poseidon-rounds/src/main.nr
```diff
@@ -1,11 +1,8 @@
-use dep::poseidon2;
-
-fn main(plains: [Field; 2], result: Field) {
-    let mut hash = poseidon2::bn254::hash_2(plains);
+fn main(plains: [Field; 4], result: Field) -> pub [Field; 4] {
+    let mut hash = std::hash::poseidon2_permutation(plains, 4);
     let rounds = 1000;
     for _ in 0..rounds {
-        hash = poseidon2::bn254::hash_1([hash]);
+        hash = std::hash::poseidon2_permutation(hash, 4);
```
ashpect, line 5: just curious, why was Poseidon rounds impl changed ?

provekit/common/src/prover.rs
ashpect: 🔴 **Critical: `#[cfg]` on serde enum variant — fragile cross-target serialization**

Postcard encodes enum variants as 0-indexed varints by source declaration order. Today this works:
- **Native**: `Noir=0, Mavros=1`
- **WASM**: `Noir=0` (only variant)

But if anyone later adds a cfg-gated variant **before** `Noir`:
```rust
#[cfg(not(target_arch = "wasm32"))]
NewNativeOnly(Something),  // Shifts Noir to index 1 on native
Noir(NoirProver),          // Still index 0 on WASM
```
Native-serialized `Noir` (discriminant 1) silently fails to deserialize on WASM. No compiler warning. No type error. Silent runtime failure.

**Postcard doesn't support `#[serde(tag = "type")]`** (requires `deserialize_any` which postcard doesn't implement), and **`#[serde(rename)]` is irrelevant** (postcard uses integer indices, not string names). There is no serde annotation that fixes this.

**Fix**: Keep both variants unconditional. On WASM, `MavrosProver` can be a stub type that `Serialize`/`Deserialize`s with identical wire format but errors on use. If `mavros_vm` truly can't compile for WASM, provide stub types for `ConstraintsLayout`/`WitnessLayout`/`Abi` behind `#[cfg]` at the *field type* level, not at the *enum variant* level.

At minimum, add a compile-time discriminant test:
```rust
#[test]
fn prover_noir_discriminant_is_zero() {
    // SAFETY: variant order is wire-format-critical for cross-target compat.
    let bytes = postcard::to_allocvec(&Prover::Noir(/* minimal */)).unwrap();
    assert_eq!(bytes[0], 0x00);
}
```

And add this comment above the enum:
```rust
// INVARIANT: Variant order is wire-format-critical (postcard uses positional
// discriminants). Do not reorder, cfg-gate, or insert variants without
// verifying cross-target deserialization (native ↔ WASM).
```

provekit/common/src/noir_proof_scheme.rs
```diff
@@ -20,23 +20,10 @@ pub struct NoirSchemeData {
...
-    pub ad_binary:          Vec<u64>,
-    pub constraints_layout: ConstraintsLayout,
-    pub witness_layout:     WitnessLayout,
-    pub hash_config:        HashConfig,
-}
-
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub enum NoirProofScheme {
     Noir(NoirSchemeData),
+    #[cfg(not(target_arch = "wasm32"))]
```
ashpect: Same cfg-on-variant concern as `Prover` enum above. keep `Mavros` variant unconditional, stub types on WASM.

provekit/prover/src/lib.rs
```diff
@@ -7,32 +7,50 @@ use {
...
+/// Proof generation trait.
+///
+/// `prove` and `prove_with_toml` are only available on native targets with the
+/// `witness-generation` feature (cfg-gated out on `wasm32`).
+/// `prove_with_witness` is available on all targets including WASM.
+///
+/// `MavrosProver` does not support `prove_with_witness` and returns an error
+/// at runtime — only Noir provers support witness-based proving.
 pub trait Prove {
+    #[cfg(all(feature = "witness-generation", not(target_arch = "wasm32")))]
```
ashpect, line 38: 🟡 **cfg-gated trait methods — target-dependent API surface**

The `Prove` trait now has different methods depending on the target. Generic code written on native that calls `.prove()` will fail to compile on WASM with a confusing "method not found" error.

Consider splitting:
```rust
/// Always available (including WASM)
pub trait Prove {
    fn prove_with_witness(self, witness: WitnessMap<NoirElement>) -> Result<NoirProof>;
}

/// Native-only extension
#[cfg(all(feature = "witness-generation", not(target_arch = "wasm32")))]
pub trait ProveFromInput: Prove {
    fn prove(self, input_map: InputMap) -> Result<NoirProof>;
    fn prove_with_toml(self, prover_toml: impl AsRef<Path>) -> Result<NoirProof>;
}
```

This keeps the trait interface stable across all targets. Not blocking but strongly recommended as a follow-up.

Also: the double-gating `#[cfg(all(feature = "witness-generation", not(target_arch = "wasm32")))]` is redundant — the WASM Cargo.toml already uses `default-features = false` which excludes `witness-generation`. The feature flag alone should be sufficient.

playground/wasm-demo/scripts/setup.mjs
```diff
@@ -0,0 +1,556 @@
...
+function setNestedValue(obj, path, value) {
+  const parts = path.split(".");
+  let current = obj;
+  for (let i = 0; i < parts.length - 1; i++) {
+    if (!(parts[i] in current)) {
+      current[parts[i]] = {};
+    }
+    current = current[parts[i]];
+  }
+  current[parts[parts.length - 1]] = value;
```
ashpect, line 324: `--hash blake3` is hardcoded. The project's default/optimized hash is Skyscraper. Is blake3 chosen because Skyscraper doesn't compile for WASM, or for other reasons? add comment for reason

playground/wasm-demo/src/demo-web.mjs
```diff
@@ -0,0 +1,566 @@
...
+    if (ch === delimiter && braces === 0 && brackets === 0) {
+      parts.push(current);
+      current = '';
+      continue;
+    }
+    current += ch;
+  }
+  if (current.trim()) parts.push(current);
+  return parts;
+}
```
ashpect, line 120: This is the second hand-rolled TOML parser — `setup.mjs` has its own (`parseProverToml`, ~170 lines). Two independent implementations of the same format parsing, neither fully TOML-compliant.

Consider extracting a shared `toml-utils.mjs` used by both, or using `smol-toml` (4KB gzipped, fully compliant). Not blocking for a playground demo, but fragile.

playground/wasm-demo/scripts/setup.mjs
Bisht13: blake3's hash engine is the fastest, hence the default for wasm, can change for provekit in the future as well.

noir-examples/poseidon-rounds/src/main.nr
Bisht13: The original was noir native implementation, the change now uses blackbox.

provekit/prover/src/lib.rs
Bisht13: Tracking here https://github.com/worldfnd/provekit/issues/337

playground/wasm-demo/src/demo-web.mjs
Bisht13: Tracking here https://github.com/worldfnd/provekit/issues/339

PR #247, opened 2025-12-25
SHA-256: Initial constraint optimizations

rose2221:
## Summary
Optimizes SHA256 compression constraint generation through byte-level operations and fused constraints.

### Results :- 

| Metric | Before | After | Improvement |
|-------|--------|-------|-------------|
| Constraints | 432,976 | 152699  | -64.73% |
| Constraints (per SHA256 compression call) | 39,293 | 20,907| -46.79% |

SHA256 Compression – R1CS Cost Breakdown 
Component | Constraints | Witnesses
-- | -- | --
Direct  | 3,600 | 13,364
AND + XOR | 144,389 | 212,998
RANGE | 4,651 | 5,145
Total | 152,640 | 231,507

Marginal SHA256 Call
Component | Constraints / call | Witnesses / call
-- | -- | --
Direct | 3,600 | 13,364
AND + XOR | 13,312 | 16,384
RANGE | 3,995 | 4,169
Total per SHA256 | 20,907 | 33,917

### Key Optimizations:- 
 **1. New Type System for Byte-Level Operations**
- Introduced U8 and U32 wrapper types to replace raw usize witness indices
- U32 represents a 32-bit word as 4 little-endian U8 bytes
- Each U8 tracks whether it has been range-checked via a `range_checked` flag
- Enables byte-level operations that avoid expensive 32-bit digital decomposition

**2. ROTR with Fused Constraints**
- Implemented ROTR using free byte reordering, byte partitioning into lo and hi via a new BytePartition witness builder, and a single fused constraint per output byte: `result * 2^k + lo = byte + lo_next * 256`
This achieves 4 constraints per ROTR (one per byte) instead of multiple decomposition constraints.

**3. SHR Implementation with Fused Constraints**
- Similar optimization to ROTR but handles zero-fill for shifted-out bits.
- Byte-level shift is free (reordering + zero padding)
- Uses same fused constraint approach for intra-byte bit shifting
- MSB byte has special handling: `result * 2^k + lo = byte` (no next byte term)

**4. Range Check Optimization**
- Only range-checks lo during byte partitioning. Soundness is preserved because the partitioned high bits (hi) contribute to a derived output byte via fused constraints, and that output byte is subsequently constrained to [0,255] by downstream byte-level lookup tables (e.g. AND/XOR). This reduces the number of explicit range checks in this context without sacrificing soundness.

**5.. New BytePartition WitnessBuilder**
- Added dedicated witness builder variant for splitting an 8-bit value at a bit boundary.
- Computes lo (lower k bits) and hi (upper 8-k bits) such that: `x = lo + hi * 2^k`
- Produces 2 witness values per invocation and is reused across ROTR and SHR operations.

**6.. Chained U32 Addition with Fused Constraints**
- Replaced chained pairwise additions in T1 computation (h + Σ₁(e) + Ch(e,f,g) + K[i] + W[i]) with a single variadic addition using the fused constraint: `a + b + c + d + e = result + carry * 2^32`
- Reduces to 1 constraint + 2 witnesses, with carry range-checks sized precisely based on input count.

**7. Byte-Level Binary Operations (AND/XOR)**
- Kept SHA256 AND/XOR operations at the 8-bit level by introducing `and_ops_byte` and `xor_ops_byte` tracking and an `is_byte_level` flag in`add_binop_constraints`, avoiding unnecessary 32-bit decomposition while preserving the full decomposition path for general Noir blackbox ops.

**8. Range Check & LogUp Optimizations**
- Introduced fused constraints for LogUp inverse computation, replacing the two-step denominator + inverse verification.
  * **Before:** Compute `(X - c·v)` and verify `denominator * inverse = 1`, 2 constraints + 2 witnesses.
  * **After:** Single fused constraint: `(X - c·v) * inverse = 1`, 1 constraint + 1 witness.
- Added a new `WitnessBuilder::LogUpInverse` variant that computes the inverse directly during batch inversion, eliminating the intermediate denominator witness.
- Fused the log-derivative multiset equality check into a single constraint.
  * **Before:** Separate constraints for table sum, witness sum, and equality check (**3 constraints**).
  * **After:** Single constraint: `(Σ table_terms - Σ witness_terms) = 0`. Removes 2 constraints and 2 intermediate sum witnesses per range-check lookup.
- Extended batch inversion support to include `LogUpInverse`.
  * `LogUpInverse` is deferred and batched alongside existing inverse operations.
  * Denominators `(X - c·v)` are computed inline during batch inversion.

**8. BinOp LogUp Constraint Fusion (`binops.rs`)**

- Replaced separate binop-side and table-side summations with a single fused sum using negated coefficients.
- Enforces equality via one constraint: `(Σ binop_terms − Σ table_terms) = 0`.
- Removes intermediate sum witnesses and reduces constraints per AND/XOR lookup.

**9. Combined AND/XOR Lookup Table**

- Replaced separate lookup tables for AND and XOR operations with a single combined table.
- Before: Two independent lookup tables (~196K constraints each).
- After: One combined table using a 4-term encoding: `sz − (lhs + rs·rhs + rs²·and + rs³·xor)`. Eliminates one full lookup table.
- Savings: ~147K constraints.

**10. Byte-Level Binary Operations (SHA256)**

- Avoided 32-bit decomposition for SHA256 AND/XOR by operating directly on U8 byte types.
- Before: 32-bit AND/XOR required decomposition into 4 bytes followed by 4 separate 8-bit lookups.
- After: SHA256 binops are executed directly at the byte level using U8, skipping decomposition entirely.
- Eliminates decomposition constraints and intermediate witnesses per binary operation.

**11. Inlined T1 / T2 Computation in SHA256 Rounds**

- Removed intermediate T1 and T2 values in SHA256 round computation.
- Before:
```
T1 = h + Σ₁(e) + Ch(e,f,g) + K[i] + W[i]
T2 = Σ₀(a) + Maj(a,b,c)
new_e = d + T1
new_a = T1 + T2
```
- After: new_e and new_a are computed directly with all terms inlined.
- Savings: 12 witnesses per round (result, carry, and unpacked bytes for T1/T2) × 64 rounds = 768 witnesses per SHA256 compression.

**12. Inline Table Entry Inverse (CombinedTableEntryInverse)**

- Optimized lookup inverse handling by inlining the denominator directly into the inverse constraint.
- Before: Create denominator witness, then verify denominator × inverse = 1 (2 witnesses per entry).
- After: Single fused constraint: `(sz − lhs − rs·rhs − rs²·and − rs³·xor) × inv = 1`. Eliminates the intermediate denominator witness.
- Savings: 65,536 witnesses (one per table entry).

**13. Pack Cache for U32 Operations**

- Introduced memoization for U32 packing operations.
- Before: Identical U32 byte sequences could be repacked multiple times within a compression.
- After: `pack_cached()` caches pack results keyed by byte indices and reuses them.
- Avoids duplicate pack witnesses and constraints across SHA256 rounds.

### DIFF
--- a/provekit/r1cs-compiler/src/noir_to_r1cs.rs
+++ b/provekit/r1cs-compiler/src/noir_to_r1cs.rs
@@ -460,18 +460,17 @@ impl NoirToR1CSCompiler {
                                 match rhs.input() {
                                     ConstantOrACIRWitness::Constant(_) => {
                                         // Both lhs and rhs are constants
-                                        let rhs_c = self.fetch_constant_or_r1cs_witness(rhs.input());
+                                        let rhs_c =
+                                            self.fetch_constant_or_r1cs_witness(rhs.input());
                                         let ConstantOrR1CSWitness::Constant(rhs_fe) = rhs_c else {
                                             unreachable!()
                                         };
                                         let rhs_val: u64 = rhs_fe.into_bigint().0[0];
                                         let rhs_bytes: [u8; 4] = (rhs_val as u32).to_le_bytes();
 
-                                        let dd = add_digital_decomposition(
-                                            self,
-                                            log_bases,
-                                            vec![out_idx],
-                                        );
+                                        let dd = add_digital_decomposition(self, log_bases, vec![
+                                            out_idx,
+                                        ]);
                                         for byte_idx in 0..NUM_DIGITS {
                                             let lhs_byte = ConstantOrR1CSWitness::Constant(
                                                 FieldElement::from(lhs_bytes[byte_idx] as u64),
@@ -486,11 +485,10 @@ impl NoirToR1CSCompiler {
                                     ConstantOrACIRWitness::Witness(rhs_w) => {
                                         // lhs constant, rhs witness - decompose rhs witness
                                         let rhs_witness = self.fetch_r1cs_witness_index(rhs_w);
-                                        let dd = add_digital_decomposition(
-                                            self,
-                                            log_bases,
-                                            vec![rhs_witness, out_idx],
-                                        );
+                                        let dd = add_digital_decomposition(self, log_bases, vec![
+                                            rhs_witness,
+                                            out_idx,
+                                        ]);
                                         for byte_idx in 0..NUM_DIGITS {
                                             let lhs_byte = ConstantOrR1CSWitness::Constant(
                                                 FieldElement::from(lhs_bytes[byte_idx] as u64),
@@ -580,18 +578,17 @@ impl NoirToR1CSCompiler {
                                 match rhs.input() {
                                     ConstantOrACIRWitness::Constant(_) => {
                                         // Both lhs and rhs are constants
-                                        let rhs_c = self.fetch_constant_or_r1cs_witness(rhs.input());
+                                        let rhs_c =
+                                            self.fetch_constant_or_r1cs_witness(rhs.input());
                                         let ConstantOrR1CSWitness::Constant(rhs_fe) = rhs_c else {
                                             unreachable!()
                                         };
                                         let rhs_val: u64 = rhs_fe.into_bigint().0[0];
                                         let rhs_bytes: [u8; 4] = (rhs_val as u32).to_le_bytes();
 
-                                        let dd = add_digital_decomposition(
-                                            self,
-                                            log_bases,
-                                            vec![out_idx],
-                                        );
+                                        let dd = add_digital_decomposition(self, log_bases, vec![
+                                            out_idx,
+                                        ]);
                                         for byte_idx in 0..NUM_DIGITS {
                                             let lhs_byte = ConstantOrR1CSWitness::Constant(
                                                 FieldElement::from(lhs_bytes[byte_idx] as u64),
@@ -606,11 +603,10 @@ impl NoirToR1CSCompiler {
                                     ConstantOrACIRWitness::Witness(rhs_w) => {
                                         // lhs constant, rhs witness - decompose rhs witness
                                         let rhs_witness = self.fetch_r1cs_witness_index(rhs_w);
-                                        let dd = add_digital_decomposition(
-                                            self,
-                                            log_bases,
-                                            vec![rhs_witness, out_idx],
-                                        );
+                                        let dd = add_digital_decomposition(self, log_bases, vec![
+                                            rhs_witness,
+                                            out_idx,
+                                        ]);
                                         for byte_idx in 0..NUM_DIGITS {
                                             let lhs_byte = ConstantOrR1CSWitness::Constant(
                                                 FieldElement::from(lhs_bytes[byte_idx] as u64),


provekit/r1cs-compiler/src/binops.rs
```diff
@@ -20,245 +22,381 @@ pub enum BinOp {
...
+            let key = (operand_key(lhs), operand_key(rhs));
+            pair_map
+                .entry(key)
+                .and_modify(|e| e.1 = Some(*xor_out))
+                .or_insert((None, Some(*xor_out), *lhs, *rhs));
+        }
+
+        // Now build combined ops, creating complementary witnesses only when needed
+        let mut combined = Vec::with_capacity(pair_map.len());
+        for (_key, (and_opt, xor_opt, lhs, rhs)) in pair_map {
```
Bisht13: HashMap iteration order is non-deterministic (randomized per run). This means constraint ordering will vary between compilations of the same circuit, breaking reproducibility.

Should use BTreeMap instead, or sort entries before iteration:
```
let mut pair_map: BTreeMap<(u64, u64), PairMapEntry> = BTreeMap::new();
```

provekit/r1cs-compiler/src/binops.rs
```diff
@@ -20,245 +22,381 @@ pub enum BinOp {
...
+    // For combined table, each operation needs both AND and XOR outputs.
+    // Convert ops to atomic (byte-level) operations with both outputs.
+    let combined_ops_atomic = if is_byte_level {
+        // Already byte-level. Create combined ops with complementary outputs.
+        // Optimization: If the same (lhs, rhs) pair appears in both AND and XOR ops,
+        // we already have both outputs and don't need to create complementary
+        // witnesses.
+
+        // Helper to create a key for (lhs, rhs) pair
+        fn operand_key(op: &ConstantOrR1CSWitness) -> u64 {
```
Bisht13: This can collide when two different field elements share the same lower 64 bits (the BigInt::0[0] is just the first limb). If field modulus > 2^64, multiple field elements map to the same key, causing incorrect constraint generation.

provekit/r1cs-compiler/src/sha256_compression.rs
```diff
@@ -24,153 +27,420 @@ const SHA256_K: [u32; 64] = [
...
-        &[(FieldElement::ONE, result_witness)],
     );
 
-    // Range check the result to ensure it's a valid 32-bit value
-    range_checks.entry(32).or_default().push(result_witness);
-
-    // Range check intermediate values
+    // Range check carry: max carry is (N-1) for N inputs
+    let n = inputs.len();
+    let max_carry = n - 1;
```
Bisht13, line 353: When inputs.len() == 1, max_carry = 0 but you allocate 1 bit. With a single input there's no overflow possible, so carry is always 0. The 1-bit range check is overly permissive (allows carry=1 which is invalid).

Either:
1. Early return for single input (line 309 already handles this)
2. Or add a comment explaining why 1-bit is acceptable here

Same issue exists in `add_u32_multi_addition_with_const`

provekit/common/src/witness/scheduling/dependency.rs
```diff
@@ -162,6 +191,24 @@ impl DependencyInfo {
                 }
                 v
             }
+            WitnessBuilder::XorTriple(_, a, b, c) => {
```
Bisht13: `XorTriple` and `ByteBitDecomposition` variants are defined with full solver implementations but never constructed anywhere in this PR. Can remove safely.

provekit/r1cs-compiler/src/sha256_compression.rs
```diff
@@ -24,153 +27,420 @@ const SHA256_K: [u32; 64] = [
...
+        r1cs_compiler.r1cs.add_constraint(
+            &[(two_pow_k, res_idx), (FieldElement::ONE, parts[i].0.idx)],
+            &[(FieldElement::ONE, r1cs_compiler.witness_one())],
+            &[
+                (FieldElement::ONE, rot[i].idx),
+                (two_pow_8, parts[next].0.idx),
+            ],
+        );
+
+        out_bytes[i] = U8::new(res_idx, true);
```
Bisht13, line 126: Result is marked range_checked: true but no explicit 8-bit range check is added. The soundness proof in `partition_byte_witnesses` explains why this is correct, but a brief comment here would help readers.

Bisht13:
Nit: Remove unused import.

provekit/r1cs-compiler/src/binops.rs
rose2221: Switched to BTreeMap.

provekit/r1cs-compiler/src/binops.rs
rose2221: Fixed. The operand_key function now uses all 4 limbs of the field element's BigInt representation instead of just the first limb

provekit/r1cs-compiler/src/sha256_compression.rs
rose2221: Done. Updated both functions to skip the range check when max_carry == 0

provekit/common/src/witness/scheduling/dependency.rs
rose2221: Removed.

provekit/r1cs-compiler/src/sha256_compression.rs
rose2221: fixed.

Bisht13 (approved):

PR #256, opened 2026-01-20
fix e2e native verification

ashpect:
- Adds r1cs to verifier.pkv
- Verifies deferred linear (r1cs/matrix evaals) and geometric constraints(pub inputs constraint evals)

### DIFF
--- a/provekit/verifier/src/whir_r1cs.rs
+++ b/provekit/verifier/src/whir_r1cs.rs
@@ -200,6 +200,11 @@ impl WhirR1CSVerifier for WhirR1CSScheme {
 
         // Linear deferred
         if self.num_challenges > 0 {
+            assert!(
+                deferred_evals.len() == offset + 6,
+                "Deferred evals length does not match"
+            );
+
             let matrix_extension_evals = evaluate_r1cs_matrix_extension_batch(
                 r1cs,
                 &data_from_sumcheck_verifier.alpha,
@@ -214,6 +219,11 @@ impl WhirR1CSVerifier for WhirR1CSScheme {
                 );
             }
         } else {
+            assert!(
+                deferred_evals.len() == offset + 3,
+                "Deferred evals length does not match"
+            );
+
             let matrix_extension_evals = evaluate_r1cs_matrix_extension(
                 r1cs,
                 &data_from_sumcheck_verifier.alpha,
@@ -230,11 +240,10 @@ impl WhirR1CSVerifier for WhirR1CSScheme {
         }
 
         // Geometric deferred
-        if !public_inputs.is_empty() {
+        if !public_inputs.is_empty() && deferred_evals.len() > 0 {
             let public_weight_eval = compute_public_weight_evaluation(
                 public_inputs,
                 &whir_folding_randomness,
-                self.whir_witness.mv_parameters.num_variables,
                 public_weights_challenge,
             );
             ensure!(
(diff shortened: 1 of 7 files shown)


provekit/verifier/src/whir_r1cs.rs
```diff
@@ -164,6 +195,54 @@ impl WhirR1CSVerifier for WhirR1CSScheme {
...
+
+        // Linear deferred
+        if self.num_challenges > 0 {
+            let matrix_extension_evals = evaluate_r1cs_matrix_extension_batch(
+                r1cs,
+                &data_from_sumcheck_verifier.alpha,
+                &whir_folding_randomness,
+                self.w1_size,
+            );
+            for i in 0..6 {
```
Bisht13, line 214: Missing bounds check on `deferred_evals`. This assumes `deferred_evals.len() >= offset + 6`. If the prover provides a malformed proof with fewer deferred values, this will panic instead of returning an error.

provekit/verifier/src/whir_r1cs.rs
```diff
@@ -164,6 +195,54 @@ impl WhirR1CSVerifier for WhirR1CSScheme {
...
+                );
+            }
+        } else {
+            let matrix_extension_evals = evaluate_r1cs_matrix_extension(
+                r1cs,
+                &data_from_sumcheck_verifier.alpha,
+                &whir_folding_randomness,
+            );
+
+            for i in 0..3 {
```
Bisht13, line 233: Same issue https://github.com/worldfnd/provekit/pull/256/changes#r2724048058.

provekit/verifier/src/whir_r1cs.rs
```diff
@@ -164,6 +195,54 @@ impl WhirR1CSVerifier for WhirR1CSScheme {
...
+                ensure!(
+                    matrix_extension_evals[i] == deferred_evals[offset + i],
+                    "Matrix extension evaluation {} does not match deferred value",
+                    i
+                );
+            }
+        }
+
+        // Geometric deferred
+        if !public_inputs.is_empty() {
```
Bisht13: Missing bounds check for `deferred_evals[0]`. When `!public_inputs.is_empty()`, we access `deferred_evals[0]` but don't verify the vector is non-empty.

provekit/verifier/src/whir_r1cs.rs
```diff
@@ -292,3 +371,101 @@ pub fn run_whir_pcs_batch_verifier(
...
+}
+
+fn compute_public_weight_evaluation(
+    public_inputs: &PublicInputs,
+    folding_randomness: &[FieldElement],
+    m: usize,
+    x: FieldElement,
+) -> FieldElement {
+    let domain_size = 1 << m;
+    let mut public_weights = vec![FieldElement::zero(); domain_size];
```
Bisht13: The `public_weights` vector is allocated and populated but **only used inside `#[cfg(test)]`**. In release builds, this is a wasted allocation of `O(2^m)` field elements.

provekit/verifier/src/whir_r1cs.rs
```diff
@@ -292,3 +371,101 @@ pub fn run_whir_pcs_batch_verifier(
...
+        }
+    }
+
+    ans
+}
+
+fn compute_public_weight_evaluation(
+    public_inputs: &PublicInputs,
+    folding_randomness: &[FieldElement],
+    m: usize,
```
Bisht13: The `m` parameter is only used for `domain_size` calculation which is inside `#[cfg(test)]`. This will cause a compiler warning.

tooling/cli/src/cmd/verify.rs
```diff
@@ -31,9 +35,12 @@ impl Command for Args {
         // Read the proof
         let proof = read(&self.proof_path).context("while reading proof")?;
 
+        // Read the prover and R1CS (similar to generate-gnark-inputs)
+        let prover: Prover = read(&self.prover_path).context("while reading Provekit Prover")?;
+
         // Verify the proof
         verifier
-            .verify(&proof)
+            .verify(&proof, &prover.r1cs)
```
Bisht13: It's common for the verifier to have access to the constraint system (it's public information). Consider embedding the R1CS data in the pkv.

provekit/verifier/src/whir_r1cs.rs
ashpect: Added an assert check

provekit/verifier/src/whir_r1cs.rs
ashpect: Added an assert check

provekit/verifier/src/whir_r1cs.rs
ashpect: Added check

provekit/verifier/src/whir_r1cs.rs
ashpect: Resolved, geometric_till test can be added separately to resolve this

provekit/verifier/src/whir_r1cs.rs
ashpect: Resolved , ref : https://github.com/worldfnd/provekit/pull/256#discussion_r2725392439

tooling/cli/src/cmd/verify.rs
ashpect: Added r1cs to verifier.pkv

PR #259, opened 2026-01-23
batched bn254 51 bits RNE multiplier 

xrvdg:
- Add 51-bit floating-point Montgomery multiplication for BN254 scalar field optimised for WASM
- Rename block-multiplier to bn254-multiplier with reorganized module structure
- Optimize carries and squaring operations in the b51 multiplier
- Add Kani verification for integer-to-float conversion
- Enable relaxed SIMD in WASI runners
- Make benchmarks WASI-compatible

**Reviewing**: Not all file movements and renames are tracked properly by Github's diff view. The important folder is `rne` which contain the new Montgomery multiplication implementation. I've tried to stay true to the structure of the one in `rtz`.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -21,6 +21,7 @@ jobs:
       - name: Replace divan with codspeed-divan-compat
         run: |
           sed -i 's/^divan = .*/divan = { package = "codspeed-divan-compat", version = "3.0.1" }/' Cargo.toml
+          grep -q 'codspeed-divan-compat' Cargo.toml || { echo "ERROR: divan replacement failed"; exit 1; }
 
       - name: Setup Rust toolchain, cache and cargo-codspeed binary
         uses: moonrepo/setup-rust@v1
--- a/skyscraper/bn254-multiplier/src/rne/portable_simd.rs
+++ b/skyscraper/bn254-multiplier/src/rne/portable_simd.rs
@@ -16,6 +16,7 @@ use {
         ops::BitAnd,
         simd::{num::SimdFloat, Simd},
     },
+    seq_macro::seq,
     std::simd::num::{SimdInt, SimdUint},
 };
 
@@ -27,7 +28,8 @@ pub fn simd_sqr(v0_a: [u64; 4], v1_a: [u64; 4]) -> ([u64; 4], [u64; 4]) {
 
     let mut t: [Simd<i64, 2>; 10] = [Simd::splat(0); 10];
 
-    for i in 0..5 {
+    // Non-diagonals first such that they can be doubled
+    seq!( i in 0..5 {
         let avi: Simd<f64, 2> = i2f(v0_a[i]);
         for j in (i + 1)..5 {
             let bvj: Simd<f64, 2> = i2f(v0_a[j]);
@@ -36,21 +38,22 @@ pub fn simd_sqr(v0_a: [u64; 4], v1_a: [u64; 4]) -> ([u64; 4], [u64; 4]) {
             t[i + j + 1] += p_hi.to_bits().cast();
             t[i + j] += p_lo.to_bits().cast();
         }
-    }
+    });
 
     // Most shifting operations are more expensive addition thus for multiplying by
     // 2 we use addition.
-    for i in 1..=8 {
+    seq!( i in 1..=8 {
         t[i] += t[i];
-    }
+    });
 
-    for i in 0..5 {
+    // Diagonals
+    seq!( i in 0..5 {
         let avi: Simd<f64, 2> = i2f(v0_a[i]);
         let p_hi = fma(avi, avi, Simd::splat(C1));
         let p_lo = fma(avi, avi, Simd::splat(C2) - p_hi);
         t[i + i + 1] += p_hi.to_bits().cast();
         t[i + i] += p_lo.to_bits().cast();
-    }
+    });
 
     t[0] += Simd::splat(make_initial(1, 0));
     t[9] += Simd::splat(make_initial(0, 6));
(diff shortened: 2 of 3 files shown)


skyscraper/bn254-multiplier/src/rtz/constants.rs
```diff
@@ -0,0 +1,71 @@
...
+    0x060c89ce5c263,
+];
+
+pub const F52_P: [f64; 5] = [
+    0x1f593f0000001_u64 as f64,
+    0x4879b9709143e_u64 as f64,
+    0x181585d2833e8_u64 as f64,
+    0xa029b85045b68_u64 as f64,
+    0x030644e72e131_u64 as f64,
+];
```
ocdbytes: This constant is not used anywhere. If for future use adding a comment can be helpful. Also on side note can you add a comment on each of these constants. Because for me (inexperienced) It looked like magic numbers and had to search through to find what they mean.

skyscraper/bn254-multiplier/src/rne/portable_simd.rs
```diff
@@ -0,0 +1,379 @@
...
+    },
+    core::{
+        ops::BitAnd,
+        simd::{num::SimdFloat, Simd},
+    },
+    std::simd::num::{SimdInt, SimdUint},
+};
+
+/// Two parallel Montgomery squarings: `(v0², v1²)`.
+/// input must fit in 2^255-1; no runtime checking
```
ocdbytes, line 24: Doubt : why are we not checking the limits of the inputs here ?

skyscraper/bn254-multiplier/src/rtz/portable_simd.rs
```diff
@@ -377,3 +383,36 @@ pub fn simd_mul(
     let v = transpose_simd_to_u256(u256_result);
     (v[0], v[1])
 }
+
+#[cfg(test)]
+mod tests {
```
ocdbytes, line 388: Nit Pick : Can we add a test for test_sim_sqr ?

skyscraper/bn254-multiplier/src/rne/portable_simd.rs
```diff
@@ -0,0 +1,379 @@
...
+#[inline]
```
ocdbytes: Nit Pick
```suggestion
#[inline(always)]
```

skyscraper/bn254-multiplier/src/rne/portable_simd.rs
xrvdg: The main reason is that these are internal functions to a kernel that runs a hot loop. The kernel needs to be designed in a way that the precondition holds because checking in a hot loop is costly. This note is there as a reminder to let the kernel have this invariant. 

I did add a debug assert to u256_to_u255 to check if the invariant is actually being uphold when running in debug mode.

ocdbytes: cool works. Thanks for clarifying

skyscraper/bn254-multiplier/src/rne/simd_utils.rs
```diff
@@ -0,0 +1,247 @@
...
+
+/// Convert 4×64-bit to 5×51-bit limb representation.
+/// Input must fit in 255 bits; no runtime checking.
+#[inline(always)]
+pub fn u256_to_u255_simd<const N: usize>(limbs: [Simd<u64, N>; 4]) -> [Simd<u64, N>; 5]
+where
+    LaneCount<N>: SupportedLaneCount,
+{
+    debug_assert!(limbs[3][0] & 1 << 63 == 0);
+    debug_assert!(limbs[3][1] & 1 << 63 == 0);
```
Bisht13: This function is generic over const N: usize (lane count), but the debug assertions hardcode lane indices [0] and [1]. When instantiated with N=1, which happens in the Kani proof `u256_to_u255_kani_roundtrip`, limbs[3][1] will panic with an index-out-of-bounds since Simd<u64, 1> only has lane 0.

I verified this: `Simd<u64,1>[1]` panics at runtime. Kani enables debug_assertions by default, so the `u256_to_u255_kani_roundtrip` proof should fail.

Suggestion. iterate over lanes:
```rs
for lane in 0..N {
     debug_assert!(limbs[3][lane] & (1 << 63) == 0);
}
```
(Also adding parens around (1 << 63) for readability, though precedence is correct without them.)

.github/workflows/benchmark.yml
```diff
@@ -18,6 +18,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - name: Replace divan with codspeed-divan-compat
```
Bisht13, line 21: The sed replacement is a clever workaround for the workspace dependency / target cfg limitation. However, it will silently fail if the divan line format ever changes (e.g., gets a trailing comment, moves to multi-line TOML, or is indented). Consider adding a verification step:
```yml
- name: Replace divan with codspeed-divan-compat
   run: |
     sed -i 's/^divan = .*/divan = { package = "codspeed-divan-compat", version = "3.0.1" }/' Cargo.toml
     grep -q 'codspeed-divan-compat' Cargo.toml || { echo "ERROR: divan replacement failed"; exit 1; }
```

skyscraper/bn254-multiplier/src/rne/portable_simd.rs
```diff
@@ -0,0 +1,380 @@
...
+    }
+
+    res[N - 1] = (t[N - 1] + borrow).cast();
+    res
+}
+
+/// Two parallel Montgomery multiplications: `(v0_a*v0_b, v1_a*v1_b)`.
+/// input must fit in 2^255-1; no runtime checking
+#[inline(always)]
+pub fn simd_mul(
```
Bisht13, line 130: Nit: `simd_sqr` uses compact nested loops for the cross-term accumulation, while `simd_mul` is fully manually unrolled into ~150 lines of repeated FMA blocks. This makes it harder to audit for indexing errors (e.g., verifying all 25 a[i]*b[j] products land in the correct t[i+j] / t[i+j+1] slots).

Was the unrolling measured to produce better codegen on WASM vs. the loop form? If so, a brief comment explaining the rationale would help. If not, the loop form (as in simd_sqr) would be more maintainable and easier to review.

skyscraper/bn254-multiplier/src/rne/portable_simd.rs
```diff
@@ -0,0 +1,380 @@
...
+                let bc_ref = ark_ff_reference(b, c);
+                let ab = Fr::new(BigInt(ab));
+                let bc = Fr::new(BigInt(bc));
+                prop_assert_eq!(ab_ref, ab, "mismatch: l = {:X}, b = {:X}", ab_ref.into_bigint(), ab.into_bigint());
+                prop_assert_eq!(bc_ref, bc, "mismatch: l = {:X}, b = {:X}", bc_ref.into_bigint(), bc.into_bigint());
+        })
+    }
+
+    #[test]
+    fn test_simd_sqr() {
```
Bisht13, line 236: Nit: `test_simd_sqr` validates against `simd_mul(a, a, b, b)` but not independently against `ark_ff_reference`. Since `test_simd_mul` already validates mul against ark_ff, the chain sqr → mul → ark_ff provides transitive coverage. But the sqr-specific code path (cross-terms doubled + diagonal terms added separately) differs from mul's schoolbook, so a direct `ark_ff_reference(a, a)` check would add confidence for free:
```rs
let a2_ref = ark_ff_reference(a, a);
prop_assert_eq!(Fr::new(BigInt(a2s)), a2_ref);
```

skyscraper/bn254-multiplier/src/rne/portable_simd.rs
xrvdg: Replaced manual unrolling with `seq!`

skyscraper/bn254-multiplier/src/rne/portable_simd.rs
xrvdg: `simd_sqr` can be bitwise compared against `simd_mul` without a transformation (happens within ark_ff_reference and Fr::new).  This makes sqr-specific code path easier to debug. Property-based tests with transitivity gives us adequate coverage.

Bisht13 (approved):

PR #269, opened 2026-02-02
mono bn254 51-bit SIMD FMA multiplier

xrvdg:
bn254 multiplier variant that uses SIMD round-to-nearest fused multiply add to compute a single multiplication. This variant is:
- the fastest single multiplier on ARM cortex, but not as high throughput as the batching variants.
- the fastest single multiplier on WASM and outperforms the batched variant #259 on WASM.

**Merge after #259**

### DIFF
--- a/skyscraper/bn254-multiplier/src/rne/mono.rs
+++ b/skyscraper/bn254-multiplier/src/rne/mono.rs
@@ -44,7 +44,7 @@ pub fn u256_to_u255(limbs: [u64; 4]) -> [u64; 5] {
         ((l0 >> 51) | (l1 << 13)) & MASK51,
         ((l1 >> 38) | (l2 << 26)) & MASK51,
         ((l2 >> 25) | (l3 << 39)) & MASK51,
-        l3 >> 12 & MASK51,
+        (l3 >> 12) & MASK51,
     ]
 }
 
@@ -115,7 +115,7 @@ pub fn reduce_ct(mut a: [i64; 5]) -> [i64; 5] {
 /// Montgomery multiplication for BN254 scalar field.
 ///
 /// Computes `a * b * R^{-256} mod P` where R = 2^256.
-
+///
 /// # Preconditions
 ///
 /// - Both inputs must be < 2^255 (i.e., fit in 5×51-bit limbs)


skyscraper/bn254-multiplier/src/rne/mono.rs
```diff
@@ -0,0 +1,340 @@
...
+    // 1. **Redundant SIMD form**: FMA produces high/low product parts in separate
+    //    SIMD vectors with misaligned lanes. E.g., `ts[k] = [limb_k_lo,
+    //    limb_{k+1}_lo]` and `ts[k+1] = [limb_k_hi, limb_{k+1}_hi]`. Swizzle
+    //    operations realign these so adjacent limbs occupy consecutive positions.
+    //
+    // 2. **Redundant limb form**: Each 51-bit limb may temporarily exceed 51 bits.
+    //    The excess bits are carries (or borrows) that propagate to higher limbs.
+    //    - `u64` limbs: excess bits are positive carries
+    //    - `i64` limbs: excess bits may be negative, representing borrows from
+    //      above    let a = u256_to_u255(a);
```
Bisht13: Nit: code leaked into the comment. This line should end at above `let a = u256_to_u255(a);` is a stray paste. (The actual code line below is fine.)

skyscraper/bn254-multiplier/src/rne/mono.rs
```diff
@@ -0,0 +1,340 @@
...
+    let [l0, l1, l2, l3, l4] = limbs;
+    [
+        (l0 >> 1) | (l1 << 50),
+        (l1 >> 14) | (l2 << 37),
+        (l2 >> 27) | (l3 << 24),
+        (l3 >> 40) | (l4 << 11),
+    ]
+}
+
+/// Montgomery squaring: `(va*v0_b, v1_a*v1_b)`.
```
Bisht13: Nit: Copy-paste from the batched variant. This mono sqr takes a single input and returns a single output. Something like  /// Montgomery squaring: `a² · R⁻¹ mod P`.  would be accurate.

Bisht13 (changes requested):

skyscraper/bn254-multiplier/src/rne/mono.rs
xrvdg: Fixed the comment. 

The result is congruent to `a² · R⁻¹ mod P` but is not the same. That's why I prefer `a²` and a mention of which Montgomery space it lives in (`R =256`).

PR #270, opened 2026-02-04
feat: single-constraint combined binop lookup table entries.

rose2221:
## Summary

- Replace the 2-constraint approach (`(γ - encoding) × inverse = 1` + `multiplicity × inverse = quotient`)
  with a single constraint `(γ - encoding) × quotient = multiplicity` for combined AND/XOR lookup table entries
- Saves ~65,536 R1CS constraints (one per table entry)

## Soundness note

**What changed:** Old approach used 2 constraints per table entry to force `(γ - encoding) != 0`.
New approach uses 1 constraint: `(γ - encoding) × quotient = multiplicity`.

**Why the 1-constraint version is safe:**

For used entries (multiplicity > 0), the constraint forces the correct quotient, same as before.

For unused entries (multiplicity = 0), the constraint becomes `0 × quotient = 0`, which is
always true. A cheater could set any quotient here. But this is only harmful if γ equals
that entry's encoding (making the denominator zero and breaking the lookup sum).

**Why the prover can't exploit this:**

The table encodings (lhs, rhs, and_out, xor_out for all 256×256 byte pairs) are hardcoded
constants baked into the R1CS at compile time, not prover-chosen witnesses.
The prover commits all w1 witnesses (including multiplicities) before γ is derived via
Fiat-Shamir (lib.rs: solve w1 → commit w1 → squeeze challenges during w2). So neither
the table entries nor γ are under the prover's control.

**Collision probability:**

  Field size (M31):           p  = 2^31 
  Chance γ hits one entry:    1/p
  Number of table entries:    256 × 256 = 65,536 = 2^16
  Chance γ hits any entry:    2^16 / 2^31 = 2^(-15) ≈ 1 in 32,768

### DIFF
--- a/provekit/common/src/witness/mod.rs
+++ b/provekit/common/src/witness/mod.rs
@@ -22,7 +22,7 @@ pub use {
     ram::{SpiceMemoryOperation, SpiceWitnesses},
     scheduling::{Layer, LayerType, LayeredWitnessBuilders, SplitError, SplitWitnessBuilders},
     witness_builder::{
-        CombinedTableEntryQuotientData, ConstantTerm, ProductLinearTerm, SumTerm, WitnessBuilder,
+        CombinedTableEntryInverseData, ConstantTerm, ProductLinearTerm, SumTerm, WitnessBuilder,
         WitnessCoefficient,
     },
     witness_generator::NoirWitnessGenerator,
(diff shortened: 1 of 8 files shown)


Bisht13:
LGTM!

Could we create an issue to track the M31 consideration? When/if we migrate to M31 in the future, this optimization would need to be revisited since the collision probability becomes 2^(-15) ≈ 1/32,768, which may not be acceptable.

PR #271, opened 2026-02-05
feat: dynamic base width optimization for range checks

rose2221:
## Summary
- Replace fixed 8-bit decomposition threshold with a dynamic base width optimizer that searches widths [2, 17] and picks the one minimizing total R1CS constraints
- Cost model accounts for both LogUp lookup and naive product-check strategies per bucket, choosing whichever is cheaper
- Fuse table entry inverse + product into a single R1CS constraint via add_range_table_entry_quotient (1 constraint per table entry instead of 2) similar to PR #270
**Collision probability to note:**
       1. Field size (M31): `p = 2³¹ - 1`
       2. Largest table (`MAX_BASE_WIDTH=17`): 2¹⁷ entries
       3. Pr[`X = t_j for some j`] = 2¹⁷ / (2³¹ - 1) ≈ 2⁻¹⁴ ≈ 1 in 16,384

- Saves ~63,740 constraints on complete_age_check (1250979 → 1,185,443 constraints)

### DIFF
--- a/provekit/r1cs-compiler/src/range_check.rs
+++ b/provekit/r1cs-compiler/src/range_check.rs
@@ -28,6 +28,21 @@ struct RangeCheckRequest {
     bits:        u32,
 }
 
+/// Determines whether LogUp is cheaper than naive range checking for a bucket
+/// of `num_bits`-wide checks containing `count` witnesses.
+///
+/// LogUp cost: 2^bits (table fused constraints) + count (witness inverse
+/// constraints) + 1 (grand sum check).
+/// Naive cost: (2^bits - 1) constraints per witness.
+///
+/// Returns `true` if LogUp should be used, `false` for naive product checks.
+fn should_use_logup(num_bits: u32, count: usize) -> bool {
+    let table_size = 1usize << num_bits;
+    let logup_cost = table_size.saturating_add(count).saturating_add(1);
+    let naive_cost = count.saturating_mul(table_size - 1);
+    logup_cost < naive_cost
+}
+
 /// Returns the constraint cost for a single atomic bucket of `num_bits`-wide
 /// checks containing `count` witnesses, choosing whichever strategy (LogUp
 /// or naive) is cheaper. Returns `usize::MAX` for impractically large bit
@@ -44,13 +59,13 @@ fn bucket_cost(num_bits: u32, count: usize) -> usize {
         return usize::MAX;
     }
     let table_size = 1usize << num_bits;
-    // LogUp cost: 2^bits (table fused constraints)
-    //           + count  (witness inverse constraints)
-    //           + 1      (grand sum check)
     let logup_cost = table_size.saturating_add(count).saturating_add(1);
-    // Naive cost: (2^bits - 1) constraints per witness
     let naive_cost = count.saturating_mul(table_size - 1);
-    logup_cost.min(naive_cost)
+    if should_use_logup(num_bits, count) {
+        logup_cost
+    } else {
+        naive_cost
+    }
 }
 
 /// Calculates the total R1CS constraint cost for a given `base_width`.
@@ -220,11 +235,7 @@ pub(crate) fn add_range_checks(
                 return;
             }
             let num_bits = num_bits as u32;
-            let count = values_to_lookup.len();
-            let table_size = 1usize << num_bits;
-            let logup_cost = table_size + count + 1;
-            let naive_cost = count.saturating_mul(table_size - 1);
-            if logup_cost < naive_cost {
+            if should_use_logup(num_bits, values_to_lookup.len()) {
                 add_range_check_via_lookup(r1cs, num_bits, &values_to_lookup);
             } else {
                 values_to_lookup.iter().for_each(|value| {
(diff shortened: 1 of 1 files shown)


provekit/r1cs-compiler/src/range_check.rs
```diff
@@ -243,3 +364,48 @@ fn add_naive_range_check(
...
     );
 }
+
+/// Computes quotient = multiplicity / (X - table_value) using a single R1CS
+/// constraint: (X - table_value) × quotient = multiplicity.
+///
+/// Internally creates an inverse witness (for batch inversion) and a product
+/// witness (inverse × multiplicity), but only emits one constraint instead
+/// of the usual two (inverse constraint + product constraint).
+fn add_range_table_entry_quotient(
```
Bisht13, line 385: #270 included an explicit soundness note for the identical fused-constraint optimization applied to binop tables (2^16 entries → 2^(−15) collision probability). This applies the same technique to range check tables but doesn't include a comparable analysis.

With `MAX_BASE_WIDTH = 17` the largest possible table has 2^17 entries, giving a worst-case collision probability of:
```
Pr[X = t_j for some j] = 2^17 / (2^31 − 1) ≈ 2^(−14) ≈ 1 in 16384
```
When this collision occurs the constraint (X − t_j) × quotient_j = 0 is trivially satisfied for any quotient_j, letting a malicious prover inject an arbitrary value into the LogUp sum.

Could you add a soundness note to the PR description and create an issue (like #270 did) documenting this? It's the same argument: X is Fiat-Shamir-derived after witness commitment, table entries are compile-time constants, but it should be stated explicitly, especially since the bound here (2^(−14)) is slightly weaker than #270's (2^(−15)).

provekit/r1cs-compiler/src/range_check.rs
```diff
@@ -106,14 +216,24 @@ pub(crate) fn add_range_checks(
...
             };
-            if values_to_lookup.len() > NUM_WITNESS_THRESHOLD_FOR_LOOKUP_TABLE {
-                add_range_check_via_lookup(r1cs, num_bits as u32, &values_to_lookup);
+            if values_to_lookup.is_empty() {
+                return;
+            }
+            let num_bits = num_bits as u32;
+            let count = values_to_lookup.len();
+            let table_size = 1usize << num_bits;
+            let logup_cost = table_size + count + 1;
```
Bisht13: Nit: This duplicates the decision logic from bucket_cost (lines 46–53), with a small inconsistency, bucket_cost uses saturating_add here while this uses plain +. Both are correct at current bounds (base_width ≤ 17 so table_size ≤ 131072), but a future change to the cost model formula could silently diverge from the actual dispatch.

Consider extracting a shared helper, e.g.:
```rs
fn should_use_logup(num_bits: u32, count: usize) -> bool {
     let table_size = 1usize << num_bits;
     let logup_cost = table_size.saturating_add(count).saturating_add(1);
     let naive_cost = count.saturating_mul(table_size - 1);
     logup_cost < naive_cost
 }
```
and calling it from both bucket_cost and here.

provekit/r1cs-compiler/src/range_check.rs
```diff
@@ -14,89 +14,199 @@ use {
...
+struct RangeCheckRequest {
+    witness_idx: usize,
+    bits:        u32,
+}
+
+/// Returns the constraint cost for a single atomic bucket of `num_bits`-wide
+/// checks containing `count` witnesses, choosing whichever strategy (LogUp
+/// or naive) is cheaper. Returns `usize::MAX` for impractically large bit
+/// widths where the table would overflow.
+fn bucket_cost(num_bits: u32, count: usize) -> usize {
```
Bisht13, line 50: `bucket_cost`, `calculate_constraint_cost`, and `get_optimal_base_width` are pure functions with clear contracts, we should have unit tests for them. The E2E tests validate the end result but wouldn't catch a regression specifically in the cost model (e.g. the optimizer picking a suboptimal width that still produces a correct but larger circuit).

Some edge cases worth covering:
```rs
assert_eq!(bucket_cost(0, 100), 0);
assert_eq!(bucket_cost(1, 1), 1);             // naive: 1×1=1, logup: 2+1+1=4 → naive wins
assert_eq!(bucket_cost(8, 5), 262);           // logup: 256+5+1=262, naive: 5×255=1275 → logup wins
assert_eq!(bucket_cost(31, 1), usize::MAX);   // overflow guard
```

rose2221: Added unit tests for the cost model. I focused on `should_use_logup` since that's where the actual decision logic lives.
I didn't add separate tests for `calculate_constraint_cost` and `get_optimal_base_width` because they don't really have independent logic worth testing, one just sums up bucket costs, the other just loops to find the min.

provekit/r1cs-compiler/src/range_check.rs
rose2221: Done, thanks for catching that.

provekit/r1cs-compiler/src/range_check.rs
rose2221: done, thanks #275

PR #273, opened 2026-02-05
feat: reduce artifact size (pkp/pkv)

ashpect:
Forr `provekit/noir-examples/noir-passport-examples/complete_age_check`, the prover.pkp file created after the prepare step is reduced from `28.6 MB -> 6.41 MB`.
Optimizations include : 
- Switching compression algo from zstd to xz2. xz2 excels our use case because most of our elements are repetitive. This saves the most size. 
- Use delta_compression while storing matrices. Delta compression saves around 1-2mb as well, by storing the diffs between elements in the r1cs matrices and reconstructing it. This works because bigger field elements takes more bytes to store.
- `analyze-pkp` cli command for analyzing artifacts. Analyzes the pkp file and gives metrics

### DIFF
--- a/provekit/common/src/file/bin.rs
+++ b/provekit/common/src/file/bin.rs
@@ -16,7 +16,7 @@ use {
 const HEADER_SIZE: usize = 20;
 const MAGIC_BYTES: &[u8] = b"\xDC\xDFOZkp\x01\x00";
 
-/// Write a compressed binary file (uses XZ level 9, slower than zstd2 but much
+/// Write a compressed binary file (uses XZ level 9, slower than zstd but much
 /// smaller).
 #[instrument(skip(value))]
 pub fn write_bin<T: Serialize>(
@@ -66,6 +66,7 @@ pub fn write_bin<T: Serialize>(
     Ok(())
 }
 
+/// Read a compressed binary file.
 #[instrument(fields(size = path.metadata().map(|m| m.len()).ok()))]
 pub fn read_bin<T: for<'a> Deserialize<'a>>(
     path: &Path,
--- a/recursive-verifier/app/circuit/circuit.go
+++ b/recursive-verifier/app/circuit/circuit.go
@@ -317,6 +317,9 @@ func verifyCircuit(
 	}
 
 	colIndicesA := internedR1CS.A.DecodeColIndices()
+	if colIndicesA == nil {
+		return fmt.Errorf("failed to decode column indices for matrix A: inconsistent data")
+	}
 	matrixA := make([]MatrixCell, len(internedR1CS.A.Values))
 	for i := range len(internedR1CS.A.RowIndices) {
 		end := len(internedR1CS.A.Values) - 1
@@ -333,6 +336,9 @@ func verifyCircuit(
 	}
 
 	colIndicesB := internedR1CS.B.DecodeColIndices()
+	if colIndicesB == nil {
+		return fmt.Errorf("failed to decode column indices for matrix B: inconsistent data")
+	}
 	matrixB := make([]MatrixCell, len(internedR1CS.B.Values))
 	for i := range len(internedR1CS.B.RowIndices) {
 		end := len(internedR1CS.B.Values) - 1
@@ -349,6 +355,9 @@ func verifyCircuit(
 	}
 
 	colIndicesC := internedR1CS.C.DecodeColIndices()
+	if colIndicesC == nil {
+		return fmt.Errorf("failed to decode column indices for matrix C: inconsistent data")
+	}
 	matrixC := make([]MatrixCell, len(internedR1CS.C.Values))
 	for i := range len(internedR1CS.C.RowIndices) {
 		end := len(internedR1CS.C.Values) - 1
--- a/recursive-verifier/app/circuit/circuit_test.go
+++ b/recursive-verifier/app/circuit/circuit_test.go
@@ -415,6 +415,9 @@ func buildCircuitAndAssignment(config Config, r1csData R1CS) (*Circuit, *Circuit
 
 func buildMatrix(sparse SparseMatrix, interner Interner) []MatrixCell {
 	colIndices := sparse.DecodeColIndices()
+	if colIndices == nil {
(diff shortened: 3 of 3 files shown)


recursive-verifier/app/circuit/matrix_evaluation.go
```diff
@@ -10,10 +10,48 @@ type SparseMatrix struct {
...
+			end = int(s.RowIndices[row+1])
+		}
+
+		rowLen := end - start
+		if rowLen == 0 {
+			continue
+		}
+
+		// First column is absolute
+		firstCol := s.ColDeltas[deltaIdx]
```
Bisht13, line 49: `s.ColDeltas[deltaIdx]` will panic if the JSON is inconsistent (e.g., `RowIndices` claims more entries than `ColDeltas` has). Since the server accepts R1CS files via HTTP upload, this could crash with malformed input. Consider a bounds check or a validation step after JSON unmarshal.

provekit/common/src/sparse_matrix.rs
```diff
@@ -20,13 +60,231 @@ pub struct SparseMatrix {
...
+    let mut deltas = Vec::with_capacity(col_indices.len());
+    let num_rows = new_row_indices.len();
+
+    for row in 0..num_rows {
+        let start = new_row_indices[row] as usize;
+        let end = new_row_indices
+            .get(row + 1)
+            .map_or(total_entries, |&v| v as usize);
+
+        let row_cols = &col_indices[start..end];
```
Bisht13, line 85: Nit: a `debug_assert!(row_cols.windows(2).all(|w| w[0] <= w[1]))` before the loop would document the sorted-columns invariant and catch violations during development.

provekit/common/src/file/bin.rs
```diff
@@ -10,54 +10,48 @@ use {
...
     },
     tracing::{info, instrument},
-    zstd::stream::{Decoder as ZstdDecoder, Encoder as ZstdEncoder},
+    xz2::{read::XzDecoder, write::XzEncoder},
 };
 
-const ZSTD_COMPRESSION: i32 = zstd::DEFAULT_COMPRESSION_LEVEL;
 const HEADER_SIZE: usize = 20;
-const MAGIC_BYTES: &[u8] = b"\xDC\xDFOZkp\x01\x00";
+const MAGIC_BYTES: &[u8] = b"\xDC\xDFOZkp\x02\x00";
```
Bisht13: I think we should keep the `MAGIC_BYTES` same but bump the `VERSION`.

provekit/common/src/file/bin.rs
```diff
@@ -10,54 +10,48 @@ use {
...
 
-/// Write a compressed binary file (fast and small).
```
Bisht13, line 20: The doc comments for write_bin and read_bin were dropped, please restore them (updating "fast and small" to reflect XZ level 9, which is slow but small).

tooling/cli/src/cmd/analyze_pkp.rs
```diff
@@ -0,0 +1,369 @@
...
+        );
+        println!();
+
+        print_witness_builder_stats(&prover.split_witness_builders);
+
+        Ok(())
+    }
+}
+
+use {
```
Bisht13: Nit: these use blocks (here and line 311) should be moved to the top of the file with the other imports.

tooling/cli/src/cmd/analyze_pkp.rs
```diff
@@ -0,0 +1,369 @@
...
+            elements.push(data.rhs);
+            elements.push(data.and_out);
+            elements.push(data.xor_out);
+        }
+        WitnessBuilder::DigitalDecomposition(_) => {}
+        WitnessBuilder::SpiceMultisetFactor(_, _, _, wc1, _, wc2) => {
+            elements.push(wc1.0);
+            elements.push(wc2.0);
+        }
+        _ => {}
```
Bisht13: This catch-all skips variants that contain `ConstantOrR1CSWitness::Constant(FieldElement)`; e.g., BinOpLookupDenominator, U32Addition, And, Xor, etc. The dedup analysis will undercount. Consider exhaustive matching (compiler catches new variants) or add a comment noting the approximation.

README.md
```diff
@@ -51,11 +51,17 @@ Analyze circuit statistics and R1CS complexity:
...
+```sh
+cargo run --release --bin provekit-cli analyze-pkp ./prover.pkp
+```
+
 Recursively verify in a Gnark proof (reads the proof from `../ProveKit/prover/proof`):
 
 ```sh
 cd ../../recursive-verifier
-go run .
+go run cmd/cli/main.go --config "path_to_config" --r1cs "path_to_r1cs"
```
Bisht13: "path_to_config" and "path_to_r1cs" are placeholders that won't work. Consider using actual paths from the example workflow, e.g.:
```bash
go run cmd/cli/main.go --config ../noir-examples/poseidon-rounds/params_for_recursive_verifier --r1cs ../noir-examples/poseidon-rounds/r1cs.json
```

provekit/common/src/sparse_matrix.rs
```diff
@@ -2,14 +2,54 @@ use {
...
-        fmt::Debug,
+        fmt::{self, Debug},
         ops::{Mul, Range},
     },
 };
-/// A sparse matrix with interned field elements
-#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+
+#[derive(Debug, Clone, Copy)]
+pub struct DeltaEncodingStats {
```
Bisht13, line 17: `DeltaEncodingStats` is a diagnostic type only used by the `analyze-pkp` CLI command, it shouldn't be re-exported from the library's public API. Remove this re-export and import it directly in `analyze_pkp.rs` via `provekit_common::sparse_matrix::DeltaEncodingStats`.

Bisht13 (changes requested):

provekit/common/src/file/bin.rs
ashpect: Kept magicbytes as same. Bump up `FileFormat` version.

PR #277, opened 2026-02-07
perf: Reduce prove peak memory and switch to jemalloc

Bisht13:
### Problem
After adding public inputs, the prove step for `complete_age_check` regressed from **1.84 GB** to **2.24 GB** peak memory.

---

#### Memory optimization (2.24 GB → 1.92 GB, −320 MB)

- Destructured `WhirR1CSCommitment` in both single and dual witness paths to take ownership of masked/random polynomials, enabling explicit `drop()` before entering WHIR's `prove_batch` / `prove`
- Split public input transcript interaction from weight vector allocation (`add_public_inputs_to_transcript` + `build_public_weights`) to defer the 64 MB allocation until after alphas are consumed
- Dropped `program` and `witness_generator` before the prove call since they are only needed during witness generation

#### Switch default allocator to jemalloc (RSS: 2.39 GB → 1.90 GB, −490 MB)

- Added feature-gated jemalloc support to `ProfilingAllocator`, enabled by default
- System allocator remains available via:
```
–no-default-features –features profiling-allocator
```
#### Add `release-fast` build profile

- `cargo build --profile release-fast`
- 30s build time vs 2.5min for full release
- Uses `codegen-units = 16`
- Uses `lto = "thin"`

---

### Benchmark (`complete_age_check`, 1.1M constraints)

| Metric                  | Before  | After  |
|--------------------------|---------|--------|
| Profiling peak           | 2.24 GB | 1.92 GB |
| RSS (system alloc)       | —       | 2.39 GB |
| RSS (jemalloc, default)  | —       | 1.90 GB |

---

### Allocator Comparison

| Allocator  | RSS     | Duration |
|------------|---------|----------|
| System     | 2.39 GB | 3.30s    |
| jemalloc ✅ | 1.90 GB | 3.81s    |
| mimalloc   | 3.12 GB | 3.37s    |

**jemalloc** was chosen as default for best RSS.  
**mimalloc** was evaluated and rejected (worst RSS despite best wall-clock time).

---

### Root Cause Analysis

The remaining ~80 MB gap from the original 1.84 GB is fully accounted for by the public inputs weight vector:

- 64 MB in the statement
- 64 MB cloned inside WHIR's `prove_batch` (line 279, read-only external crate)

Before public inputs, there were 6 weights; after, 7.

This overhead is inherent to the protocol and cannot be reduced without modifying the WHIR crate or changing the proof transcript structure.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -141,6 +141,7 @@ zerocopy = "0.8.25"
 zeroize = "1.8.1"
 xz2 = "0.1.7"
 
+
 # Noir language dependencies
 acir = { git = "https://github.com/noir-lang/noir", rev = "v1.0.0-beta.11" }
 bn254_blackbox_solver = { git = "https://github.com/noir-lang/noir", rev = "v1.0.0-beta.11" }
--- a/provekit/common/src/lazy_r1cs.rs
+++ b/provekit/common/src/lazy_r1cs.rs
@@ -1,0 +1,274 @@
+use {
+    crate::{HydratedSparseMatrix, Interner, SparseMatrix, R1CS},
+    serde::{Deserialize, Deserializer, Serialize, Serializer},
+    std::{io::Read, sync::OnceLock},
+    xz2::{read::XzDecoder, write::XzEncoder},
+};
+
+/// Matrices data that gets compressed/decompressed lazily.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+struct R1CSMatrices {
+    interner: Interner,
+    a:        SparseMatrix,
+    b:        SparseMatrix,
+    c:        SparseMatrix,
+}
+
+/// A lazily-decompressed R1CS constraint system.
+///
+/// Stores the R1CS matrices in XZ-compressed form and only decompresses
+/// them when first accessed. This reduces memory usage during loading since
+/// the compressed data is much smaller than the expanded matrices.
+pub struct LazyR1CS {
+    pub num_public_inputs: usize,
+    num_constraints:       usize,
+    num_witnesses:         usize,
+    compressed_matrices:   Vec<u8>,
+    #[allow(clippy::type_complexity)]
+    cached:                OnceLock<(Interner, SparseMatrix, SparseMatrix, SparseMatrix)>,
+}
+
+impl std::fmt::Debug for LazyR1CS {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("LazyR1CS")
+            .field("num_public_inputs", &self.num_public_inputs)
+            .field("num_constraints", &self.num_constraints)
+            .field("num_witnesses", &self.num_witnesses)
+            .field("compressed_size", &self.compressed_matrices.len())
+            .field("is_decompressed", &self.cached.get().is_some())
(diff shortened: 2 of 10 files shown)


ashpect:
@Paradox Can you please rebase it with main, pr makes quite a few changes which breaks compatibility, such as - 
1. Change in proof format : transcript to narg_string which breaks existing proof
2. In lazy r1cs, using old zstd method.
3. `DomainSeparator`.instance(&empty) causing panic in debug mode (i've pushed commit which fixes this)
4. Gnark verifier : this still expects the old proof format 'transcript' etc. (from 1)

provekit/prover/src/r1cs.rs
```diff
@@ -157,3 +157,106 @@ impl R1CSSolver for R1CS {
         Ok(())
     }
 }
+
+impl R1CSSolver for LazyR1CS {
```
ashpect, line 161: The R1CSSolver for `LazyR1CS` is same as that of the `R1CS` implementation. Instead of the common code, consider extracting into common func, using macros etc.

provekit/common/src/lazy_r1cs.rs
```diff
@@ -0,0 +1,274 @@
...
+        Self {
+            num_public_inputs,
+            num_constraints,
+            num_witnesses,
+            compressed_matrices: compressed,
+            cached: OnceLock::new(),
+        }
+    }
+
+    fn ensure_decompressed(&self) -> &(Interner, SparseMatrix, SparseMatrix, SparseMatrix) {
```
ashpect, line 88: Consider using  `Result<&(...)>` for better logging

provekit/common/src/lazy_r1cs.rs
```diff
@@ -0,0 +1,274 @@
...
+            a:        r1cs.a,
+            b:        r1cs.b,
+            c:        r1cs.c,
+        };
+
+        let serialized =
+            postcard::to_allocvec(&matrices).expect("Failed to serialize R1CS matrices");
+        let mut compressed = Vec::new();
+        {
+            let mut encoder = XzEncoder::new(&mut compressed, 6);
```
ashpect, line 73: In file/bin.rs, the encoding used was xz level 9. it's better to have a global const which is 9 and used here as well

Cargo.toml
```diff
@@ -134,6 +141,7 @@ zerocopy = "0.8.25"
 zeroize = "1.8.1"
 xz2 = "0.1.7"
 
+
```
ashpect, line 144: extra space

provekit/common/src/lazy_r1cs.rs
```diff
@@ -0,0 +1,274 @@
...
+            (matrices.interner, matrices.a, matrices.b, matrices.c)
+        })
+    }
+
+    /// Free the compressed byte buffer once matrices are cached.
+    ///
+    /// After the first access the decompressed matrices live in `cached`,
+    /// so the compressed blob is dead weight. Call this after the first
+    /// access to reclaim ~10 MB for a typical circuit.
+    pub fn free_compressed(&mut self) {
```
ashpect, line 107: Consider adding an assertion in `free_compressed()` to verify cache is populated: `assert!(self.cached.get().is_some(), "Must access matrices before freeing");`

Issue #278, opened 2026-02-09
Witness splitting fails with "error in splitting witnesses into w1 and w2"


x-senpai-x:
The `provekit-cli prepare` command fails with the error "error in splitting witnesses into w1 and w2" when preparing certain Noir circuits (eg. [t_attest](https://github.com/worldfnd/provekit/tree/main/noir-examples/noir-passport-examples/merkle_age_check/t_attest)) that return tuples of public outputs. 

### Error Output

```
Error: while compiling Noir program

Caused by:
error in splitting witnesses into w1 and w2
```

### Reproduction
```bash
cd noir-examples/noir-passport-examples/fragmented/merkle_age_check/t_attest
cargo run --release --bin provekit-cli prepare ../target/t_attest.json  \
  --pkp ../../benchmark-inputs/circuit-prover.pkp \
  --pkv ../../benchmark-inputs/circuit-verifier.pkv

PR #279, opened 2026-02-09
bugfix: sanity_check in witness_split in prepare step

ashpect:
Fixes [278](https://github.com/worldfnd/provekit/issues/278)
- Fix witness splitter failing on complex circuits when builder index doesn't necessarily match ACIR index

### DIFF
--- a/provekit/common/src/witness/scheduling/splitter.rs
+++ b/provekit/common/src/witness/scheduling/splitter.rs
@@ -8,11 +8,32 @@ use {
 
 /// Error returned when witness splitting validation fails.
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct SplitError;
+pub enum SplitError {
+    /// No witness builder exists for a public input ACIR index.
+    NoBuilderForPublicInput { acir_idx: u32 },
+    /// A public input's builder was partitioned into w2 instead of w1.
+    PublicInputNotInW1 {
+        acir_idx:    u32,
+        builder_idx: usize,
+    },
+}
 
 impl fmt::Display for SplitError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "error in splitting witnesses into w1 and w2")
+        match self {
+            Self::NoBuilderForPublicInput { acir_idx } => {
+                write!(f, "no builder for public input ACIR index {acir_idx}")
+            }
+            Self::PublicInputNotInW1 {
+                acir_idx,
+                builder_idx,
+            } => {
+                write!(
+                    f,
+                    "public input ACIR index {acir_idx} (builder {builder_idx}) not in w1"
+                )
+            }
+        }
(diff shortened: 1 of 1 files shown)


provekit/common/src/witness/scheduling/splitter.rs
```diff
@@ -215,13 +213,30 @@ impl<'a> WitnessSplitter<'a> {
...
+                    Some((*acir_idx as u32, builder_idx))
+                } else {
+                    None
+                }
+            })
+            .collect();
+
+        // Sanity check: all public inputs must have builders in w1 (O(P) lookups)
+        for &acir_idx in acir_public_inputs_indices_set {
+            if acir_idx == 0 {
```
Bisht13, line 259: Worth a brief comment explaining why index 0 is skipped, e.g.:
```noir
// ACIR witness 0 is always the constant-one witness, handled
// separately via mandatory_w1.insert(0) above.
if acir_idx == 0 {
    continue;
}
```
Without context, a reader might think this is papering over a bug rather than an intentional invariant.

provekit/common/src/witness/scheduling/splitter.rs
```diff
@@ -215,13 +213,30 @@ impl<'a> WitnessSplitter<'a> {
...
+                continue;
+            }
+            match acir_to_builder.get(&acir_idx) {
```
Bisht13, line 262:
```noir
_ => return Err(SplitError),
```
This silently conflates two distinct failure modes:
- None → no builder exists for this ACIR public input (structural corruption)
- Some(&builder_idx) where builder is NOT in w1 (mispartitioned)

Consider at minimum a debug log, or ideally making SplitError carry context:
```noir
None => return Err(SplitError::NoBuilderForPublicInput(acir_idx)),
Some(&builder_idx) => return Err(SplitError::PublicInputNotInW1 { acir_idx, builder_idx }),
```

provekit/common/src/witness/scheduling/splitter.rs
```diff
@@ -162,7 +162,6 @@ impl<'a> WitnessSplitter<'a> {
         let mut w2_set = mandatory_w2;
 
         for idx in free_builders {
-            // Check if any dependency is in w2
```
Bisht13, line 165: Restore the comment.

provekit/common/src/witness/scheduling/splitter.rs
```diff
@@ -172,7 +171,6 @@ impl<'a> WitnessSplitter<'a> {
 
             let witness_count = DependencyInfo::extract_writes(&self.witness_builders[idx]).len();
 
-            // If free builder writes a public witness, add it to w1_set.
```
Bisht13, line 175: Restore the comment.

Bisht13 (changes requested):

Bisht13 (approved):

PR #283, opened 2026-02-10
Mavros integration

veljkovranic:
## Mavros Compiler Integration

This PR integrates the Mavros compiler as an alternative compilation path for ProveKit.

#### 1. Mavros Compiler Support
- Added `mavros_compiler` feature flag to enable conditional compilation
- Updated `NoirProofScheme`, `Prover`, and related structs with Mavros-specific variants
- Integrated Mavros R1CS generation and witness building

### Feature Flag
The Mavros integration is behind the `mavros_compiler` feature flag. When disabled, the codebase uses the original compilation path.

### Testing
- Existing tests continue to pass with default features
- Mavros path can be tested with `--features mavros_compiler`

### DIFF
--- a/provekit/r1cs-compiler/src/noir_proof_scheme.rs
+++ b/provekit/r1cs-compiler/src/noir_proof_scheme.rs
@@ -134,6 +134,12 @@ impl MavrosCompiler {
             .map(|p| p.typ.field_count() as usize)
             .sum();
 
+        if let Some(ret) = &abi.return_type {
+            if matches!(ret.visibility, AbiVisibility::Public) {
+                num_public_inputs += ret.abi_type.field_count() as usize;
+            }
+        }
+
         let whir_for_witness = WhirR1CSScheme::new_from_mavros_r1cs(
             &mavros_r1cs,
             mavros_r1cs.witness_layout.pre_commitment_size(),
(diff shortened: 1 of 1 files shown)


provekit/r1cs-compiler/src/whir_r1cs.rs
```diff
@@ -74,6 +91,51 @@ impl WhirR1CSSchemeBuilder for WhirR1CSScheme {
...
+        w1_size: usize,
+        num_challenges: usize,
+        has_public_inputs: bool,
+    ) -> Self {
+        let m_raw = next_power_of_two(num_witnesses);
+        let m0_raw = next_power_of_two(num_constraints);
+
+        let m = m_raw.max(MIN_WHIR_NUM_VARIABLES);
+        let m_0 = m0_raw.max(MIN_SUMCHECK_NUM_VARIABLES);
+
```
ysh-red, line 128: Missing the overflow check from new_for_r1cs : if (1 << m_raw) - w1_size < 4 * m_0 { m_raw += 1 }.

provekit/r1cs-compiler/src/noir_proof_scheme.rs
```diff
@@ -80,35 +70,94 @@ impl NoirProofSchemeBuilder for NoirProofScheme {
...
+            false,
+        );
+        let r1cs = convert_mavros_r1cs_to_provekit(&mavros_r1cs);
+
+        let num_public_inputs: usize = abi
+            .parameters
+            .iter()
+            .filter(|p| p.is_public())
+            .map(|p| p.typ.field_count() as usize)
+            .sum();
```
ysh-red: This doesn't account for abi.return_type. Public returns live as a public input in abi.return_type, not abi.parameters. The Noir path does not need this cause it's reading from program.functions[0].public_inputs().indices() which covers both. Since the Mavros path doesn't have the ACIR program, abi.return_type needs to be checked here too.

veljkovranic: great catch!

PR #284, opened 2026-02-12
Witness Optimizations for SHA256: Dynamic AND/XOR Bit-Width + Spread-Based Binops

rose2221:
This PR implements two major witness-count optimizations for the R1CS compiler targeting SHA256 circuits:

### 1. Dynamic Bit-Width for Combined AND/XOR Lookup Table

Introduces a cost model that searches over candidate atomic widths {2, 4, 8} and selects the width minimizing total witness count (table + decomposition + complementary + overhead)
Currently only applies to 8-bit operands; wider operands fall back to w=8

### 2. Spread Trick for SHA256 Bitwise Operations

Replaces the generic combined AND/XOR table approach for SHA256 with the spread-based technique from Eli Ben-Sasson et al.

**Key idea:** "spreading" a value by interleaving zeros between bits (`0b1011` → `0b01000101`) converts bitwise XOR/AND into field addition/multiplication, eliminating per-operation lookup costs

**Why ROTR/SHR are free:** SHA256 decomposes each 32-bit word into chunks aligned to rotation boundaries. For example, `Sigma_0` uses chunks `[2, 11, 9, 10`] bits exactly matching ROTR2, ROTR13, ROTR22. A rotation is just reading the same chunks in a different order with adjusted coefficients: ROTR2(a) reads chunks starting from index 1 instead of index 0. No new witnesses are allocated — the spread values computed during decomposition are reused with permuted coefficients. Similarly, SHR drops the lowest chunks entirely. The cost of rotation/shift is zero witnesses, zero constraints.

**SHA256 operations implemented via spread:** `sigma_0`, `sigma_1`, `Sigma_0`, `Sigma_1`, `Ch`, `Maj`, message schedule, compression rounds
- Uses a fixed 8-bit spread table (256 entries, 768 witnesses for LogUp)
- Multi-operand `u32` addition with single carry witness per addition

### SHA256 (35 compression calls )

| Metric      | Before    | After (Spread) | Reduction |
| ----------- | --------- | -------------- | --------- |
| Constraints | 911,769   | 445,739        | -51.1%    |
| Witnesses   | 1,588,306 | 1,033,576     |  -34.9%  |

### SHA256 (1 compression call)

| Metric      | Before  | Spread | Dynamic AND/XOR |
| ----------- | ------- | ------ | --------------- |
| Constraints | 86,607  |  14389  | 44,506          |
| Witnesses   | 231,044 | 31680 | 68,251          |

## Known Soundness Limitation (for M31)

- Carry witness in u32 spread addition is range-checked to [0, 255] sound for bn254, unsound for M31. See #285
- Collision probability for fused-constraint spread table LogUp. See #286

## Future Work

Dynamic spread table width: Currently fixed at w=8 (256 entries). A dynamic search over {4, 8, 16} could further reduce witnesses for circuits with fewer SHA256 calls

Shared lookup table: The spread table implicitly range-checks values to [0, 2^w - 1]. Merging with the existing range-check system would eliminate redundant table entries and save additional witnesses.

### DIFF
--- a/provekit/common/src/witness/mod.rs
+++ b/provekit/common/src/witness/mod.rs
@@ -34,8 +34,6 @@ pub const WITNESS_ONE_IDX: usize = 0;
 
 /// Compute spread(val): interleave bits of val with zeros.
 /// E.g., `0b1011` → `0b01_00_01_01`.
-/// Used by both the compiler (to build table entries) and the prover
-/// (to solve SpreadWitness builders).
 pub fn compute_spread(val: u64) -> u64 {
     let mut result = 0u64;
     for i in 0..32 {


noir-examples/noir_sha256/src/main.nr
```diff
@@ -11,5 +11,6 @@ fn main() {
     ];
 
     let digest = sha256_var(bytes, 32);
-    assert(digest == result);
+   // let r = sha256_var(digest, 32);
```
Bisht13: Assert is commented out, this example computes a SHA256 digest but never checks the result. Please uncomment.

provekit/r1cs-compiler/src/spread.rs
```diff
@@ -0,0 +1,564 @@
...
+pub(crate) const BYTE_CHUNKS: [u32; 4] = [8, 8, 8, 8];
+
+/// Max sub-chunk bit width for spread table
+const SPREAD_TABLE_BITS: u32 = 8;
+
+/// spread(0xFFFFFFFF) = 0x5555555555555555
+const SPREAD_ALL_ONES: u64 = 0x5555_5555_5555_5555;
+
+/// Pure function: compute spread(val) — interleave bits with zeros.
+pub(crate) fn compute_spread(val: u64) -> u64 {
```
Bisht13: This function is duplicated identically in `provekit/prover/src/witness/witness_builder.rs:423`. The compiler uses one copy to build constraints, the prover uses the other to compute witnesses. If either copy is modified without updating the other, proofs break silently. Please move to `provekit-common`.

provekit/r1cs-compiler/src/sha256_compression.rs
```diff
@@ -27,635 +25,193 @@ const SHA256_K: [u32; 64] = [
...
-    and_ops: &mut Vec<(ConstantOrR1CSWitness, ConstantOrR1CSWitness, usize)>,
-    xor_ops: &mut Vec<(ConstantOrR1CSWitness, ConstantOrR1CSWitness, usize)>,
-    range_checks: &mut BTreeMap<u32, Vec<usize>>,
     inputs_and_outputs: Vec<(
         Vec<ConstantOrR1CSWitness>,
         Vec<ConstantOrR1CSWitness>,
         Vec<usize>,
     )>,
-) {
+) -> BTreeMap<u32, Vec<usize>> {
```
Bisht13, line 207: The spread approach requires p >> 2^64. The 3-way Maj spread sum reaches 3 × 0x5555555555555555 = 2^64 − 1, spread coefficients reach 4^31 = 2^62, and carry range checking needs 255 × 2^32 < p. All of these break silently for small fields (e.g. M31 with p ≈ 2^31). There's no guard anywhere. Consider:
```rs
assert!(FieldElement::MODULUS_BIT_SIZE > 64, "Spread-based SHA256 requires field modulus > 2^64");
```

Bisht13 (changes requested):

noir-examples/noir_sha256/src/main.nr
rose2221: done.

provekit/r1cs-compiler/src/spread.rs
rose2221: done!

provekit/r1cs-compiler/src/sha256_compression.rs
rose2221: done, I've added the `assert!(MODULUS_BIT_SIZE > 64)` guard in this PR and opened #285 to track the M31 soundness concerns.

Bisht13 (approved):

PR #294, opened 2026-02-18
Reduce proving memory

Bisht13:
## Summary

- **Compress idle data**: R1CS matrices and w2 witness-builder layers are serialized into postcard blobs while idle during the commit phase, then decompressed on demand. This frees significant heap that would otherwise sit unused.
- **PrefixCovector**: A `LinearForm` implementation that stores only the non-zero power-of-two prefix of a covector, avoiding zero-padding to full domain size. Used for R1CS alpha weights and public-input weights.
- **Ownership transfer**: Integrate [whir PR 225](https://github.com/WizardOfMenlo/whir/pull/225)'s `Cow`-based `prove()` API to pass witnesses and evaluations by ownership (`Cow::Owned`) instead of cloning.
- **In-place allocation**: Extend masked polynomial in-place instead of allocating a new vec; drop program, witness generator, and ACIR map as early as possible.

## Benchmark (complete age check circuit, `release-fast`)

| Metric | `main` | This PR | Delta |
|--------|--------|---------|-------|
| Peak memory | 1.97 GB | 1.19 GB | **-40%** |
| Total time | 7.04s | 2.86s | **-59%** |

## Test plan

- [x] `provekit-cli prove` succeeds on complete age check circuit
- [x] Proof verifies correctly
- [x] Memory trace confirms reduced peak allocation

### DIFF
--- a/provekit/prover/src/lib.rs
+++ b/provekit/prover/src/lib.rs
@@ -1,5 +1,3 @@
-#[cfg(test)]
-use crate::r1cs::R1CSSolver;
 use {
     crate::{
         r1cs::{CompressedLayers, CompressedR1CS},
@@ -98,8 +96,15 @@ impl Prove for Prover {
             );
         }
 
-        let compressed_w2_layers =
-            CompressedLayers::compress(self.split_witness_builders.w2_layers);
+        let has_challenges = self.whir_for_witness.num_challenges > 0;
+        let compressed_w2_layers = if has_challenges {
+            Some(CompressedLayers::compress(
+                self.split_witness_builders.w2_layers,
+            ))
+        } else {
+            drop(self.split_witness_builders.w2_layers);
+            None
+        };
 
         debug!(
             witness_heap_bytes = witness.capacity() * size_of::<Option<FieldElement>>(),
@@ -120,8 +125,8 @@ impl Prove for Prover {
             .commit(&mut merlin, num_witnesses, num_constraints, w1, true)
             .context("While committing to w1")?;
 
-        let commitments = if self.whir_for_witness.num_challenges > 0 {
-            let w2_layers = compressed_w2_layers.decompress();
+        let commitments = if has_challenges {
+            let w2_layers = compressed_w2_layers.unwrap().decompress();
             {
                 let _s = info_span!("solve_w2").entered();
                 crate::r1cs::solve_witness_vec(
@@ -156,8 +161,11 @@ impl Prove for Prover {
         let r1cs = compressed_r1cs.decompress();
 
         #[cfg(test)]
-        r1cs.test_witness_satisfaction(&witness.iter().map(|w| w.unwrap()).collect::<Vec<_>>())
-            .context("While verifying R1CS instance")?;
+        crate::r1cs::test_witness_satisfaction(
+            &r1cs,
+            &witness.iter().map(|w| w.unwrap()).collect::<Vec<_>>(),
+        )
+        .context("While verifying R1CS instance")?;
 
         let public_inputs = if num_public_inputs == 0 {
             PublicInputs::new()
--- a/provekit/prover/src/r1cs.rs
+++ b/provekit/prover/src/r1cs.rs
@@ -61,11 +61,6 @@ impl CompressedR1CS {
     }
 }
 
-#[cfg(test)]
-pub trait R1CSSolver {
-    fn test_witness_satisfaction(&self, witness: &[FieldElement]) -> anyhow::Result<()>;
-}
-
 /// Solves the R1CS witness vector using layered execution with batch
 /// inversion.
 ///
@@ -88,7 +83,7 @@ pub trait R1CSSolver {
 /// This indicates a bug in the layer scheduling algorithm.
 #[instrument(skip_all)]
 pub fn solve_witness_vec(
-    witness: &mut Vec<Option<FieldElement>>,
+    witness: &mut [Option<FieldElement>],
     plan: LayeredWitnessBuilders,
     acir_map: &WitnessMap<NoirElement>,
     transcript: &mut ProverState<TranscriptSponge>,
@@ -196,30 +191,25 @@ pub fn solve_witness_vec(
 }
 
 #[cfg(test)]
-impl R1CSSolver for R1CS {
-    // Tests R1CS Witness satisfaction given the constraints provided by the
-    // R1CS Matrices.
-    #[instrument(skip_all, fields(size = witness.len()))]
-    fn test_witness_satisfaction(&self, witness: &[FieldElement]) -> anyhow::Result<()> {
-        use anyhow::ensure;
+#[instrument(skip_all, fields(size = witness.len()))]
+pub fn test_witness_satisfaction(r1cs: &R1CS, witness: &[FieldElement]) -> anyhow::Result<()> {
+    use anyhow::ensure;
 
-        ensure!(
-            witness.len() == self.num_witnesses(),
-            "Witness size does not match"
-        );
+    ensure!(
+        witness.len() == r1cs.num_witnesses(),
+        "Witness size does not match"
+    );
 
-        // Verify
-        let a = self.a() * witness;
-        let b = self.b() * witness;
-        let c = self.c() * witness;
-        for (row, ((a_val, b_val), c_val)) in a
-            .into_iter()
-            .zip(b.into_iter())
-            .zip(c.into_iter())
-            .enumerate()
-        {
-            ensure!(a_val * b_val == c_val, "Constraint {row} failed");
-        }
-        Ok(())
+    let a = r1cs.a() * witness;
+    let b = r1cs.b() * witness;
+    let c = r1cs.c() * witness;
+    for (row, ((a_val, b_val), c_val)) in a
+        .into_iter()
+        .zip(b.into_iter())
+        .zip(c.into_iter())
+        .enumerate()
+    {
+        ensure!(a_val * b_val == c_val, "Constraint {row} failed");
     }
+    Ok(())
 }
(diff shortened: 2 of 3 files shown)


provekit/prover/src/r1cs.rs
```diff
@@ -10,155 +10,210 @@ use {
...
+        postcard::from_bytes(&self.blob).expect("LayeredWitnessBuilders deserialization failed")
+    }
+}
+
+impl CompressedR1CS {
+    #[must_use]
+    pub fn compress(r1cs: R1CS) -> Self {
+        let num_constraints = r1cs.num_constraints();
+        let num_witnesses = r1cs.num_witnesses();
+        let blob = postcard::to_allocvec(&r1cs).expect("R1CS serialization failed");
```
ashpect: return result and not panic

provekit/prover/src/whir_r1cs.rs
```diff
@@ -10,39 +10,116 @@ use {
...
+impl LinearForm<FieldElement> for PrefixCovector {
+    fn size(&self) -> usize {
+        self.logical_size
+    }
+
+    fn deferred(&self) -> bool {
+        self.deferred
+    }
+
+    fn mle_evaluate(&self, point: &[FieldElement]) -> FieldElement {
```
ashpect, line 79: can you add some tests for this ?

provekit/prover/src/r1cs.rs
```diff
@@ -10,155 +10,210 @@ use {
...
-    fn solve_witness_vec(
-        &self,
-        witness: &mut Vec<Option<FieldElement>>,
-        plan: LayeredWitnessBuilders,
-        acir_map: &WitnessMap<NoirElement>,
-        transcript: &mut ProverState<TranscriptSponge>,
-    );
-
-    #[cfg(test)]
     fn test_witness_satisfaction(&self, witness: &[FieldElement]) -> anyhow::Result<()>;
```
ashpect: `solve_witness_vec` is now a free function and the trait only contains `test_witness_satisfaction`, seems unconventional, can be simplified.

provekit/prover/src/lib.rs
```diff
@@ -70,59 +83,91 @@ impl Prove for Prover {
...
+                self.split_witness_builders.w1_layers,
+                &acir_witness_idx_to_value_map,
+                &mut merlin,
+            );
+        }
+
+        // Compress w2 layers to free memory during w1 commit (only when
+        // challenges exist; otherwise just drop them).
+        let has_challenges = self.whir_for_witness.num_challenges > 0;
+        let compressed_w2_layers = if has_challenges {
```
ashpect, line 100: its correct but better way of writing :

let compressed_w2_layers = has_challenges.then(|| {
    CompressedLayers::compress(self.split_witness_builders.w2_layers)
});
if !has_challenges {
    drop(self.split_witness_builders.w2_layers);
}

provekit/prover/src/lib.rs
```diff
@@ -70,59 +83,91 @@ impl Prove for Prover {
...
         let mut merlin = ProverState::new(&ds, TranscriptSponge::default());
 
-        let mut witness: Vec<Option<FieldElement>> = vec![None; self.r1cs.num_witnesses()];
+        let mut witness: Vec<Option<FieldElement>> = vec![None; num_witnesses];
+
+        // Solve w1 (or all witnesses if no challenges).
+        // Outer span captures memory AFTER w1_layers parameter is freed
+        // (parameter drop happens before outer span close).
+        {
+            let _s = info_span!("solve_w1").entered();
```
ashpect, line 90: `info_span!` creates allocations even when tracing is disabled at runtime. We should put it behind #[cfg(feature = "profiling")]`

provekit/prover/src/whir_r1cs.rs
```diff
@@ -227,43 +279,49 @@ impl WhirR1CSProver for WhirR1CSScheme {
...
-            drop(alphas_2);
 
             // Compute cross-evaluations: weights_1 on c2's polynomials and
             // weights_2 on c1's polynomials. Whir's prove() expects evaluations
             // for ALL (weight, polynomial) pairs in row-major order.
-            let c1m_evals = coeffs_to_evals(&c1.masked_polynomial_coeff);
-            let c1r_evals = coeffs_to_evals(&c1.random_polynomial_coeff);
-            let c2m_evals = coeffs_to_evals(&c2.masked_polynomial_coeff);
-            let c2r_evals = coeffs_to_evals(&c2.random_polynomial_coeff);
             let cross_f_12: Vec<FieldElement> = weights_1
```
ashpect: we can parallelize all of these cross ealuations

provekit/prover/src/r1cs.rs
```diff
@@ -10,155 +10,210 @@ use {
     whir::transcript::ProverState,
 };
 
+/// Serialized R1CS matrices held as a compact postcard blob.
+///
+/// The R1CS is only needed during the sumcheck phase. Compressing it
+/// into a serialized blob frees that memory during the commit phase,
+/// then decompresses when the sumcheck begins.
+pub struct CompressedR1CS {
```
ashpect, line 13: we should add tests for compression and decompression round trip.

for eg:
#[test]
fn test_compressed_r1cs_roundtrip() {
    let r1cs = <create_test_r1cs>
    let original_num_witnesses = r1cs.num_witnesses();
    let compressed = CompressedR1CS::compress(r1cs);
    assert_eq!(compressed.num_witnesses(), original_num_witnesses);
    let decompressed = compressed.decompress();
    assert_eq!(decompressed.num_witnesses(), original_num_witnesses);
}

this would help if serde or postcard version becme incompatible

provekit/common/src/utils/zk_utils.rs
```diff
@@ -1,37 +1,12 @@
...
-/// Dot product of a covector's weight vector against pre-transformed
-/// evaluations.
-pub fn covector_dot(w: &Covector<FieldElement>, evals: &[FieldElement]) -> FieldElement {
-    dot(&w.vector, evals)
-}
+use {crate::FieldElement, ark_ff::UniformRand, rayon::prelude::*};
 
 pub fn create_masked_polynomial(
-    original: EvaluationsList<FieldElement>,
+    mut original: Vec<FieldElement>,
```
ashpect: This is a nice. (as it was returning evaluations list but new whir api needed coeffs list and hence was getting converted) But instead of working on vec, we couls directly use coeffs list to switch from evals list, it's better.

provekit/prover/src/lib.rs
```diff
@@ -70,59 +83,91 @@ impl Prove for Prover {
...
+                    &mut witness,
+                    w2_layers,
+                    &acir_witness_idx_to_value_map,
+                    &mut merlin,
+                );
+            }
+            drop(acir_witness_idx_to_value_map);
+
+            let w2 = {
+                let _s = info_span!("allocate_w2").entered();
```
ashpect, line 142: same as the above info span comment

provekit/prover/src/whir_r1cs.rs
```diff
@@ -10,39 +10,116 @@ use {
...
+}
+
+/// A covector that stores only a power-of-two prefix, with the rest
+/// implicitly zero-padded to `logical_size`. Saves memory when the
+/// covector is known to be zero beyond the prefix (e.g. R1CS alpha
+/// weights that are zero-padded from 2^(m-1) to 2^m).
+///
+/// Implements [`LinearForm`] so it can be passed directly to whir's
+/// `prove()` in place of a full-length `Covector`.
+struct PrefixCovector {
```
ashpect, line 47: Verifier still uses covector and full allocation. Can we check if its possible to optimize on verifier side as well ?

provekit/prover/src/whir_r1cs.rs
```diff
@@ -10,39 +10,116 @@ use {
...
+        // So the r = len - k leading variables must all be 0 for the
+        // prefix region, contributing a factor of ∏(1 − pⱼ) for j < r.
+        let r = point.len() - k;
+        let head_factor: FieldElement =
+            point[..r].iter().map(|p| FieldElement::one() - p).product();
+        let prefix_mle = multilinear_extend(&self.vector, &point[r..]);
+        head_factor * prefix_mle
+    }
+
+    fn accumulate(&self, accumulator: &mut [FieldElement], scalar: FieldElement) {
```
ashpect, line 93: add ` debug_assert!(accumulator.len() >= self.vector.len())`

provekit/prover/src/whir_r1cs.rs
ashpect, line 47: Similarily, iirc get_public_weights also allocate full Covector. We should look into all places where we can optimise using prefixcovector

provekit/prover/src/r1cs.rs
Bisht13: `postcard::to_allocvec` on our own Serde types is infallible, the only failure mode would be a broken Serialize impl, which is a programmer bug. expect with a descriptive message is idiomatic Rust for invariants that should never be violated. Returning Result here would force callers to handle an error that can't actually occur.

provekit/prover/src/lib.rs
Bisht13: tracing's info_span! checks dispatcher::has_been_set() first, when no subscriber is active it returns Span::none() with zero allocation. The overhead without a subscriber is a single atomic load + branch. This is tracing's core design guarantee. Same applies to #[instrument] on the function itself.

provekit/prover/src/lib.rs
Bisht13: Same as above, info_span! is zero-cost without an active subscriber (single atomic load + branch, no allocation).

provekit/prover/src/whir_r1cs.rs
Bisht13: verifier-side optimization is being tracked in https://github.com/worldfnd/provekit/pull/296.

Bisht13: get_public_weights already returns a PrefixCovector (see line 791), it only allocates the non-zero prefix, not a full-domain covector.

provekit/common/src/utils/zk_utils.rs
Bisht13: `CoefficientList` and `EvaluationsList` were removed in the latest whir, they no longer exist upstream.

PR #295, opened 2026-02-18
Dynamic spread table width + Ch algebraic optimization

rose2221:
# Summary
- Replaces the hardcoded 8-bit spread table with a cost-model-driven dynamic width selection. A new `calculate_spread_witness_cost` function estimates total witness cost across table overhead, inline witnesses, and LogUp queries for each candidate width (2–20 bits), and `get_optimal_spread_width` picks the minimum. All spread internals are generalized from fixed 4-byte chunks to variable-width chunks (`subchunks`, `spread_decompose`, `pack_chunks`, `spread_from_chunk_spreads`). The chosen width is displayed in `circuit_stats` output.

- Reduces `Ch(e,f,g)` from 3 spread_decompose calls to 2 + one algebraic constraint, using the identity `Ch = (e&f) + g - (e&g)`. Saves ~39 witnesses/round (~2,496/compression) from eliminated spread decomposition and lookups. Dead code removed: `spread_not_terms`, `spread_from_chunk_spreads`, `SPREAD_ALL_ONES.`
### 35 SHA256 compressions (spread width: 8-bit → 16-bit)

| Metric       | Before    | After     | Reduction |
|--------------|-----------|-----------|-----------|
| Constraints  | 445,739   | 328,739 | -26.3% |
| Witnesses    | 1,033,576 | 705,076 | -31.8%|

### 1 SHA256 compression (spread width: 8-bit → 11-bit)

| Metric       | Before | After  | Reduction |
|--------------|--------|--------|-----------|
| Constraints  | 14,389 | 11,561 | -19.7% |
| Witnesses    | 31,680 | 25,475 | -19.6% |

### DIFF
--- a/provekit/r1cs-compiler/src/spread.rs
+++ b/provekit/r1cs-compiler/src/spread.rs
@@ -604,10 +604,21 @@ pub(crate) fn calculate_spread_witness_cost(w: u32, n_sha: usize) -> usize {
     table + n_sha * (inline + 3 * lookups)
 }
 
-/// Find the spread table width in [2, 20] minimizing total witness
+/// Find the spread table width in [3, 20] minimizing total witness
 /// count for `n_sha` SHA256 compression calls.
 pub(crate) fn get_optimal_spread_width(n_sha: usize) -> u32 {
-    (2u32..=20)
+    (3u32..=20)
         .min_by_key(|&w| calculate_spread_witness_cost(w, n_sha))
         .unwrap()
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_optimal_spread_width() {
+        assert_eq!(get_optimal_spread_width(1), 11);
+        assert_eq!(get_optimal_spread_width(35), 16);
+    }
 }


provekit/r1cs-compiler/src/spread.rs
```diff
@@ -553,3 +542,72 @@ pub(crate) fn add_spread_table_constraints(
...
+    let table = 2 * (1usize << w) + 4;
+
+    // Total: table + n_sha × (inline + 3 × lookups)
+    table + n_sha * (inline + 3 * lookups)
+}
+
+/// Find the spread table width in [2, 20] minimizing total witness
+/// count for `n_sha` SHA256 compression calls.
+pub(crate) fn get_optimal_spread_width(n_sha: usize) -> u32 {
+    (2u32..=20)
```
Bisht13: The lower bound should be 3. The worst-case carry in SHA256 is 6 (the new_a addition at line 331: 6 u32 inputs + Ki), so the carry table lookup requires 2^table_bits > 6 → table_bits ≥ 3. With table_bits = 2 the honest prover can't generate a valid proof.

provekit/r1cs-compiler/src/spread.rs
```diff
@@ -553,3 +542,72 @@ pub(crate) fn add_spread_table_constraints(
...
+    // Table: 2×2^w multiplicities + 2 challenges + 2 sums
+    let table = 2 * (1usize << w) + 4;
+
+    // Total: table + n_sha × (inline + 3 × lookups)
+    table + n_sha * (inline + 3 * lookups)
+}
+
+/// Find the spread table width in [2, 20] minimizing total witness
+/// count for `n_sha` SHA256 compression calls.
+pub(crate) fn get_optimal_spread_width(n_sha: usize) -> u32 {
```
Bisht13, line 609: Could you add a unit test?
```rs
#[test]
fn test_optimal_spread_width() {
    assert_eq!(get_optimal_spread_width(1), 11);
    assert_eq!(get_optimal_spread_width(35), 16);
}
```

rose2221: done.

provekit/r1cs-compiler/src/spread.rs
rose2221: the cost model already prevents this w=2 has ~4× higher witness cost than w=8 so it’s never selected in practice. But agreed this should be enforced defensively, i’ve updated the lower bound to 3.

PR #296, opened 2026-02-20
feat: port provekit to zkWHIR 2.0

Bisht13:
## Summary

Ports provekit to the updated whir protocol ([WizardOfMenlo/whir#226](https://github.com/WizardOfMenlo/whir/pull/226)), where zk-critical code now lives upstream in the whir crate. Includes memory optimizations, verification speedups, and codebase cleanup.

### WHIR protocol migration
- Pin `whir` dependency to git rev `9ee8885`
- Adapt prover and verifier to the new `whir_zk` API (`Cow`-based `prove()`, `Box<dyn LinearForm>` weights)
- Deduplicate zk-critical code that now lives upstream
- Bump `MIN_WHIR_NUM_VARIABLES` from 12 to 13 — the new `whir_zk` blinding assertion requires `num_blinding_variables < num_witness_variables`; at 128-bit security with folding factor 4, blinding yields exactly 12 variables
- Verifier takes `&NoirProof` / `&WhirR1CSProof` instead of consuming by value

### Memory & performance
- **Reduce proving memory**: compress idle R1CS/witness data between phases via `CompressedR1CS` / `CompressedLayers`, eliminate clones, introduce `PrefixCovector` for zero-padded weight vectors
- **Parallelize sparse matrix right-multiplication** over rows via `into_par_iter`
- **Optimize verification path**: overlap R1CS transpose with sumcheck, parallel file I/O, parallel sort, eliminate proof clone (~2s → ~510ms on `complete_age_check`)
- **Parameterize file compression**: zstd for `.pkv` / `.np` (fast), xz for `.pkp` (smaller). Reader auto-detects format via magic bytes, streams instead of buffering entire file
- **Switch serde to base64**: 33% overhead vs hex's 100% — cuts proof file size ~25% (deserializer auto-detects hex for backwards compatibility)
- **Optimize sumcheck allocations**: in-place resize instead of `pad_to_power_of_two` re-allocs

### Bug fixes
- **Fix verifier double-call panic**: `.take().unwrap()` → `.take().context()` with descriptive error
- **Replace panic-prone unwraps** in verifier proof parsing with proper `Result` propagation
- **Sort public input indices** for deterministic ordering
- **Fix signed integer two's complement display** in show-inputs

### Code quality
- **Extract `PrefixCovector`** and 6 shared helper functions to `provekit-common` (was duplicated across prover/verifier)
- **Deduplicate AND/XOR opcode handling**: ~250 lines of near-identical match arms → single `process_binop_opcode` helper
- **Add `#[must_use]`** to all public constructors and accessors
- **Improve prover/verifier construction** with destructuring
- Remove dead code, empty test modules, stale comments

### Features
- `show-inputs` CLI command with nested struct field name display
- Enable `profiling-allocator` as default CLI feature

## Upstream dependency
Depends on [WizardOfMenlo/whir#226](https://github.com/WizardOfMenlo/whir/pull/226).

### DIFF
--- a/provekit/r1cs-compiler/src/whir_r1cs.rs
+++ b/provekit/r1cs-compiler/src/whir_r1cs.rs
@@ -6,7 +6,7 @@ use {
 };
 
 // Minimum log2 of the WHIR evaluation domain (lower bound for m).
-const MIN_WHIR_NUM_VARIABLES: usize = 14;
+const MIN_WHIR_NUM_VARIABLES: usize = 13;
 // Minimum number of variables in the sumcheck's multilinear polynomial (lower
 // bound for m_0).
 const MIN_SUMCHECK_NUM_VARIABLES: usize = 1;


provekit/common/src/file/bin.rs
```diff
@@ -28,14 +32,8 @@ pub fn write_bin<T: Serialize>(
...
 
-    let mut compressed_data = Vec::new();
-    {
-        let mut encoder = XzEncoder::new(&mut compressed_data, 9);
-        encoder
-            .write_all(&postcard_data)
-            .context("while compressing with XZ")?;
-        encoder.finish().context("while finishing XZ compression")?;
-    }
+    let compressed_data =
```
ashpect: Are we defaulting to zstd again for all artifacts ?
`decompress_bin` below already detects format via magic bytes, but this is zstd-only. Should we parameterize the compression format here? 
Also we can open an issue to add support of using different compressions (if not implemented already).

provekit/common/src/utils/sumcheck.rs
```diff
@@ -102,29 +102,48 @@ fn sumcheck_fold_map_reduce_inner<const N: usize, const M: usize>(
     }
 }
 
-/// List of evaluations for eq(r, x) over the boolean hypercube
+/// List of evaluations for eq(r, x) over the boolean hypercube, truncated to
+/// `num_entries` elements. When `num_entries < 2^r.len()`, avoids allocating
+/// the full hypercube.
 #[instrument(skip_all)]
 pub fn calculate_evaluations_over_boolean_hypercube_for_eq(
```
ashpect, line 111: can we have tests for evals functions ? or make an issue tracking them

provekit/common/src/witness/mod.rs
```diff
@@ -45,6 +45,7 @@ pub enum ConstantOrR1CSWitness {
 }
 
 impl ConstantOrR1CSWitness {
+    #[must_use]
```
ashpect: Why must use ?


ASSASASAS

provekit/common/src/lib.rs
```diff
@@ -18,22 +19,20 @@ pub use {
...
         WhirConfig, WhirDomainSeparator, WhirProof, WhirProverState, WhirR1CSProof, WhirR1CSScheme,
+        WhirZkConfig,
     },
     witness::PublicInputs,
 };
 
 /// SHA-256 based transcript sponge for Fiat-Shamir.
 pub type TranscriptSponge = spongefish::instantiations::SHA256;
 
-/// Register provekit's custom implementations in whir's global registries.
```
ashpect, line 33: Restore this comment.

provekit/common/src/prefix_covector.rs
```diff
@@ -0,0 +1,163 @@
...
+        );
+        Self {
+            vector,
+            logical_size,
+        }
+    }
+
+    /// Access the underlying prefix vector.
+    #[must_use]
+    pub fn vector(&self) -> &[FieldElement] {
```
ashpect, line 47: it would benefit to add a `len` function as well

provekit/common/src/prefix_covector.rs
```diff
@@ -0,0 +1,163 @@
...
+        }
+    }
+}
+
+/// Expand a slice of field elements into `[1, x, x², x³]` tuples.
+///
+/// Used to build weight vectors for the spartan blinding polynomial
+/// evaluation in both prover and verifier.
+#[must_use]
+pub fn expand_powers(values: &[FieldElement]) -> Vec<FieldElement> {
```
ashpect: Use a const generic here instead of hardcoding the powers to be expanded till

provekit/common/src/prefix_covector.rs
```diff
@@ -0,0 +1,163 @@
...
+/// Builds the vector `[1, x, x², …, x^{n-1}]` padded to a power of two,
+/// where `n = public_inputs_len`.
+#[must_use]
+pub fn make_public_weight(x: FieldElement, public_inputs_len: usize, m: usize) -> PrefixCovector {
+    let domain_size = 1 << m;
+    let prefix_len = public_inputs_len.next_power_of_two().max(2);
+    let mut public_weights = vec![FieldElement::zero(); prefix_len];
+
+    let mut current_pow = FieldElement::one();
+    for slot in public_weights.iter_mut().take(public_inputs_len) {
```
ashpect, line 120: After above comment resolution, we can use the above generic expand_powers here on value 1

provekit/common/src/prefix_covector.rs
```diff
@@ -0,0 +1,163 @@
...
+/// covector is known to be zero beyond the prefix (e.g. R1CS alpha
+/// weights that are zero-padded from witness_size to 2^m).
+///
+/// Implements [`LinearForm`] so it can be passed directly to whir's
+/// `prove()` / `verify()` in place of a full-length `Covector`.
+pub struct PrefixCovector {
+    /// The non-zero prefix. Length must be a power of two.
+    vector:       Vec<FieldElement>,
+    /// The full logical domain size (also a power of two, >= vector.len()).
+    logical_size: usize,
```
ashpect: suggestion : domain size is a better variable name

provekit/common/src/prefix_covector.rs
```diff
@@ -0,0 +1,163 @@
...
+        }
+    }
+
+    /// Access the underlying prefix vector.
+    #[must_use]
+    pub fn vector(&self) -> &[FieldElement] {
+        &self.vector
+    }
+}
+
```
ashpect, line 63: Add hyperlink to github/whir's linear form so interested people can read comments for the functions. Better than writing comments for below functions

provekit/common/src/whir_r1cs.rs
```diff
@@ -3,10 +3,14 @@ use whir::transcript::Interaction;
...
     crate::{utils::serde_hex, FieldElement},
     serde::{Deserialize, Serialize},
-    whir::{protocols::whir::Config as GenericWhirConfig, transcript},
+    whir::{
+        protocols::{whir::Config as GenericWhirConfig, whir_zk::Config as GenericWhirZkConfig},
+        transcript,
+    },
 };
 
 pub type WhirConfig = GenericWhirConfig<FieldElement>;
```
ashpect, line 14: WhirConfig would turn redundant once gnark is updated to use WhirzkConfig, add a comment to make sure we remove it then.

provekit/prover/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#[cfg(test)]
```
ashpect: Confirm THIS

AAAAAAAAA

provekit/prover/src/whir_r1cs.rs
```diff
@@ -272,132 +177,70 @@ impl WhirR1CSProver for WhirR1CSScheme {
...
-                )
+            // Phase 1 — compute all transcript hints (cheap dot products,
+            // no PrefixCovector allocation).  Both polynomials must be
+            // accessible here but we defer weight construction.
+            let evals_1 = compute_alpha_evals(&c1.polynomial, &alphas_1);
+            let evals_2 = compute_alpha_evals(&c2.polynomial, &alphas_2);
+            merlin.prover_hint_ark(&evals_1);
+            merlin.prover_hint_ark(&evals_2);
+
+            let (public_1, public_2) = if !public_inputs.is_empty() {
```
ashpect: public inputs are always inside `w1` as made sure in prepare step

c1.polynomial = witness[0..w1_size] , contains public inputs at indices 0..public_inputs.len()
but  p2 is not a public input evaluation, it's evaluating arbitrary witness values that happen to be at the start of c2. completely redundant hint and constraint if added

provekit/prover/src/whir_r1cs.rs
```diff
@@ -527,7 +352,7 @@ pub fn pad_to_pow2_len_min2(v: &mut Vec<FieldElement>) {
 
     let target = match min.checked_next_power_of_two() {
         Some(p2) => p2,
-        None => min, // fallback: can't grow to power-of-two, keep `min`
```
ashpect, line 530: Restore comment

provekit/prover/src/whir_r1cs.rs
```diff
@@ -561,29 +384,16 @@ pub fn run_zk_sumcheck_prover(
...
+    let spartan_num_vars = whir_for_blinding_of_spartan_config.num_witness_variables();
+    let target_b = 1usize << spartan_num_vars;
 
-    //  Flatten and pad to exactly 1 << blinding_num_vars - 1
-    let mut flat = blinding_polynomial
-        .iter()
-        .flatten()
-        .cloned()
-        .collect::<Vec<_>>();
+    let mut flat: Vec<FieldElement> = blinding_polynomial.iter().flatten().cloned().collect();
```
ashpect, line 391: is writing type necessary ?

provekit/prover/src/whir_r1cs.rs
```diff
@@ -561,29 +384,16 @@ pub fn run_zk_sumcheck_prover(
 
     let blinding_polynomial = generate_blinding_spartan_univariate_polys(m_0);
 
-    // Spartan blinding: m = log2(domain), target_len = 2^(m-1).
-    let blinding_num_vars = whir_for_blinding_of_spartan_config.initial_num_variables();
-    let target_b = 1usize << (blinding_num_vars - 1);
+    let spartan_num_vars = whir_for_blinding_of_spartan_config.num_witness_variables();
+    let target_b = 1usize << spartan_num_vars;
```
ashpect: WHY remove 1 ?????
CHECKK ASHISH

provekit/r1cs-compiler/src/whir_r1cs.rs
```diff
@@ -1,13 +1,12 @@
...
+    provekit_common::{utils::next_power_of_two, WhirR1CSScheme, WhirZkConfig, R1CS},
     whir::parameters::{
         default_max_pow, FoldingFactor, MultivariateParameters, ProtocolParameters, SoundnessType,
     },
 };
 
-// Minimum log2 of the WHIR evaluation domain (lower bound for m).
-const MIN_WHIR_NUM_VARIABLES: usize = 12;
-// Minimum number of variables in the sumcheck’s multilinear polynomial (lower
+const MIN_WHIR_NUM_VARIABLES: usize = 14;
```
ashpect: what is the logical reasoning for increasing this from 12 to 14 ?

provekit/r1cs-compiler/src/whir_r1cs.rs
```diff
@@ -44,34 +43,39 @@ impl WhirR1CSSchemeBuilder for WhirR1CSScheme {
         let m_0 = m0_raw.max(MIN_SUMCHECK_NUM_VARIABLES);
 
         Self {
-            m: m_raw + 1,
```
ashpect, line 47: is the reason for reducing +1 is because the new domain in whir 2.0 equal to witness size ?

tooling/provekit-bench/benches/bench.rs
```diff
@@ -63,7 +63,7 @@ fn verify_poseidon_1000(bencher: Bencher) {
     let mut verifier: Verifier = read(&proof_verifier_path).unwrap();
     let proof_path = crate_dir.join("noir-proof.np");
     let proof: NoirProof = read(&proof_path).unwrap();
-    bencher.bench_local(|| black_box(&mut verifier).verify(black_box(&proof)));
+    bencher.bench_local(|| black_box(&mut verifier).verify(black_box(proof.clone())));
```
ashpect: Cloning would introduce overhead in benchmarking. It's useless.

provekit/prover/src/whir_r1cs.rs
```diff
@@ -205,57 +122,45 @@ impl WhirR1CSProver for WhirR1CSScheme {
...
                     &mut weights,
-                    &commitment.masked_polynomial,
-                    &commitment.random_polynomial,
+                    &commitment.polynomial,
                     public_weight,
                 )
             };
 
-            merlin.prover_hint_ark(&(public_f_sum, public_g_sum));
+            merlin.prover_hint_ark(&public_eval);
```
ashpect: We can skip the above if and also make hint as if as well

if !public_inputs.is_empty() {
    let public_eval = compute_public_weight_evaluation(...);
    merlin.prover_hint_ark(&public_eval);
}

and similarly in the verifier do so. This was the old case when we needed to send hint regardless because we predefined the io transcript and didn't want to use if there, since that's not the case anymore we can optimize it.

provekit/common/src/utils/sumcheck.rs
Bisht13: Created an issue: https://github.com/worldfnd/provekit/issues/298

provekit/common/src/file/bin.rs
Bisht13: We're using xz was pkp and zstd for pkv and np.

provekit/r1cs-compiler/src/whir_r1cs.rs
Bisht13: whir_zk handles blinding internally, its domain is 2^m and the witness fills it directly. The old API needed m+1 because blinding consumed one extra variable. So m_raw is now used as-is.

provekit/prover/src/whir_r1cs.rs
Bisht13: Changed, public hint is now conditional in both single-commit and dual-commit paths. Both prover and verifier skip the hint entirely when public_inputs.is_empty().

provekit/common/src/prefix_covector.rs
Bisht13: These are different operations. expand_powers<const D> produces D powers per element ([1, x, x², x³] for each x in the slice). make_public_weight builds [1, x, x², …, x^{n-1}] for a single x where n = public_inputs_len is runtime, can't be a const generic. Keeping as-is.

provekit/prover/src/whir_r1cs.rs
Bisht13: Yes, bare collect() without a type annotation doesn't compile here. Rust can't back-propagate from the downstream resize() call to infer the collection type.

provekit/r1cs-compiler/src/whir_r1cs.rs
Bisht13: whir_zk's compute_num_blinding_variables() asserts num_blinding_variables < num_witness_variables (strict). At our security params (128-bit, folding factor 4, ConjectureList, log_inv_rate=1), the blinding computation yields exactly 12 blinding variables. So the strict minimum is 13. Changed it.

PR #297, opened 2026-02-21
Optimizations: ROM Fusion, Spread Multiset Fusion & Constant Spread Path

rose2221:
### ROM Table Constraint Fusion (`rom.rs`)

Added `add_indexed_table_entry_quotient()` replacing the old
`add_indexed_lookup_factor()` + `add_product()` pattern.

**Before:**

* 3 constraints per table entry
  (denominator definition + denom × inverse = 1 + mult × inverse = quotient)

**After:**

* 2 constraints per table entry
  (denominator definition + fused denom × quotient = mult)

The inverse witness is still created for batch inversion but is no longer separately constrained.
Soundness holds since `denom × quotient = mult` uniquely implies `quotient = mult / denom` in the field.

---

###  Spread Table Multiset Fusion (`spread.rs`)

Replaced two `add_sum()` calls + grand sum equality with a single fused A-vector constraint (same pattern as `add_range_check_via_lookup()`).

**Before:**

* 2 sum witnesses
* 3 constraints (query sum + table sum + equality)

**After:**

* 0 extra witnesses
* 1 constraint
  `((Σ table_quotients − Σ query_inverses) × 1 = 0)`

**Savings:** 2 witnesses + 2 constraints per spread table instance.

Updated cost model:
`2 × 2^w + 4  →  2 × 2^w + 2`

---

###  `decompose_constant_to_spread_word` — Constant-Path Spread Decomposition

Previously, spread decomposition panicked for constant witnesses.
Now constants (e.g. SHA256 IV `H[0]..H[7]`) are handled via a dedicated compile-time path:

* Chunk values + spreads precomputed at compile time
* Each spread witness pinned via an R1CS equality constraint
* No spread table lookups required

**Saves ~3 spread table lookups per constant**
(no multiplicity / inverse / quotient witnesses)

#### Soundness

* Compile-time assert verifies decomposition round-trip
* R1CS constraint pins `packed_witness = constant_value` inside the function, preventing under-constrained usage

**Eliminate phantom witnesses in spread decomposition** : spread_decompose previously allocated a Sum witness to hold the intermediate spread sum, but this witness was never referenced by any R1CS constraint — it only existed so SpreadBitExtract could read it. Now SpreadBitExtract stores the Vec<SumTerm> directly and computes the sum inline during witness solving. 
**Deduplicate product witnesses** : add_product(a, b) now caches results keyed by (min(a,b), max(a,b)), returning the existing witness index when the same operand pair is requested again. Eliminates redundant Product witness builders and R1CS constraints for repeated multiplications.

### DIFF
--- a/provekit/r1cs-compiler/src/memory/rom.rs
+++ b/provekit/r1cs-compiler/src/memory/rom.rs
@@ -95,7 +95,7 @@ pub(crate) fn add_rom_checking(r1cs_compiler: &mut NoirToR1CSCompiler, block: &M
 /// Table-side LogUp factor: computes quotient = multiplicity / denominator.
 /// See `add_indexed_lookup_factor` for the query-side counterpart (returns
 /// inverse instead of quotient).
-/// Uses 2 R1CS constraints that
+/// Uses 2 R1CS constraints instead of the 3 that
 /// `add_indexed_lookup_factor` + `add_product` would need:
 ///   Constraint 1: rs × value = sz − index·index_witness − denominator
 ///   Constraint 2: denominator × quotient = multiplicity   (fused)


provekit/r1cs-compiler/src/spread.rs
Bisht13: This PR removes the WitnessBuilder::Sum phantom witness from spread_decompose, but this cost formula still includes 1 + to account for it.
Should be:
```rs
let sd = 4 * n_sd; // spread_decompose
```

Bisht13: Stale comment: says table cost (2×2^w + 4) but the code on line 695 is now 2 * (1usize << w) + 2.

provekit/r1cs-compiler/src/spread.rs
```diff
@@ -421,6 +418,109 @@ pub(crate) fn add_u32_addition_spread(
...
+            // prove the (input, spread) pair membership.
+            let spread_idx =
+                compiler.add_sum(vec![SumTerm(Some(FieldElement::from(spread_val)), w_one)]);
+
+            sub_spreads.push(spread_idx);
+        }
+
+        chunks.push(SpreadChunk {
+            total_bits: chunk_spec[ci],
+            sub_values: sub_spreads.clone(),
```
Bisht13, line 510: sub_values semantically holds chunk value indices, but this sets it to spread indices. Currently harmless (field is never read anywhere), but could cause issues for future code. Worth a comment on this line.

provekit/r1cs-compiler/src/spread.rs
```diff
@@ -421,6 +418,109 @@ pub(crate) fn add_u32_addition_spread(
...
+    }
+    assert_eq!(
+        recomposed, constant_value as u64,
+        "constant spread decomposition mismatch: {constant_value:#x} decomposed to {recomposed:#x}"
+    );
+
+    // Pin packed_witness = constant_value in R1CS. Makes this function
+    // self-contained: even if the caller forgets to constrain
+    // packed_witness, the prover cannot set it to an arbitrary value.
+    compiler.r1cs.add_constraint(
```
Bisht13: packed_witness is already pinned by add_sum at the call site. This is a redundant constraint. The "self-contained" rationale in the comment is reasonable, just flagging it.

provekit/r1cs-compiler/src/spread.rs
```diff
@@ -421,6 +418,109 @@ pub(crate) fn add_u32_addition_spread(
...
+/// Chunk values and their spreads are pre-computed; spread witnesses
+/// are pinned via direct equality constraints (`spread_w = spread(c)`).
+/// Sub-chunk value witnesses are assigned by the solver but NOT
+/// constrained (they are unused in downstream R1CS—only spread
+/// witnesses appear in spread_decompose constraints).
+///
+/// Soundness: each spread witness is individually pinned to a specific
+/// constant baked into the constraint system. The verifier checks the
+/// pinning constraints; the prover cannot deviate.
+pub(crate) fn decompose_constant_to_spread_word(
```
Bisht13, line 433: Please add unit tests for this.

provekit/r1cs-compiler/src/memory/rom.rs
Bisht13, line 163: Now that there are two similar functions, a cross-reference comment would help: this one is read-side (returns inverse), add_indexed_table_entry_quotient is table-side (returns quotient).

provekit/r1cs-compiler/src/spread.rs
rose2221: done, thanks

rose2221: fixed, the doc comment now says 2×2^w + 2 to match the code. The formula was also rewritten to take n_const_hash and separately account for constant-hash compressions (pinned spreads, no table lookups).

provekit/r1cs-compiler/src/spread.rs
rose2221: added the required comment, thanks

provekit/r1cs-compiler/src/memory/rom.rs
rose2221: added the necessary documentation for the same.

provekit/r1cs-compiler/src/spread.rs
rose2221: added two unit tests:

- `constant_decomp_pins_correct_spread_values` - verifies each sub-spread witness is pinned to compute_spread(byte_value) using BYTE_CHUNKS at w=8 (single sub-chunk per chunk).

- `constant_decomp_multi_sub_chunk` — verifies SIGMA0_CHUNKS = [2,11,9,10] at w=8, which exercises multi-sub-chunk decomposition (e.g. 11-bit -> [8,3]).

provekit/r1cs-compiler/src/spread.rs
rose2221: it's been removed, thanks.

PR #303, opened 2026-02-23
fix : verification failure on linux-aarch64 AND chore: disable gnark and benchamark ci 

ashpect:
end-to-end ci not just gnark verifier but also rust proving/verification
disabling gnark verification and benchmarking file till it's fixed, tracking in https://github.com/worldfnd/provekit/issues/302

fixes https://github.com/worldfnd/provekit/issues/304


.github/workflows/ci.yml
```diff
@@ -21,6 +21,8 @@ jobs:
   build_and_test:
     name: Build and test
     runs-on: ubuntu-24.04-arm
+    env:
+      RUSTFLAGS: "-C debug-assertions=yes"
```
Bisht13, line 29: `RUSTFLAGS` env is a mutually exclusive source, it silently drops the
`opt-level=2` workaround from `.cargo/config.toml` (Cargo checks RUSTFLAGS
env *before* target.<triple>.rustflags config, and uses only the first
match). Safe today since this job is debug-only, but if `--release` is ever
added here the #304 bug silently resurfaces.
Suggest:
```yaml
    env:
      # Overrides .cargo/config.toml [target.aarch64-unknown-linux-gnu].rustflags
      # entirely (Cargo treats RUSTFLAGS env and config rustflags as mutually
      # exclusive). Safe: this job does debug builds only (opt-level=0).
      # Do NOT add --release without addressing this interaction. See #304.
      RUSTFLAGS: "-C debug-assertions=yes"
```

.cargo/config.toml
```diff
@@ -8,3 +8,9 @@ rustflags = ["-C", "target-feature=+simd128,+relaxed-simd"]
 [target.wasm32-wasip1]
 runner = "wasmtime run --dir . "
 rustflags = ["-C", "target-feature=+simd128,+relaxed-simd"]
+
+# LLVM miscompilation workaround for Linux aarch64.
+# Bug requires: opt-level=3 + codegen-units=1 + 256-bit field arithmetic + generics
+# See: https://github.com/worldfnd/provekit/issues/304
+[target.aarch64-unknown-linux-gnu]
+rustflags = ["-C", "opt-level=2"]
```
Bisht13: nit: Missing trailing newline at EOF (the file had one before this PR).

.github/workflows/end-to-end.yml
```diff
@@ -90,40 +98,42 @@ jobs:
...
-          MONITOR_PID=$!
-
-          # Run the main process 
-          ./gnark-verifier --config "../noir-examples/noir-passport/merkle_age_check/params_for_recursive_verifier" --r1cs "../noir-examples/noir-passport/merkle_age_check/r1cs.json"
-
-          # Stop monitoring
-          kill $MONITOR_PID
\ No newline at end of file
+
+      # Disabled gnark, check https://github.com/worldfnd/provekit/issues/302
```
Bisht13, line 105: Consider deleting the gnark step rather than commenting out ~35 lines.
#302 already tracks re-enablement and git history preserves the original.

ashpect: I think commenting is better in dev friendly terms.

PR #307, opened 2026-02-25
Port to whir b3c86d7 with sound parameters

Bisht13:
Port to whir b3c86d7 (Johnson bound) with tuned WHIR parameters. Old `ConjectureList` soundness was disproven — new config uses `starting_log_inv_rate=2`, `iff=3`, `ff=3`, `pow_bits=10`, verified at **128-bit security**.

### Benchmark (`complete_age_check`, 712K constraints)

| Config | Prove | Proof size |
|---|---|---|
| Old whir (unsound) | 2.20s | 3.56 |
| New whir, defaults | 9.60s | 11.3 MB |
| **New whir, tuned** | **1.45s** | **3.26 MB** |

### Changes
- Adapt to new whir API (`Config<M: Embedding>`, restructured `ProtocolParameters`, owned `prove()` args, removed `deferred()`/`final_in_domain_samples()`)
- Tune WHIR parameters: rate 1→2, folding 4→3, pow_bits fixed at 10
- Add security regression test (`verify_security_level`)

### DIFF
--- a/noir-examples/noir-r1cs-test-programs/acir_assert_zero/Nargo.toml
+++ b/noir-examples/noir-r1cs-test-programs/acir_assert_zero/Nargo.toml
@@ -4,5 +4,3 @@ type = "bin"
 authors = [""]
 compiler_version = ">=0.22.0"
 
-[dependencies]
-poseidon2 = { tag = "main", git = "https://github.com/TaceoLabs/noir-poseidon", directory = "poseidon2"}


tooling/provekit-gnark/src/gnark_config.rs
```diff
@@ -83,7 +86,12 @@ impl WHIRConfigGnark {
             })
             .collect();
 
-        let final_queries = whir_params.final_in_domain_samples();
+        let final_queries = whir_params
```
ocdbytes, line 91: nitpick : we can add a small comment explaining that if no last round params are available thus no rounds meaning we use initial rounds in domain sample values.

provekit/r1cs-compiler/src/whir_r1cs.rs
```diff
@@ -59,22 +57,40 @@ impl WhirR1CSSchemeBuilder for WhirR1CSScheme {
...
-            num_polynomials,
-        )
+        WhirZkConfig::new(1 << nv, &whir_params, num_polynomials)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
```
ocdbytes, line 82:
```suggestion
#[test]
fn verify_security_level_min_variables() {
        let config =
            WhirR1CSScheme::new_whir_zk_config_for_size(MIN_WHIR_NUM_VARIABLES, 1);
        let sec_blinded = config
            .blinded_commitment
            .security_level(config.blinded_commitment.initial_committer.num_vectors, 1);
        let sec_blinding = config
            .blinding_commitment
            .security_level(config.blinding_commitment.initial_committer.num_vectors, 1);
        eprintln!(
            "nv={}: blinded={sec_blinded:.2} bits, blinding={sec_blinding:.2} bits",
           MIN_WHIR_NUM_VARIABLES
        );
        assert!(
            sec_blinded >= 128.0,
            "Blinded commitment security {sec_blinded:.2} < 128 bits at nv={}",
            MIN_WHIR_NUM_VARIABLES
        );
        assert!(
            sec_blinding >= 128.0,
            "Blinding commitment security {sec_blinding:.2} < 128 bits at nv={}",
            MIN_WHIR_NUM_VARIABLES
        );
}
```

Let's add a test for minimum number of variables too so that we can be extra safe here that if any other parameter is tweaked we can be sure that security params hit 128 bits of security accumulated.

provekit/r1cs-compiler/src/whir_r1cs.rs
```diff
@@ -59,22 +57,40 @@ impl WhirR1CSSchemeBuilder for WhirR1CSScheme {
...
-            security_level:        128,
-            pow_bits:              default_max_pow(nv, 1),
-            folding_factor:        FoldingFactor::Constant(4),
-            soundness_type:        SoundnessType::ConjectureList,
-            starting_log_inv_rate: 1,
-            batch_size:            1,
-            hash_id:               whir::hash::SHA2,
+            unique_decoding:        false,
+            security_level:         128,
+            pow_bits:               10,
```
ocdbytes, line 68: nitpick (personal) : I feel like hardcoding these params without explanation seems off to me but they were like this before also. If we can add some description here in provekit side. Not necessary in this PR.

PR #310, opened 2026-02-26
Feat : MSM black box implementation

ocdbytes:
# MSM Black Box Support

- Implements black-box Multi-Scalar Multiplication (MSM) in the R1CS compiler with full elliptic
  curve arithmetic over multi-limb field elements.
- Adds a cost model that analytically selects optimal (limb_bits, window_size) parameters,
  following the same pattern as SHA256 spread optimization.
- Implements FakeGLV scalar decomposition for single-point MSM, reducing scalar width by ~50% via
  half-GCD and interleaved two-point windowed scalar multiplication.
- Adds bigint_mod.rs in the prover for 256-bit modular arithmetic (half-GCD, Barrett reduction,
  modular inverse) used in witness solving.
- 22.8% decrease in constraint count as compared to native noir implementation of msm.
- To test MSM blackbox :
    ```sh
    cargo test -p provekit-bench --test compiler -- "embedded_curve_msm"
    ```

**Performance Comparison :**
This is for 2 points call for msm: s_1[p_1] * s_2[p_2]

| Metric | Native Noir MSM | Black Box MSM | % Reduction |
|---|---|---|---|
| R1CS Constraints | 5,741 | 4,434 | -22.8% |
| R1CS Witnesses | 6,721 | 6,154 | -8.4% |
| A Entries | 6,749 | 7,825 | +15.9% |
| B Entries | 6,364 | 7,272 | +14.3% |
| C Entries | 14,453 | 8,539 | -40.9% |
| W1 Size | 3,354 | 2,276 | -32.1% |
| W2 Size | 3,367 | 3,878 | +15.2% |

### DIFF
--- a/provekit/prover/src/bigint_mod.rs
+++ b/provekit/prover/src/bigint_mod.rs
@@ -533,7 +533,8 @@ fn mul_mod_no_reduce(a: &[u64; 4], b: &[u64; 4]) -> [u64; 4] {
 
 /// Convert a `[u64; 4]` bigint to a `FieldElement`.
 pub fn bigint_to_fe(val: &[u64; 4]) -> FieldElement {
-    FieldElement::from_bigint(ark_ff::BigInt(*val)).unwrap()
+    FieldElement::from_bigint(ark_ff::BigInt(*val))
+        .expect("bigint value exceeds BN254 field modulus")
 }
 
 /// Read a `FieldElement` witness as a `[u64; 4]` bigint.
--- a/provekit/prover/src/witness/witness_builder.rs
+++ b/provekit/prover/src/witness/witness_builder.rs
@@ -39,11 +39,16 @@ use super::limb_io::{
     ColumnEqParams, ColumnEqTerms,
 };
 
+/// Read a solved witness value, panicking with the index if not yet solved.
+fn get_witness(witness: &[Option<FieldElement>], idx: usize) -> FieldElement {
+    witness[idx].unwrap_or_else(|| panic!("witness at index {idx} not yet solved"))
+}
+
 /// Resolve a ConstantOrR1CSWitness to its FieldElement value.
 fn resolve(witness: &[Option<FieldElement>], v: &ConstantOrR1CSWitness) -> FieldElement {
     match v {
         ConstantOrR1CSWitness::Constant(c) => *c,
-        ConstantOrR1CSWitness::Witness(idx) => witness[*idx].unwrap(),
+        ConstantOrR1CSWitness::Witness(idx) => get_witness(witness, *idx),
     }
 }
 
@@ -62,7 +67,7 @@ impl WitnessBuilderSolver for WitnessBuilder {
                 witness[*witness_idx] = Some(noir_to_native(
                     *acir_witness_idx_to_value_map
                         .get_index(*acir_witness_idx as u32)
-                        .unwrap(),
+                        .expect("ACIR witness index not found in witness map"),
                 ));
             }
             WitnessBuilder::Sum(witness_idx, operands) => {
@@ -71,17 +76,17 @@ impl WitnessBuilderSolver for WitnessBuilder {
                         .iter()
                         .map(|SumTerm(coeff, witness_idx)| {
                             if let Some(coeff) = coeff {
-                                *coeff * witness[*witness_idx].unwrap()
+                                *coeff * get_witness(witness, *witness_idx)
                             } else {
-                                witness[*witness_idx].unwrap()
+                                get_witness(witness, *witness_idx)
                             }
                         })
                         .fold(FieldElement::zero(), |acc, x| acc + x),
                 );
             }
             WitnessBuilder::Product(witness_idx, operand_idx_a, operand_idx_b) => {
-                let a: FieldElement = witness[*operand_idx_a].unwrap();
-                let b: FieldElement = witness[*operand_idx_b].unwrap();
+                let a: FieldElement = get_witness(witness, *operand_idx_a);
+                let b: FieldElement = get_witness(witness, *operand_idx_b);
                 witness[*witness_idx] = Some(a * b);
             }
             WitnessBuilder::Inverse(..) | WitnessBuilder::LogUpInverse(..) => {
@@ -90,21 +95,22 @@ impl WitnessBuilderSolver for WitnessBuilder {
                 )
             }
             WitnessBuilder::SafeInverse(witness_idx, operand_idx) => {
-                let val = witness[*operand_idx].unwrap();
+                let val = get_witness(witness, *operand_idx);
                 witness[*witness_idx] = Some(if val == FieldElement::zero() {
                     FieldElement::zero()
                 } else {
-                    val.inverse().unwrap()
+                    val.inverse()
+                        .expect("non-zero field element must have an inverse")
                 });
             }
             WitnessBuilder::ModularInverse(witness_idx, operand_idx, modulus) => {
-                let a_limbs = fe_to_bigint(witness[*operand_idx].unwrap());
+                let a_limbs = fe_to_bigint(get_witness(witness, *operand_idx));
                 let m_limbs = modulus.into_bigint().0;
                 let exp = sub_u64(&m_limbs, 2);
                 witness[*witness_idx] = Some(bigint_to_fe(&mod_pow(&a_limbs, &exp, &m_limbs)));
             }
             WitnessBuilder::IntegerQuotient(witness_idx, dividend_idx, divisor) => {
-                let d_limbs = fe_to_bigint(witness[*dividend_idx].unwrap());
+                let d_limbs = fe_to_bigint(get_witness(witness, *dividend_idx));
                 let m_limbs = divisor.into_bigint().0;
                 let (quotient, _) = divmod(&d_limbs, &m_limbs);
                 witness[*witness_idx] = Some(bigint_to_fe(&quotient));
@@ -117,7 +123,7 @@ impl WitnessBuilderSolver for WitnessBuilder {
                 let sum: FieldElement = terms
                     .iter()
                     .map(|SumTerm(coeff, idx)| {
-                        let val = witness[*idx].unwrap();
+                        let val = get_witness(witness, *idx);
                         coeff.map_or(val, |c| c * val)
                     })
                     .fold(FieldElement::zero(), |acc, x| acc + x);
@@ -133,10 +139,10 @@ impl WitnessBuilderSolver for WitnessBuilder {
                 rs_challenge,
                 value,
             ) => {
-                let index = witness[*index].unwrap();
-                let value = witness[*value].unwrap();
-                let rs_challenge = witness[*rs_challenge].unwrap();
-                let sz_challenge = witness[*sz_challenge].unwrap();
+                let index = get_witness(witness, *index);
+                let value = get_witness(witness, *value);
+                let rs_challenge = get_witness(witness, *rs_challenge);
+                let sz_challenge = get_witness(witness, *sz_challenge);
                 witness[*witness_idx] =
                     Some(sz_challenge - (*index_coeff * index + rs_challenge * value));
             }
@@ -145,7 +151,7 @@ impl WitnessBuilderSolver for WitnessBuilder {
                 for value_witness_idx in value_witnesses {
                     // If the value is representable as just a u64, then it should be the least
                     // significant value in the BigInt representation.
-                    let value = witness[*value_witness_idx].unwrap().into_bigint().0[0];
+                    let value = get_witness(witness, *value_witness_idx).into_bigint().0[0];
                     multiplicities[value as usize] += 1;
                 }
                 for (i, count) in multiplicities.iter().enumerate() {
@@ -162,7 +168,8 @@ impl WitnessBuilderSolver for WitnessBuilder {
                 WitnessCoefficient(value_coeff, value),
             ) => {
                 witness[*witness_idx] = Some(
-                    witness[*sz_challenge].unwrap() - (*value_coeff * witness[*value].unwrap()),
+                    get_witness(witness, *sz_challenge)
+                        - (*value_coeff * get_witness(witness, *value)),
                 );
             }
             WitnessBuilder::ProductLinearOperation(
@@ -170,8 +177,9 @@ impl WitnessBuilderSolver for WitnessBuilder {
                 ProductLinearTerm(x, a, b),
                 ProductLinearTerm(y, c, d),
             ) => {
-                witness[*witness_idx] =
-                    Some((*a * witness[*x].unwrap() + *b) * (*c * witness[*y].unwrap() + *d));
+                witness[*witness_idx] = Some(
+                    (*a * get_witness(witness, *x) + *b) * (*c * get_witness(witness, *y) + *d),
+                );
             }
             WitnessBuilder::DigitalDecomposition(dd_struct) => {
                 dd_struct.solve(witness);
@@ -185,13 +193,13 @@ impl WitnessBuilderSolver for WitnessBuilder {
                 WitnessCoefficient(timer, timer_witness),
             ) => {
                 witness[*witness_idx] = Some(
-                    witness[*sz_challenge].unwrap()
-                        - (*addr * witness[*addr_witness].unwrap()
-                            + witness[*rs_challenge].unwrap() * witness[*value].unwrap()
-                            + witness[*rs_challenge].unwrap()
-                                * witness[*rs_challenge].unwrap()
+                    get_witness(witness, *sz_challenge)
+                        - (*addr * get_witness(witness, *addr_witness)
+                            + get_witness(witness, *rs_challenge) * get_witness(witness, *value)
+                            + get_witness(witness, *rs_challenge)
+                                * get_witness(witness, *rs_challenge)
                                 * *timer
-                                * witness[*timer_witness].unwrap()),
+                                * get_witness(witness, *timer_witness)),
                 );
             }
             WitnessBuilder::SpiceWitnesses(spice_witnesses) => {
@@ -210,10 +218,10 @@ impl WitnessBuilderSolver for WitnessBuilder {
                 let rhs = resolve(witness, rhs);
                 let output = resolve(witness, output);
                 witness[*witness_idx] = Some(
-                    witness[*sz_challenge].unwrap()
+                    get_witness(witness, *sz_challenge)
                         - (lhs
-                            + witness[*rs_challenge].unwrap() * rhs
-                            + witness[*rs_challenge_sqrd].unwrap() * output),
+                            + get_witness(witness, *rs_challenge) * rhs
+                            + get_witness(witness, *rs_challenge_sqrd) * output),
                 );
             }
             WitnessBuilder::CombinedBinOpLookupDenominator(
@@ -233,11 +241,11 @@ impl WitnessBuilderSolver for WitnessBuilder {
                 let xor_out = resolve(witness, xor_output);
                 // Encoding: sz - (lhs + rs*rhs + rs²*and_out + rs³*xor_out)
                 witness[*witness_idx] = Some(
-                    witness[*sz_challenge].unwrap()
+                    get_witness(witness, *sz_challenge)
                         - (lhs
-                            + witness[*rs_challenge].unwrap() * rhs
-                            + witness[*rs_sqrd].unwrap() * and_out
-                            + witness[*rs_cubed].unwrap() * xor_out),
+                            + get_witness(witness, *rs_challenge) * rhs
+                            + get_witness(witness, *rs_sqrd) * and_out
+                            + get_witness(witness, *rs_cubed) * xor_out),
                 );
             }
             WitnessBuilder::MultiplicitiesForBinOp(witness_idx, atomic_bits, operands) => {
@@ -426,7 +434,7 @@ impl WitnessBuilderSolver for WitnessBuilder {
                 witness[*output] = Some(FieldElement::from(q));
             }
             WitnessBuilder::BytePartition { lo, hi, x, k } => {
-                let x_val = witness[*x].unwrap().into_bigint().0[0];
+                let x_val = get_witness(witness, *x).into_bigint().0[0];
                 assert!(
                     x_val < 256,
                     "BytePartition input must be 8-bit, got {x_val}"
@@ -446,8 +454,8 @@ impl WitnessBuilderSolver for WitnessBuilder {
                 curve_order,
             } => {
                 let s_val = reconstruct_from_halves(
-                    &fe_to_bigint(witness[*s_lo].unwrap()),
-                    &fe_to_bigint(witness[*s_hi].unwrap()),
+                    &fe_to_bigint(get_witness(witness, *s_lo)),
+                    &fe_to_bigint(get_witness(witness, *s_hi)),
                 );
 
                 let (val1, val2, neg1, neg2) = half_gcd(&s_val, curve_order);
@@ -464,8 +472,8 @@ impl WitnessBuilderSolver for WitnessBuilder {
                 curve_a,
                 field_modulus_p,
             } => {
-                let px_val = fe_to_bigint(witness[*px].unwrap());
-                let py_val = fe_to_bigint(witness[*py].unwrap());
+                let px_val = fe_to_bigint(get_witness(witness, *px));
+                let py_val = fe_to_bigint(get_witness(witness, *py));
 
                 let (lambda, x3, y3) =
                     ec_point_double_with_lambda(&px_val, &py_val, curve_a, field_modulus_p);
@@ -482,10 +490,10 @@ impl WitnessBuilderSolver for WitnessBuilder {
                 y2,
                 field_modulus_p,
             } => {
-                let x1_val = fe_to_bigint(witness[*x1].unwrap());
-                let y1_val = fe_to_bigint(witness[*y1].unwrap());
-                let x2_val = fe_to_bigint(witness[*x2].unwrap());
-                let y2_val = fe_to_bigint(witness[*y2].unwrap());
+                let x1_val = fe_to_bigint(get_witness(witness, *x1));
+                let y1_val = fe_to_bigint(get_witness(witness, *y1));
+                let x2_val = fe_to_bigint(get_witness(witness, *x2));
+                let y2_val = fe_to_bigint(get_witness(witness, *y2));
 
                 let (lambda, x3, y3) =
                     ec_point_add_with_lambda(&x1_val, &y1_val, &x2_val, &y2_val, field_modulus_p);
@@ -772,17 +780,17 @@ impl WitnessBuilderSolver for WitnessBuilder {
             } => {
                 let n = *num_limbs as usize;
                 let scalar = reconstruct_from_halves(
-                    &fe_to_bigint(witness[*s_lo].unwrap()),
-                    &fe_to_bigint(witness[*s_hi].unwrap()),
+                    &fe_to_bigint(get_witness(witness, *s_lo)),
+                    &fe_to_bigint(get_witness(witness, *s_hi)),
                 );
 
                 let px_val = if n == 1 {
-                    fe_to_bigint(witness[px_limbs[0]].unwrap())
+                    fe_to_bigint(get_witness(witness, px_limbs[0]))
                 } else {
                     read_witness_limbs(witness, px_limbs, *limb_bits)
                 };
                 let py_val = if n == 1 {
-                    fe_to_bigint(witness[py_limbs[0]].unwrap())
+                    fe_to_bigint(get_witness(witness, py_limbs[0]))
                 } else {
                     read_witness_limbs(witness, py_limbs, *limb_bits)
                 };
@@ -805,14 +813,14 @@ impl WitnessBuilderSolver for WitnessBuilder {
                 on_false,
                 on_true,
             } => {
-                let f = witness[*flag].unwrap();
-                let a = witness[*on_false].unwrap();
-                let b = witness[*on_true].unwrap();
+                let f = get_witness(witness, *flag);
+                let a = get_witness(witness, *on_false);
+                let b = get_witness(witness, *on_true);
                 witness[*output] = Some(a + f * (b - a));
             }
             WitnessBuilder::BooleanOr { output, a, b } => {
-                let a_val = witness[*a].unwrap();
-                let b_val = witness[*b].unwrap();
+                let a_val = get_witness(witness, *a);
+                let b_val = get_witness(witness, *b);
                 witness[*output] = Some(a_val + b_val - a_val * b_val);
             }
             WitnessBuilder::SignedBitHint {
@@ -825,7 +833,7 @@ impl WitnessBuilderSolver for WitnessBuilder {
                     "SignedBitHint: num_bits={} exceeds 128; scalar would be silently truncated",
                     num_bits
                 );
-                let s_fe = witness[*scalar].unwrap();
+                let s_fe = get_witness(witness, *scalar);
                 let s_big = s_fe.into_bigint().0;
                 let s_val: u128 = s_big[0] as u128 | ((s_big[1] as u128) << 64);
                 let n = *num_bits;
@@ -855,7 +863,7 @@ impl WitnessBuilderSolver for WitnessBuilder {
                 packed,
                 chunk_bits,
             } => {
-                let packed_val = witness[*packed].unwrap().into_bigint().0[0];
+                let packed_val = get_witness(witness, *packed).into_bigint().0[0];
                 let mut offset = 0u32;
                 for (i, &bits) in chunk_bits.iter().enumerate() {
                     let mask = (1u64 << bits) - 1;
@@ -865,7 +873,7 @@ impl WitnessBuilderSolver for WitnessBuilder {
                 }
             }
             WitnessBuilder::SpreadWitness(output_idx, input_idx) => {
-                let input_val = witness[*input_idx].unwrap().into_bigint().0[0];
+                let input_val = get_witness(witness, *input_idx).into_bigint().0[0];
                 let spread = compute_spread(input_val);
                 witness[*output_idx] = Some(FieldElement::from(spread));
             }
@@ -879,7 +887,7 @@ impl WitnessBuilderSolver for WitnessBuilder {
                 let sum_fe: FieldElement = sum_terms
                     .iter()
                     .map(|SumTerm(coeff, idx)| {
-                        let v = witness[*idx].unwrap();
+                        let v = get_witness(witness, *idx);
                         if let Some(c) = coeff {
                             *c * v
                         } else {
@@ -916,8 +924,8 @@ impl WitnessBuilderSolver for WitnessBuilder {
                 }
             }
             WitnessBuilder::SpreadLookupDenominator(idx, sz, rs, input, spread_output) => {
-                let sz_val = witness[*sz].unwrap();
-                let rs_val = witness[*rs].unwrap();
+                let sz_val = get_witness(witness, *sz);
+                let rs_val = get_witness(witness, *rs);
                 let input_val = resolve(witness, input);
                 let spread_val = resolve(witness, spread_output);
                 // sz - (input + rs * spread_output)
--- a/provekit/r1cs-compiler/src/msm/mod.rs
+++ b/provekit/r1cs-compiler/src/msm/mod.rs
@@ -9,10 +9,7 @@ mod scalar_relation;
 
 pub use provekit_common::witness::{Limbs, MAX_LIMBS};
 use {
-    crate::{
-        constraint_helpers::{add_constant_witness, constrain_boolean},
-        noir_to_r1cs::NoirToR1CSCompiler,
-    },
+    crate::{constraint_helpers::constrain_boolean, noir_to_r1cs::NoirToR1CSCompiler},
     ark_ff::{AdditiveGroup, Field, PrimeField},
     curve::Curve,
     ec_points::{NativeEcOps, NonNativeEcOps},
--- a/provekit/prover/src/whir_r1cs.rs
+++ b/provekit/prover/src/whir_r1cs.rs
@@ -260,7 +260,10 @@ fn prove_from_alphas(
 
     if is_single {
         // Single commitment path
-        let commitment = commitments.into_iter().next().unwrap();
+        let commitment = commitments
+            .into_iter()
+            .next()
+            .expect("single-commitment path requires at least one commitment");
         let (mut weights, evals) =
             create_weights_and_evaluations::<3>(scheme.m, &commitment.polynomial, alphas);
 
@@ -296,8 +299,12 @@ fn prove_from_alphas(
     } else {
         // Dual commitment path
         let mut commitments = commitments.into_iter();
-        let c1 = commitments.next().unwrap();
-        let c2 = commitments.next().unwrap();
+        let c1 = commitments
+            .next()
+            .expect("dual-commitment path requires first commitment");
+        let c2 = commitments
+            .next()
+            .expect("dual-commitment path requires second commitment");
 
         let (alphas_1, alphas_2): (Vec<_>, Vec<_>) = alphas
             .into_iter()
@@ -307,8 +314,12 @@ fn prove_from_alphas(
             })
             .unzip();
 
-        let alphas_1: [Vec<FieldElement>; 3] = alphas_1.try_into().unwrap();
-        let alphas_2: [Vec<FieldElement>; 3] = alphas_2.try_into().unwrap();
+        let alphas_1: [Vec<FieldElement>; 3] = alphas_1
+            .try_into()
+            .expect("alphas_1 must have exactly 3 elements");
+        let alphas_2: [Vec<FieldElement>; 3] = alphas_2
+            .try_into()
+            .expect("alphas_2 must have exactly 3 elements");
 
         let evals_1 = compute_alpha_evals(&c1.polynomial, &alphas_1);
         let evals_2 = compute_alpha_evals(&c2.polynomial, &alphas_2);
--- a/provekit/prover/src/witness/limb_io.rs
+++ b/provekit/prover/src/witness/limb_io.rs
@@ -14,7 +14,8 @@ use {
 
 /// Convert a u128 value to a FieldElement.
 pub(super) fn u128_to_fe(val: u128) -> FieldElement {
-    FieldElement::from_bigint(BigInt([val as u64, (val >> 64) as u64, 0, 0])).unwrap()
+    FieldElement::from_bigint(BigInt([val as u64, (val >> 64) as u64, 0, 0]))
+        .expect("u128 value exceeds BN254 field modulus")
 }
 
 /// Read witness limbs and reconstruct as [u64; 4].
@@ -31,7 +32,10 @@ pub(super) fn read_witness_limbs(
                 "read_witness_limbs: index {idx} out of bounds (witness len {})",
                 witness.len()
             );
-            let bigint = witness[idx].unwrap().into_bigint().0;
+            let bigint = witness[idx]
+                .unwrap_or_else(|| panic!("witness limb at index {idx} not yet solved"))
+                .into_bigint()
+                .0;
             bigint[0] as u128 | ((bigint[1] as u128) << 64)
         })
         .collect();


provekit/r1cs-compiler/src/msm/multi_limb_ops.rs
```diff
@@ -0,0 +1,397 @@
...
+            let r = if a[0] == b[0] {
+                self.compiler
+                    .add_sum(vec![SumTerm(Some(FieldElement::from(2u64)), a[0])])
+            } else {
+                self.compiler
+                    .add_sum(vec![SumTerm(None, a[0]), SumTerm(None, b[0])])
+            };
+            Limbs::single(r)
+        } else if self.is_non_native_single() {
+            let modulus = self.params.modulus_fe.unwrap();
```
Bisht13: modulus_fe is Option<FieldElement> that is Some only when !is_native, but nothing in the type system enforces this. If MultiLimbParams construction logic ever changes, this becomes a panic with no context.

The root issue is that MultiLimbParams uses a single struct + runtime booleans for three distinct configurations (native-single, non-native-single, multi-limb). If field arithmetic dispatch used a trait (matching the EcOps pattern), the non-native-single impl would hold modulus: FieldElement directly, no Option, no unwrap().

At minimum: change to .expect("non-native single-limb path requires modulus_fe") at all 3 sites.

provekit/r1cs-compiler/src/msm/ec_points/hints_native.rs
```diff
@@ -0,0 +1,135 @@
...
+    let lambda = hint_start;
+    let x3 = hint_start + 1;
+    let y3 = hint_start + 2;
+
+    // x_sq = px * px (1W + 1C)
+    let x_sq = compiler.add_product(px, px);
+
+    // Constraint: lambda * (2 * py) = 3 * x_sq + a
+    // A = [lambda], B = [2*py], C = [3*x_sq + a_const]
+    let a_fe = FieldElement::from_bigint(ark_ff::BigInt(params.curve_a_raw)).unwrap();
```
Bisht13: FieldElement::from_bigint(ark_ff::BigInt(params.curve_a_raw)).unwrap() assumes curve params fit in the BN254 scalar field. True for native curves by definition, but NativeEcOps receives a MultiLimbParams with no type-level guarantee that it's native. A misconfiguration (e.g., passing secp256r1 params where a = p - 3 exceeds BN254 Fr) would panic here.

Add .expect("curve param must fit in native field") or validate in a constructor.

provekit/prover/src/bigint_mod.rs
```diff
@@ -0,0 +1,1195 @@
...
+    let neg1 = true; // always negate r_i: -r_i + t_i * s ≡ 0 (mod n)
+    let neg2 = t_curr_neg;
+
+    (val1, val2, neg1, neg2)
+}
+
+/// Multiply two 4-limb values without modular reduction.
+/// Returns the lower 4 limbs (ignoring overflow beyond 256 bits).
+/// Used internally by half_gcd for q * t_curr where the result is known to fit.
+fn mul_mod_no_reduce(a: &[u64; 4], b: &[u64; 4]) -> [u64; 4] {
```
Bisht13, line 521: Returns only the lower 4 limbs of widening_mul with no overflow check. The comment says "result is known to fit" but there's no debug_assert! verifying the upper limbs are zero. In a prover, silent truncation can produce incorrect witnesses that pass constraint checks with wrong values.

Add: debug_assert!(wide[4..].iter().all(|&x| x == 0), "mul_mod_no_reduce overflow");

provekit/r1cs-compiler/src/msm/mod.rs
```diff
@@ -0,0 +1,205 @@
...
+
+    let (limb_bits, window_size, num_limbs) = cost_model::get_optimal_msm_params(
+        native_bits,
+        curve_bits,
+        n_points,
+        scalar_bits,
+        is_native,
+    );
+
+    assert_eq!(
```
Bisht13, line 86: The existing provekit codebase uses anyhow::Result with .context() for recoverable errors (see witness_generator.rs). This module uses assert!/panic! for input validation that could fail with bad parameters — here, in add_sub_mod_p_core, mul_mod_p_multi, and scalar_relation_limb_bits (cost_model.rs:389). These produce cryptic panics instead of actionable error messages.

Consider returning Result from public entry points, reserving assert!/debug_assert! for internal invariants that should never fire.

provekit/r1cs-compiler/src/msm/multi_limb_ops.rs
```diff
@@ -0,0 +1,397 @@
...
+            let zero = self.constant_limbs(&zero_vals);
+            self.sub(zero, value)
+        }
+    }
+
+    #[must_use]
+    pub fn add(&mut self, a: Limbs, b: Limbs) -> Limbs {
+        assert_eq!(a.len(), self.n(), "add: a.len() != num_limbs");
+        assert_eq!(b.len(), self.n(), "add: b.len() != num_limbs");
+        if self.is_native_single() {
```
Bisht13: add(), sub(), mul(), and negate() each have a 3-way if/else if/else chain checking is_native_single(), is_non_native_single(), and the multi-limb fallback. That's 12 runtime dispatches across 4 methods.

The EcOps trait already does compile-time dispatch for EC operations (native vs non-native). Field arithmetic should follow the same pattern:

```rs
trait FieldArith {
    fn add(&self, ctx: &mut CompilerCtx, a: Limbs, b: Limbs) -> Limbs;
    fn sub(&self, ctx: &mut CompilerCtx, a: Limbs, b: Limbs) -> Limbs;
    fn mul(&self, ctx: &mut CompilerCtx, a: Limbs, b: Limbs) -> Limbs;
    fn negate(&self, ctx: &mut CompilerCtx, value: Limbs) -> Limbs;
}

struct NativeSingleField;
struct NonNativeSingleField { modulus: FieldElement } // no Option
struct MultiLimbField { p_limbs: Vec<FieldElement>, ... }
```

 This eliminates the Option<FieldElement>, the is_native boolean, all .unwrap() calls, and the dummy fields in for_curve_order().

provekit/r1cs-compiler/src/msm/multi_limb_ops.rs
```diff
@@ -0,0 +1,397 @@
...
+            curve_a_raw: curve.curve_a(),
+            curve_b_limbs: curve.curve_b_limbs(limb_bits, num_limbs),
+            curve_b_raw: curve.curve_b(),
+            is_native,
+            modulus_fe,
+        }
+    }
+
+    /// Build params for scalar relation verification (mod curve_order_n).
+    pub fn for_curve_order<C: super::curve::Curve>(
```
Bisht13, line 222: for_curve_order() fills curve_a_limbs, curve_a_raw, curve_b_limbs, curve_b_raw with zeros and marks them // unused. This means the struct has fields that are never read in one of its two use cases.

In Rust, this should be split into ModulusParams (shared: p_limbs, p_minus_1_limbs, modulus_raw, limb_bits, num_limbs) and CurveParams (EC-specific: curve_a, curve_b). The scalar relation path would only hold ModulusParams. No dummy fields, no "unused" comments, no invalid state representable.

provekit/r1cs-compiler/src/msm/ec_points/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+    crate::noir_to_r1cs::NoirToR1CSCompiler,
+    std::collections::BTreeMap,
+};
+
+// ---------------------------------------------------------------------------
+// EcOps trait — strategy interface for EC point arithmetic
+// ---------------------------------------------------------------------------
+
+/// Strategy for constraining elliptic curve operations in the circuit.
+pub trait EcOps {
```
Bisht13, line 26: Every method takes compiler, range_checks, and params as explicit arguments. NativeEcOps ignores _range_checks. The trait is a type-level tag rather than a proper interface.

If methods took &mut self, each impl would carry its own state, NativeEcOps wouldn't need a range_checks parameter it never uses, and the API would be honest about what each impl actually needs:

```rs
trait EcOps {
    fn point_double(&mut self, p: EcPoint) -> EcPoint;
    fn point_add(&mut self, p1: EcPoint, p2: EcPoint) -> EcPoint;
    fn verify_on_curve(&mut self, p: EcPoint);
}
```

provekit/r1cs-compiler/src/msm/multi_limb_ops.rs
```diff
@@ -0,0 +1,397 @@
...
+    compiler.r1cs.add_constraint(
+        &[(FieldElement::ONE, compiler.witness_one())],
+        &[(FieldElement::ONE, w)],
+        &[(value, compiler.witness_one())],
+    );
+    w
+}
+
+/// Unified field + EC operations struct.
+pub struct MultiLimbOps<'a, 'p, E> {
```
Bisht13: This single struct is simultaneously: a field arithmetic dispatcher (3-way runtime branching), an EC operations proxy, a compiler state holder, a constant allocator, a boolean constraint helper, and a range check registrar. The PhantomData<E> is the tell, the EC type parameter is only used in a separate impl<E: EcOps> block.

Consider decomposing into CompilerCtx (compiler + range_checks) with field ops and EC ops as separate traits on that context, rather than one struct doing everything.

provekit/r1cs-compiler/src/msm/multi_limb_arith.rs
```diff
@@ -0,0 +1,626 @@
...
+
+    let modulus_bits = modulus.into_bigint().num_bits();
+    range_checks.entry(modulus_bits).or_default().push(result);
+
+    result
+}
+
+/// a + b mod p (N=1 path)
+#[must_use]
+pub fn add_mod_p_single(
```
Bisht13: Throughout the module, witness indices, limb counts, bit widths, array lengths, and loop counters are all usize. A WitnessIdx(usize) newtype would prevent transposition bugs (passing a limb count where a witness index is expected) and make APIs self-documenting. This is standard practice in circuit compiler codebases.

provekit/prover/src/witness/witness_builder.rs
```diff
@@ -319,9 +390,107 @@ impl WitnessBuilderSolver for WitnessBuilder {
                     lh_val.into_bigint() ^ rh_val.into_bigint(),
                 ));
             }
+            WitnessBuilder::MultiLimbMulModHint {
```
Bisht13, line 337: This PR adds 10+ new variants (NonNativeEcHint, FakeGLVHint, EcScalarMulHint, SelectWitness, BooleanOr, SignedBitHint, MultiLimbAddQuotient, MultiLimbSubBorrow, MultiLimbMulModHint, MultiLimbModularInverse, EcDoubleHint, EcAddHint). Each requires matching changes in 4 files: the enum definition, the solver, dependency.rs extract_reads/extract_writes, and remapper.rs remap_builder.

A WitnessOp trait with solve(), reads(), writes(), remap() methods would let MSM define its own ops without touching shared infrastructure. I understand this is a pre-existing pattern, but this PR significantly exacerbates the maintenance burden.

provekit/r1cs-compiler/src/msm/curve/mod.rs
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    /// Convert modulus to a native field element.
+    fn p_native_fe(&self) -> FieldElement {
+        curve_native_point_fe(&self.field_modulus_p())
+    }
+
+    // ===== Provided: limb decomposition helpers =====
+
+    fn p_limbs(&self, limb_bits: u32, num_limbs: usize) -> Vec<FieldElement> {
```
Bisht13, line 71: p_limbs, p_minus_1_limbs, curve_a_limbs, curve_b_limbs, curve_order_n_limbs, curve_order_n_minus_1_limbs, generator_x_limbs, offset_x_limbs, offset_y_limbs — none are overridden by any curve. They're utility functions shoved into a trait.

These should be a precomputed CurveDecomposition struct built once from the Curve trait, rather than recomputing on every call:

```rs
struct CurveDecomposition {
    p_limbs: Vec<FieldElement>,
    curve_a_limbs: Vec<FieldElement>,
    // ...
}
impl CurveDecomposition {
    fn new(curve: &impl Curve, limb_bits: u32, num_limbs: usize) -> Self { ... }
}
```

provekit/r1cs-compiler/src/msm/multi_limb_arith.rs
```diff
@@ -0,0 +1,626 @@
...
+
+    is_zero
+}
+
+// ---------------------------------------------------------------------------
+// N≥2 multi-limb path (generalization of wide_ops.rs)
+// ---------------------------------------------------------------------------
+
+/// Shared core for `add_mod_p_multi` and `sub_mod_p_multi`.
+fn add_sub_mod_p_core(
```
Bisht13, line 71: A boolean that switches function behavior should be an enum or two separate functions:

```rs
enum ModularOp { Add, Sub }
```

Or factor the common carry-chain logic into a shared helper and let add_mod_p_multi / sub_mod_p_multi each provide their per-limb terms.

provekit/r1cs-compiler/src/msm/limbs.rs
```diff
@@ -0,0 +1,111 @@
...
+/// A fixed-capacity `Copy` array of witness indices, indexed by limb position.
+#[derive(Clone, Copy)]
+pub struct Limbs {
+    data: [usize; MAX_LIMBS],
+    len:  usize,
+}
+
+impl Limbs {
+    /// Sentinel value for uninitialized limb slots.
+    const UNINIT: usize = usize::MAX;
```
Bisht13: An uninitialized limb slot silently contains usize::MAX, which only causes an out-of-bounds panic when the witness array is accessed, potentially far from where the logic error occurred. Consider using arrayvec::ArrayVec<usize, 32> (already an indirect dependency) which provides push/pop semantics and immediate detection of uninitialized access.

provekit/r1cs-compiler/src/msm/multi_limb_arith.rs
```diff
@@ -0,0 +1,626 @@
...
+}
+
+// ---------------------------------------------------------------------------
+// Schoolbook column equations (shared by mul_mod_p_multi and non-native EC
+// hints)
+// ---------------------------------------------------------------------------
+
+/// Merge terms with the same witness index by summing their coefficients.
+fn merge_terms(terms: &[(FieldElement, usize)]) -> Vec<(FieldElement, usize)> {
+    let mut map: HashMap<usize, FieldElement> = HashMap::new();
```
Bisht13: The rest of the codebase uses BTreeMap for deterministic iteration. For the small term counts here (< 20 entries), a BTreeMap would be cleaner and avoid the HashMap import. You'd also get deterministic ordering without the explicit sort.

provekit/r1cs-compiler/src/msm/ec_points/hints_non_native.rs
```diff
@@ -0,0 +1,494 @@
...
+    ark_ff::{Field, PrimeField},
+    provekit_common::{
+        witness::{NonNativeEcOp, WitnessBuilder},
+        FieldElement,
+    },
+    std::collections::BTreeMap,
+};
+
+/// Collect witness indices from `start..start+len`.
+fn witness_range(start: usize, len: usize) -> Vec<usize> {
```
Bisht13, line 27: Called ~10 times per EC operation. Each call is (start..start+len).collect(). These are contiguous ranges, the indices (os + 3*n, os + 5*n - 2, etc.) can be computed inline without allocation, similar to how digits.rs uses get_digit_witness_index().

provekit/r1cs-compiler/src/msm/ec_points/hints_non_native.rs
```diff
@@ -0,0 +1,494 @@
...
+        5 + 2 * n as u64
+    };
+    check_column_equation_fits(params.limb_bits, max_coeff_sum, n, "On-curve");
+
+    // Allocate hint
+    let os = compiler.num_witnesses();
+    compiler.add_witness_builder(WitnessBuilder::NonNativeEcHint {
+        output_start:    os,
+        op:              NonNativeEcOp::OnCurve,
+        inputs:          vec![px.as_slice().to_vec(), py.as_slice().to_vec()],
```
Bisht13: Every WitnessBuilder::NonNativeEcHint construction copies Limbs data to Vec<usize> via .as_slice().to_vec(). This happens on every EC operation. If WitnessBuilder accepted Limbs (or [usize; MAX_LIMBS]) directly, these heap allocations would be eliminated.

provekit/r1cs-compiler/src/msm/pipeline.rs
```diff
@@ -0,0 +1,407 @@
...
+
+        let p = EcPoint {
+            x: san.px_limbs,
+            y: san.py_limbs,
+        };
+        let r = EcPoint { x: rx, y: ry };
+
+        // On-curve checks
+        {
+            let mut ops = MultiLimbOps::<E>::new(&mut *compiler, &mut *range_checks, params);
```
Bisht13: This explicit reborrowing to create temporary MultiLimbOps values within scoped blocks appears 4 times in this file. It's a workaround for MultiLimbOps holding &mut references, a sign that the struct boundaries are wrong. If compiler and range_checks were passed as function arguments (matching the rest of the codebase) rather than held in a wrapper struct, this wouldn't be needed.

provekit/r1cs-compiler/src/msm/ec_points/hints_non_native.rs
```diff
@@ -0,0 +1,494 @@
...
+        range_checks,
+        Limbs::from(v),
+        &params.p_minus_1_limbs,
+        params.two_pow_w,
+        params.limb_bits,
+    );
+}
+
+/// Compute carry range bits for hint-verified column equations.
+fn carry_range_bits(limb_bits: u32, max_coeff_sum: u64, n: usize) -> u32 {
```
Bisht13: carry_range_bits here and hint_carry_bits in cost_model.rs:79 implement the same formula: limb_bits + ceil_log2(max_coeff_sum * n) + 1 + 1. Similarly, check_column_equation_fits here duplicates column_equation_fits_native_field in the cost model. These should be shared functions, divergence between them would be a soundness bug.

provekit/r1cs-compiler/src/msm/ec_points/hints_non_native.rs
```diff
@@ -0,0 +1,494 @@
...
+        params.limb_bits,
+        crb,
+    );
+
+    // Less-than-p check for x_sq
+    less_than_p_check_vec(compiler, range_checks, &x_sq, params);
+}
+
+/// Hint-verified point doubling for non-native field (multi-limb).
+pub fn point_double_verified_non_native(
```
Bisht13: Both follow: allocate hint → parse layout (identical 15N-6 pattern) → make products → emit 3 column equations → range checks → less-than-p checks. A shared helper parameterized by the equation definitions (product sets + linear terms per equation) would eliminate ~100 lines of structural duplication.

provekit/r1cs-compiler/src/msm/curve/mod.rs
```diff
@@ -0,0 +1,296 @@
...
+    fn offset_x_limbs(&self, limb_bits: u32, num_limbs: usize) -> Vec<FieldElement> {
+        decompose_to_limbs(&self.offset_point().0, limb_bits, num_limbs)
+    }
+    fn offset_y_limbs(&self, limb_bits: u32, num_limbs: usize) -> Vec<FieldElement> {
+        decompose_to_limbs(&self.offset_point().1, limb_bits, num_limbs)
+    }
+
+    /// Compute `[2^n_doublings] * offset_point` on the curve (compile-time
+    /// only).
+    fn accumulated_offset(&self, n_doublings: usize) -> (U256, U256) {
```
Bisht13, line 105: 256 modular inversions (each ~256 squarings via Fermat's little theorem) recomputed on every call. If multiple MSM operations use the same curve and window size, this work is repeated. Consider memoizing or precomputing in the Curve impl.

provekit/r1cs-compiler/src/msm/curve/grumpkin.rs
```diff
@@ -0,0 +1,60 @@
...
+    }
+    fn generator(&self) -> ([u64; 4], [u64; 4]) {
+        ([1, 0, 0, 0], [
+            0x833fc48d823f272c,
+            0x2d270d45f1181294,
+            0xcf135e7506a45d63,
+            0x0000000000000002,
+        ])
+    }
+    fn offset_point(&self) -> ([u64; 4], [u64; 4]) {
```
Bisht13: This is a seemingly random point with no documentation of how it was generated (hash-to-curve? NUMS? Random?). Same for secp256r1.rs:51. For security-relevant parameters in a proof system, provenance should be documented so reviewers can verify the points were generated honestly.

provekit/r1cs-compiler/src/msm/cost_model.rs
```diff
@@ -0,0 +1,539 @@
...
+        assert!(
+            column_equation_fits_native_field(254, limb_bits, num_limbs),
+            "secp256r1 limb_bits={limb_bits} (N={num_limbs}) doesn't fit native field"
+        );
+    }
+
+    #[test]
+    fn test_scalar_relation_cost_grumpkin() {
+        let (sr, rc) = scalar_relation_cost(254, 256);
+        assert!(sr > 50 && sr < 200, "unexpected scalar_relation={sr}");
```
Bisht13: sr > 50 && sr < 200, the actual scalar relation cost is a specific value. If a refactoring halves or doubles the cost, this test won't catch it. Use exact expected values (or tight ±5% bounds) for regression tests.

provekit/r1cs-compiler/src/msm/tests.rs
```diff
@@ -0,0 +1,117 @@
+use {
```
Bisht13: The test suite only verifies happy paths. Missing coverage for:
- Points not on the curve (should be caught by on-curve verification)
- Scalars exceeding the curve order
- Mismatched limb counts
- R1CS constraints correctly rejecting bad witnesses (malicious prover)

provekit/r1cs-compiler/src/msm/mod.rs
```diff
@@ -0,0 +1,205 @@
...
+    },
+    std::collections::BTreeMap,
+};
+
+/// Scalar inputs are split into two 128-bit halves (s_lo, s_hi).
+pub(crate) const SCALAR_HALF_BITS: usize = 128;
+
+/// Integer ceiling of log2.
+/// ceil_log2(1) = 0, ceil_log2(2) = 1, ceil_log2(3) = 2, ceil_log2(4) = 2.
+pub(crate) fn ceil_log2(n: u64) -> u32 {
```
Bisht13: This function is also defined in prover/bigint_mod.rs. The new provekit_common::u256_arith module was created for exactly this kind of shared utility, ceil_log2 should live there.

provekit/r1cs-compiler/src/msm/curve/mod.rs
```diff
@@ -0,0 +1,296 @@
...
+    0
+}
+
+// ---------------------------------------------------------------------------
+// Free functions
+// ---------------------------------------------------------------------------
+
+/// Decompose a 256-bit value into `num_limbs` limbs of `limb_bits` width each,
+/// returned as FieldElements.
+pub fn decompose_to_limbs(val: &U256, limb_bits: u32, num_limbs: usize) -> Vec<FieldElement> {
```
Bisht13, line 135: If limb_bits is 0, word_shift and bit_shift are both 0, the shift loop becomes a no-op, and the function enters an infinite conceptual loop (remaining never changes). Add an assert!(limb_bits > 0) or return an error.

provekit/r1cs-compiler/src/msm/curve/u256_arith.rs
```diff
@@ -0,0 +1,29 @@
+//! 256-bit modular arithmetic for compile-time EC point computations.
```
Bisht13: Does this file need to exist in curve module?

provekit/r1cs-compiler/src/msm/mod.rs
ocdbytes: This will be a very big refactor and unwraps are right now used in the whole compiler. witness_generator is a different logic. In other blackboxes also we use unwrap()s throughout. If error propagation is necessary we can create a new issue.

provekit/r1cs-compiler/src/msm/multi_limb_ops.rs
ocdbytes: the zero values are blocked by type system and are unreachable in the code. If this is the case do we need this ?

provekit/r1cs-compiler/src/msm/ec_points/mod.rs
ocdbytes: this would lead to ownership restructuring across modules. The code is monomorphized right now If i implement this change there would be a lot of places for ownership issues. let me know if this is worth doing ?

provekit/r1cs-compiler/src/msm/multi_limb_ops.rs
ocdbytes: idts I can decouple multilimb ops in a way that It would reduce the complexity. phantom data is needed for the

provekit/r1cs-compiler/src/msm/multi_limb_arith.rs
ocdbytes: Not worth it I would say usize is being used over the whole compiler. If I introduce this change in this PR then it would become very difficult to review. If this type is actually needed we can have a separate issue for this.

provekit/prover/src/witness/witness_builder.rs
ocdbytes: Again this is a very big change and if I add a witness op trait I would need to refactor a lot of files and it will be over engineering here for this use case. match arm works properly for our use case. I mean it is a design choice for the provekit let me know your thoughts.

provekit/r1cs-compiler/src/msm/curve/mod.rs
ocdbytes: this will increase the duplication right ?

provekit/r1cs-compiler/src/msm/limbs.rs
ocdbytes: we track the length separately right ? is this needed then or am I understanding something wrong ?

provekit/r1cs-compiler/src/msm/ec_points/hints_non_native.rs
ocdbytes: this is during compilation phase inlining would increase duplication

provekit/r1cs-compiler/src/msm/pipeline.rs
ocdbytes: the current structure is for readability if I dissolve it then it would add extra params.

provekit/r1cs-compiler/src/msm/curve/mod.rs
ocdbytes: it is called only once per circuit compilation and idts it uses field inversions.

provekit/r1cs-compiler/src/msm/tests.rs
ocdbytes: these tests can actually be removed because they were added initially just to check if constraints are generated properly. Actual testing now happens in the msm_witness_solving.rs file. I will remove these tests

provekit/r1cs-compiler/src/msm/curve/u256_arith.rs
ocdbytes: it contains a single function which is very specific to curves so this file is ok here. maybe I can rename it to something aprop.

provekit/r1cs-compiler/src/msm/ec_points/hints_non_native.rs
ocdbytes: it is an overestimate of deduplication as the terms and everything is very different in each implementation

provekit/r1cs-compiler/src/msm/mod.rs
Bisht13: Agree, not specific to this PR. Can you create a separate issue to track?

provekit/r1cs-compiler/src/msm/multi_limb_ops.rs
Bisht13: The zero values aren't blocked by the type system, for_curve_order() returns a MultiLimbParams with curve_a_raw: [0; 4], and any code holding a &MultiLimbParams can read those fields. They're unreachable by convention, not by types.

provekit/r1cs-compiler/src/msm/ec_points/mod.rs
Bisht13: Non blocking right now, can we create an issue to track this?

provekit/prover/src/witness/witness_builder.rs
Bisht13: Can you track this via an issue?

provekit/r1cs-compiler/src/msm/curve/mod.rs
Bisht13: No, a CurveDecomposition struct with a single fn new(curve: &impl Curve, limb_bits, num_limbs) -> Self would reduce duplication. Instead of 10 trait methods each calling decompose_to_limbs, you'd have one struct construction.

provekit/r1cs-compiler/src/msm/curve/mod.rs
Bisht13: It does use field inversions, ec_point_double calls mod_inverse (Fermat's little theorem: a^{p-2} mod p via mod_pow). But it's once per compilation, so the cost is negligible.

provekit/r1cs-compiler/src/msm/tests.rs
Bisht13: Okay can we keep the tests where they belong and add the missing coverage?

provekit/r1cs-compiler/src/msm/curve/u256_arith.rs
Bisht13: Maybe shift to mod.rs then?

provekit/r1cs-compiler/src/msm/ec_points/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+}
+
+// ---------------------------------------------------------------------------
+// NativeEcOps — hint-verified via raw R1CS (num_limbs=1)
+// ---------------------------------------------------------------------------
+
+/// Native-field EC operations via hint-verified R1CS constraints.
+pub(crate) struct NativeEcOps;
+
+impl FieldArith for NativeEcOps {
```
Bisht13: NativeEcOps implements FieldArith by forwarding all 4 methods to NativeSingleField (lines 59-98). Same for NonNativeEcOps → MultiLimbField (lines 147-186). That's 32 lines of pure delegation boilerplate.

Consider making MultiLimbOps generic over separate <F: FieldArith, E: EcOps> type parameters so EC types don't need to re-implement field ops.

provekit/r1cs-compiler/src/msm/multi_limb_ops.rs
```diff
@@ -0,0 +1,456 @@
...
+    ) -> Self {
+        let two_pow_w = FieldElement::from(2u64).pow([limb_bits as u64]);
+        Self {
+            num_limbs,
+            limb_bits,
+            p_limbs: curve.curve_order_n_limbs(limb_bits, num_limbs),
+            p_minus_1_limbs: curve.curve_order_n_minus_1_limbs(limb_bits, num_limbs),
+            two_pow_w,
+            modulus_raw: curve.curve_order_n(),
+            curve_a_limbs: vec![FieldElement::ZERO; num_limbs], // unused
```
Bisht13: for_curve_order() fills curve_a_* and curve_b_* with zeros (lines 257-260). The is_native/modulus_fe removal was a good improvement, but these 4 fields are still meaningless for scalar relation verification. A ModulusParams / CurveParams split would make this impossible to misuse.

provekit/r1cs-compiler/src/msm/tests.rs
ocdbytes: I ahve removed the tests because they are now obsolete after witness solving tests

provekit/r1cs-compiler/src/msm/mod.rs
ocdbytes: #332

provekit/r1cs-compiler/src/msm/ec_points/mod.rs
ocdbytes: https://github.com/worldfnd/provekit/issues/333

provekit/prover/src/witness/witness_builder.rs
ocdbytes: https://github.com/worldfnd/provekit/issues/334

provekit/r1cs-compiler/src/msm/multi_limb_arith.rs
```diff
@@ -0,0 +1,486 @@
...
+
+    is_zero
+}
+
+// ---------------------------------------------------------------------------
+// N≥2 multi-limb path (generalization of wide_ops.rs)
+// ---------------------------------------------------------------------------
+
+/// Shared core for `add_mod_p_multi` and `sub_mod_p_multi`.
+fn add_sub_mod_p_core(
```
rose2221, line 71: In `add_sub_mod_p_core`,` negate_mod_p_multi`, and `less_than_p_check_multi`, each limb iteration emits two linear constraints: one for the intermediate sum (`v_offset/v/v_diff`) and one for the remainder (`r[i]/ri/d_i`). These can be merged into a single constraint per limb.

For example in `add_sub_mod_p_core`, currently:
```
C1: (a[i] + b[i] + offset - p[i]*q + carry_prev) * 1 = v_offset
C2: (v_offset - carry * 2^W) * 1 = r[i]
```
Can become:
`CM: (a[i] + b[i] + offset - p[i]*q + carry_prev - carry * 2^W) * 1 = r[i]`

provekit/r1cs-compiler/src/msm/multi_limb_arith.rs
```diff
@@ -0,0 +1,486 @@
...
+}
+
+/// Checks if value is zero or not (used by all N values).
+/// Returns a boolean witness: 1 if zero, 0 if non-zero.
+///
+/// Uses SafeInverse (not Inverse) because the input value may be zero.
+/// SafeInverse outputs 0 when the input is 0, and is solved in the Other
+/// layer (not batch-inverted), so zero inputs don't poison the batch.
+#[must_use]
+pub fn compute_is_zero(compiler: &mut NoirToR1CSCompiler, value: usize) -> usize {
```
rose2221, line 30: nit: `compute_is_zero` allocates a `value_mul_value_inv` witness that never appears in any constraint, it's only used to `compute is_zero` via the Sum builder. Could be eliminated with a direct `WitnessBuilder::IsZero` which can be created explicitly

provekit/r1cs-compiler/src/msm/pipeline.rs
```diff
@@ -0,0 +1,412 @@
...
+    // Expected accumulated offset
+    let glv_num_windows = (half_bits + window_size - 1) / window_size;
+    let glv_n_doublings = glv_num_windows * window_size;
+    let (acc_off_x_raw, acc_off_y_raw) = curve.accumulated_offset(glv_n_doublings);
+    let acc_off_x_values = decompose_to_limbs_pub(&acc_off_x_raw, limb_bits, num_limbs);
+    let acc_off_y_values = decompose_to_limbs_pub(&acc_off_y_raw, limb_bits, num_limbs);
+
+    for pt in merged_points {
+        let offset = EcPoint {
+            x: ops.constant_limbs(offset_x_values),
```
rose2221: `constant_limbs(offset_x_values)` and `constant_limbs(offset_y_values)` are called inside the per-point loop, allocating fresh pinned witnesses and constraints each iteration for identical values. Limbs is `Copy,` so these can be hoisted before the loop and reused. Same applies to the duplicate `offset_x` allocations in `accumulate_and_constrain_outputs`

provekit/r1cs-compiler/src/msm/multi_limb_arith.rs
ocdbytes: makes sense

provekit/r1cs-compiler/src/msm/multi_limb_arith.rs
ocdbytes: it is called once per msm point and change will add another variant tot witness builder for like 2-3 witness count. Do you think we should go for this change ?

provekit/r1cs-compiler/src/msm/pipeline.rs
ocdbytes: after guassian elimination idts it should matter but I have added this change because it looks clean and the right way to add constants

provekit/r1cs-compiler/src/msm/pipeline.rs
```diff
@@ -0,0 +1,415 @@
...
+
+    let sub_pt = EcPoint {
+        x: {
+            let g_x = ops.constant_limbs(&gen_x_limb_values);
+            ops.select(all_skipped, offset_x, g_x)
+        },
+        y: {
+            let neg_off_y = ops.constant_limbs(&neg_offset_y_values);
+            let neg_g_y = ops.constant_limbs(&neg_gen_y_values);
+            ops.select(all_skipped, neg_off_y, neg_g_y)
```
rose2221: `all_skipped` is already boolean-constrained by the `ops.select` at line 395. The second `ops.select` adds a duplicate `flag * flag = flag` constraint (via `constrain_flag` inside select). Could use `select_unchecked` here instead.

provekit/r1cs-compiler/src/msm/mod.rs
```diff
@@ -0,0 +1,208 @@
...
+            config,
+            range_checks,
+            curve,
+        );
+    }
+}
+
+/// Resolves a `ConstantOrR1CSWitness` to a witness index.
+#[must_use]
+fn resolve_input(compiler: &mut NoirToR1CSCompiler, input: &ConstantOrR1CSWitness) -> usize {
```
Bisht13, line 148: Move to prover/src/witness/witness_builder.rs since there is resolve() there already.

provekit/prover/src/witness/witness_builder.rs
```diff
@@ -23,6 +32,102 @@ pub trait WitnessBuilderSolver {
...
+/// Resolve a ConstantOrR1CSWitness to its FieldElement value.
+fn resolve(witness: &[Option<FieldElement>], v: &ConstantOrR1CSWitness) -> FieldElement {
+    match v {
+        ConstantOrR1CSWitness::Constant(c) => *c,
+        ConstantOrR1CSWitness::Witness(idx) => witness[*idx].unwrap(),
+    }
+}
+
+/// Convert a u128 value to a FieldElement.
+fn u128_to_fe(val: u128) -> FieldElement {
```
Bisht13: Move this function somewhere like utils or similar.

provekit/prover/src/witness/witness_builder.rs
```diff
@@ -23,6 +32,102 @@ pub trait WitnessBuilderSolver {
...
+    FieldElement::from_bigint(ark_ff::BigInt([val as u64, (val >> 64) as u64, 0, 0])).unwrap()
+}
+
+/// Read witness limbs and reconstruct as [u64; 4].
+fn read_witness_limbs(
```
Bisht13: Move to limbs.rs?

provekit/prover/src/witness/witness_builder.rs
```diff
@@ -23,6 +32,102 @@ pub trait WitnessBuilderSolver {
...
+            );
+            let bigint = witness[idx].unwrap().into_bigint().0;
+            bigint[0] as u128 | ((bigint[1] as u128) << 64)
+        })
+        .collect();
+    reconstruct_from_u128_limbs(&limb_values, limb_bits)
+}
+
+/// Write u128 limb values as FieldElement witnesses starting at `start`.
+fn write_limbs(witness: &mut [Option<FieldElement>], start: usize, vals: &[u128]) {
```
Bisht13: Move to limbs.rs?

provekit/prover/src/witness/witness_builder.rs
```diff
@@ -23,6 +32,102 @@ pub trait WitnessBuilderSolver {
...
+    for (i, &val) in vals.iter().enumerate() {
+        witness[start + i] = Some(u128_to_fe(val));
+    }
+}
+
+/// Split a signed quotient into `(q_pos, q_neg)` limb vectors.
+fn split_quotient(q_abs: Vec<u128>, is_neg: bool, n: usize) -> (Vec<u128>, Vec<u128>) {
```
Bisht13: Move to bigint_mod.rs?

provekit/prover/src/witness/witness_builder.rs
```diff
@@ -23,6 +32,102 @@ pub trait WitnessBuilderSolver {
...
+        (q_abs, vec![0u128; n])
+    }
+}
+
+/// Solve one column equation: quotient → split → write → carries → write.
+///
+/// Computes `(LHS - RHS) / p` as a signed quotient, splits into (q_pos, q_neg),
+/// computes verification carries, and writes all witnesses at `os + offset`.
+/// Witness layout at offset: `[q_pos(N), q_neg(N), carries(2N-2)]`.
+fn solve_and_write_equation(
```
Bisht13: Move to bigint_mod.rs?

provekit/prover/src/bigint_mod.rs
```diff
@@ -0,0 +1,1194 @@
...
+            if !b {
+                break;
+            }
+        }
+    }
+    result
+}
+
+/// Add two 4-limb (256-bit) numbers, returning a 5-limb result with carry.
+pub fn add_4limb(a: &[u64; 4], b: &[u64; 4]) -> [u64; 5] {
```
Bisht13, line 126: Do we want to keep limb fns here or limbs.rs?

provekit/prover/src/bigint_mod.rs
```diff
@@ -0,0 +1,1194 @@
...
+/// Reconstruct a 256-bit scalar from two 128-bit halves: `scalar = lo + hi *
+/// 2^128`.
+pub fn reconstruct_from_halves(lo: &[u64; 4], hi: &[u64; 4]) -> [u64; 4] {
+    [lo[0], lo[1], hi[0], hi[1]]
+}
+
+/// EC point doubling with lambda exposed: returns (lambda, x3, y3).
+///
+/// Used by the `EcDoubleHint` prover which needs lambda as a witness.
+pub fn ec_point_double_with_lambda(
```
Bisht13: Maybe create an ec rust file or ec_utils?

provekit/prover/src/lib.rs
```diff
@@ -25,9 +25,10 @@ use {
     whir::transcript::VerifierMessage,
 };
 
+pub mod bigint_mod;
 #[cfg(not(target_arch = "wasm32"))]
 pub mod input_utils;
-mod r1cs;
+pub mod r1cs;
```
Bisht13: Does these modules need to be public?

provekit/r1cs-compiler/src/msm/curve/secp256r1.rs
```diff
@@ -0,0 +1,74 @@
+use super::Curve;
+
+/// SECP256R1 (NIST P-256).
+/// Equation: y² = x³ + ax + b
+pub struct Secp256r1;
+
+impl Curve for Secp256r1 {
+    fn field_modulus_p(&self) -> [u64; 4] {
```
Bisht13: Codebase uses [u64; 4] and U256, use only U256 by type alias.

provekit/r1cs-compiler/src/constraint_helpers.rs
```diff
@@ -0,0 +1,133 @@
...
+/// on_false)`.
+///
+/// Uses a single witness + single R1CS constraint:
+///   flag * (on_true - on_false) = result - on_false
+pub(crate) fn select_witness(
+    compiler: &mut NoirToR1CSCompiler,
+    flag: usize,
+    on_false: usize,
+    on_true: usize,
+) -> usize {
```
Bisht13, line 31: Every witness index is usize, something like this function has 4 usize parameters which is bad practice. We should have something like WitnessIdx.

provekit/prover/src/witness/witness_builder.rs
Bisht13: Several functions take 8-12 parameters of the same type, like this and emit_schoolbook_column_equations, group similar parameters into structs, for example:
```rs
struct ColumnEquationInput<'a> {
      products: &'a [(&'a [Vec<usize>], FieldElement)],
      linear: &'a [(&'a [usize], FieldElement)],
      quotient_pos: &'a [usize],
      quotient_neg: Option<&'a [usize]>,
      carries: &'a [usize],
  }
```

provekit/r1cs-compiler/src/msm/multi_limb_arith.rs
```diff
@@ -0,0 +1,485 @@
...
+/// Shared core for `add_mod_p_multi` and `sub_mod_p_multi`.
+fn add_sub_mod_p_core(
+    compiler: &mut NoirToR1CSCompiler,
+    range_checks: &mut BTreeMap<u32, Vec<usize>>,
+    op: ModularOp,
+    a: Limbs,
+    b: Limbs,
+    params: &ModulusParams,
+) -> Limbs {
+    let n = a.len();
```
Bisht13, line 79: Use better naming for variables. Issues at other places as well.

provekit/r1cs-compiler/src/msm/ec_points/hints_non_native.rs
```diff
@@ -0,0 +1,462 @@
...
+    } else {
+        5 + 2 * n as u64
+    };
+    check_column_equation_fits(params.limb_bits, max_coeff_sum, n, "On-curve");
+
+    // Allocate hint
+    let os = compiler.num_witnesses();
+    compiler.add_witness_builder(WitnessBuilder::NonNativeEcHint {
+        output_start:    os,
+        op:              NonNativeEcOp::OnCurve,
```
Bisht13, line 177: Three very similar blocks for Double/Add/OnCurve equation setup, each follows the pattern: allocate hint → parse layout → make products → emit column equations → range check. A table-driven approach would  reduce repetition.

Bisht13: Similar structure in WitnessBuilder.

provekit/r1cs-compiler/src/msm/ec_points/hints_non_native.rs
```diff
@@ -0,0 +1,462 @@
...
+    });
+
+    let h = EcHint3Eq::parse(os, n);
+    let px_s = px.as_slice();
+    let py_s = py.as_slice();
+
+    // Eq1: 2*lambda*py - 3*px*px - a = q1*p
+    let prod_lam_py = make_products(compiler, &h.lambda, py_s);
+    let prod_px_px = make_products(compiler, px_s, px_s);
+    let a_limbs = allocate_pinned_constant_limbs(compiler, &params.ec.curve_a_limbs[..n]);
```
rose2221: `allocate_pinned_constant_limbs` is called here on every invocation of `point_double_verified_non_native`, allocating n witnesses and n constraints each time, all holding the same curve_a values.

For secp256r1 with w=4, this function is called ~130 times per point, creating ~129×N redundant witnesses and constraints per point.

Same pattern applies to curve_b_limbs at line 215 in `verify_on_curve_non_nativ`e .

provekit/r1cs-compiler/src/constraint_helpers.rs
ocdbytes: #347

provekit/r1cs-compiler/src/msm/mod.rs
ocdbytes: these are two different functions

provekit/prover/src/witness/witness_builder.rs
ocdbytes: moved to limb_io.rs file

provekit/prover/src/bigint_mod.rs
ocdbytes: moved to ec_arith.rs

provekit/prover/src/lib.rs
ocdbytes: converted to pub(crate)

PR #311, opened 2026-03-03
feat: Implement Gaussian elimination optimization for R1CS

rose2221:
### Summary
- Implements a Gaussian elimination (GE) optimization pass that identifies linear R1CS constraints, picks pivot variables by occurrence-count heuristic, and substitutes them into remaining constraints, reducing constraint count before proving.
- Makes SHA256 call count configurable in noir_sha256 example circuit via `NUM_SHA_CALLS `constant
### Constraint Reduction Results
| Circuit                     | Before   | After    | Reduction                                   |
|-----------------------------|----------|----------|---------------------------------------------|
| Poseidon2                  | 272      | 264      | 2.9% (8 linear constraints eliminated)      |
| SHA256 (1 compression)     | 16,561   | 13,988   | 15.5% (2,573 linear constraints eliminated) |
| Complete Age Check         | 712,399  | 630,346  | 11.5% (82,053 linear constraints eliminated)|
### How it works
The optimization runs in 4 phases after Noir ACIR is compiled to R1CS:

**Phase 1** — Scan for linear constraints (where A or B matrix row is constant)
**Phase 2** — For each linear constraint, pick a pivot variable (fewest occurrences, not public input) and build a substitution expression. Forward chain resolution inlines earlier pivots into new substitutions. Self-referencing pivots are algebraically rescaled via (1 - r_p)^{-1}
**Phase 2b** — Reverse-order pass resolves backward chains (earlier substitutions referencing later pivots), ensuring every substitution's terms reference only non-pivot columns
**Phase 3** — Single-pass substitution into all remaining constraints, followed by row removal

### DIFF
--- a/provekit/common/src/optimize.rs
+++ b/provekit/common/src/optimize.rs
@@ -28,10 +28,10 @@ struct Substitution {
 
 /// Statistics from the optimization pass.
 pub struct OptimizationStats {
-    pub constraints_before:  usize,
-    pub constraints_after:   usize,
-    pub eliminated:          usize,
-    pub eliminated_columns:  HashSet<usize>,
+    pub constraints_before: usize,
+    pub constraints_after:  usize,
+    pub eliminated:         usize,
+    pub eliminated_columns: HashSet<usize>,
 }
 
 impl OptimizationStats {
@@ -371,9 +371,10 @@ mod tests {
             .collect()
     }
 
-    /// Assert that no remaining constraint references an eliminated pivot column.
-    /// This is the chain-resolution invariant: after GE, every substituted pivot
-    /// must have been fully inlined into all remaining constraints.
+    /// Assert that no remaining constraint references an eliminated pivot
+    /// column. This is the chain-resolution invariant: after GE, every
+    /// substituted pivot must have been fully inlined into all remaining
+    /// constraints.
     fn assert_no_dangling_pivots(r1cs: &R1CS, stats: &OptimizationStats) {
         for row in 0..r1cs.num_constraints() {
             for (col, _) in r1cs.a.iter_row(row) {


provekit/common/src/optimize.rs
```diff
@@ -0,0 +1,663 @@
...
+    sorted_entries.sort_by_key(|(col, _)| *col);
+
+    matrix.replace_row(row, &sorted_entries);
+}
+
+#[cfg(test)]
+mod tests {
+    use {super::*, crate::witness::SumTerm, ark_std::One};
+
+    #[test]
```
Bisht13, line 420: Tests check structure (counts, pivot columns absent) but never check arithmetic correctness, a sign/coefficient error in substitution would pass all tests while producing invalid R1CS.

Need at least one test that constructs a known witness `w`, runs `optimize_r1cs`, then asserts `A·w ⊙ B·w == C·w` on the optimized matrices.

provekit/common/src/optimize.rs
```diff
@@ -0,0 +1,663 @@
...
+        .into_iter()
+        .filter(|(_, v)| !v.is_zero())
+        .map(|(col, val)| (col, interner.intern(val)))
+        .collect();
+    sorted_entries.sort_by_key(|(col, _)| *col);
+
+    matrix.replace_row(row, &sorted_entries);
+}
+
+#[cfg(test)]
```
Bisht13, line 356: All 4 tests use `A=[1*w0], B=[1*w0]`, only the "both constant" branch of extract_linear_expression is covered. Missing:
- A-only constant branch (else if a_is_const)
- B-only constant branch (else)
- Non-trivial coefficients (everything uses ±1)
- Self-referencing pivot rescaling (lines 158–170)
- Degenerate skip case (1 - r_p) = 0 (line 167)

provekit/common/src/sparse_matrix.rs
```diff
@@ -501,6 +501,91 @@ impl SparseMatrix {
             self.values.push(value);
         }
     }
+
+    /// Check if a row has exactly one entry and it's at the given column.
+    pub fn row_is_single_entry_at_col(&self, row: usize, col: usize) -> bool {
```
Bisht13: row_is_single_entry_at_col, row_is_empty (here), and get_or_none (interner.rs:45) are unused.

provekit/common/src/optimize.rs
```diff
@@ -0,0 +1,663 @@
...
+    // Note: We do NOT modify witness builders. The witnesses are still
+    // computed by their original builders. GE only removes redundant
+    // constraints and substitutes pivots into remaining constraints.
+
+    let constraints_after = r1cs.num_constraints();
+    let eliminated = substitutions.len();
+
+    // witnesses_after = witnesses_before since we don't actually remove columns,
+    // just make some witnesses derived. The column count doesn't change.
+    let witnesses_after = witnesses_before;
```
Bisht13: Always identical, witness_reduction_percent() always returns 0%. The display still prints WITNESS REDUCTION: 0.00% which implies this pass reduces witnesses.

provekit/common/src/optimize.rs
```diff
@@ -0,0 +1,663 @@
...
+///
+/// Identifies linear constraints (where at least one of A or B is constant),
+/// picks pivots, substitutes into remaining constraints, and removes the
+/// eliminated rows.
+///
+/// `num_public_inputs` columns (1..=num_public_inputs) and column 0 (constant
+/// one) are never chosen as pivots.
+pub fn optimize_r1cs(
+    r1cs: &mut R1CS,
+    _witness_builders: &mut [WitnessBuilder],
```
Bisht13: Takes &mut but never reads or writes. Consider &[WitnessBuilder] until mutation is actually needed.

provekit/common/src/r1cs.rs
```diff
@@ -127,4 +129,111 @@ impl R1CS {
...
+        }
+
+        terms
+            .into_iter()
+            .filter(|(_, v)| !v.is_zero())
+            .map(|(col, val)| (val, col))
+            .collect()
+    }
+
+    /// Remove constraints at the given sorted row indices from all three
```
Bisht13: Doc says "sorted row indices" but implementation uses HashSet, ordering doesn't matter. Either enforce with debug_assert! or drop the "sorted" claim.

provekit/common/src/r1cs.rs
```diff
@@ -127,4 +129,111 @@ impl R1CS {
...
+                    .expect("interned value missing");
+                *terms.entry(col).or_insert_with(FieldElement::zero) -= val;
+            }
+        }
+
+        terms
+            .into_iter()
+            .filter(|(_, v)| !v.is_zero())
+            .map(|(col, val)| (val, col))
+            .collect()
```
Bisht13: Collected from HashMap::into_iter(), non-deterministic ordering. Pivot selection with tied counts varies across runs. Not a bug, but sorting by column index would make optimization reproducible.

provekit/common/src/optimize.rs
rose2221: Added `test_arithmetic_correctness` :- it builds an R1CS with a concrete witness, runs optimize_r1cs, then asserts A·w ⊙ B·w == C·w on the optimized matrices.

provekit/common/src/sparse_matrix.rs
rose2221: removed

provekit/common/src/optimize.rs
rose2221: Added test_branch_coverage that covers all these cases.

provekit/common/src/optimize.rs
rose2221: i removed the witness reduction display as of now and will introduce in the next PR.

provekit/common/src/r1cs.rs
rose2221: edited the doc

provekit/common/src/r1cs.rs
rose2221: Sorted` extract_linear_expression` output by column index so pivot selection is deterministic across runs.

provekit/common/src/optimize.rs
```diff
@@ -0,0 +1,735 @@
...
+/// one) are never chosen as pivots.
+pub fn optimize_r1cs(r1cs: &mut R1CS) -> OptimizationStats {
+    let constraints_before = r1cs.num_constraints();
+
+    // Columns that must not be eliminated:
+    // - Column 0: constant one
+    // - Columns 1..=num_public_inputs: public inputs
+    let mut forbidden: HashSet<usize> = HashSet::new();
+    forbidden.insert(0);
+    for i in 1..=r1cs.num_public_inputs {
```
Bisht13: `num_public_inputs` is never set in the Noir compilation path, `R1CS::new()` defaults it to 0, and `noir_to_r1cs::finalize()` never updates it. So this loop is 1..=0, which means the forbidden set is just {0} and public input columns are eligible for pivot elimination.

Not a soundness bug (the proof protocol binds public inputs independently), but it defeats the documented intent. Additionally, even if set, the positional assumption 1..=N doesn't hold for Noir, public inputs land at arbitrary R1CS column positions.

provekit/common/src/sparse_matrix.rs
```diff
@@ -501,6 +501,80 @@ impl SparseMatrix {
...
+    }
+
+    /// Get all (col, value) entries for a row as a Vec.
+    pub fn get_row_entries(&self, row: usize) -> Vec<(usize, InternedFieldElement)> {
+        self.iter_row(row).collect()
+    }
+
+    /// Replace a row's entries entirely. The new entries must be sorted by
+    /// column.
+    pub fn replace_row(&mut self, row: usize, entries: &[(usize, InternedFieldElement)]) {
```
Bisht13, line 522: Doc says entries must be sorted by column, but there's no debug_assert!. encode_col_deltas already has the precedent:
```rs
debug_assert!(row_cols.windows(2).all(|w| w[0] <= w[1]), "Column indices must be sorted within each row");
```
Worth adding the same assertion here to catch future callers that forget to sort.

provekit/common/src/optimize.rs
```diff
@@ -0,0 +1,735 @@
...
+
+    // Build occurrence counts across all three matrices for pivot selection
+    // heuristic
+    let mut occurrence_counts = build_occurrence_counts(r1cs);
+
+    // Also track pivot_col -> substitution index for chain resolution
+    let mut sub_map_phase2: HashMap<usize, usize> = HashMap::new();
+
+    for &row in &linear_rows {
+        // Extract the linear expression from C[row]: sum of (coeff * w_i) = 0
```
Bisht13: Misleading. extract_linear_expression combines const * A/B[row] - C[row], not just C[row]. Should say something like "Extract the combined linear expression from this constraint".

provekit/common/src/optimize.rs
```diff
@@ -0,0 +1,735 @@
...
+
+        let terms: Vec<(FieldElement, usize)> = resolved
+            .into_iter()
+            .filter(|(_, v)| !v.is_zero())
+            .map(|(col, val)| (val, col))
+            .collect();
+
+        // Decrement occurrence counts for all columns in this row (they're being
+        // removed)
+        for (_, col) in &expr {
```
Bisht13, line 164: Nit: this decrements using the combined expr, but occurrence_counts was built from raw A+B+C entries. A column in all three matrices gets counted 3× but only decremented 1×. Only affects pivot-selection heuristic quality, not correctness, but worth a comment noting the approximation.

rose2221: addressed it by updating the comment.

provekit/common/src/optimize.rs
rose2221: updated the comemnt.

provekit/common/src/sparse_matrix.rs
rose2221: added the same, thanks.

provekit/common/src/optimize.rs
rose2221: fixed, removed the dead loop and replaced hard-coded pivot assertions in tests with a generic `assert_no_dangling_pivots` check using eliminated_columns from OptimizationStats

Bisht13 (approved):

PR #316, opened 2026-03-06
fix: R1CS binding in transcript [LA - I] 

ashpect:


provekit/prover/src/lib.rs
```diff
@@ -227,10 +232,20 @@ impl Prove for MavrosProver {
             &params,
         );
 
+        let num_public_inputs = self.num_public_inputs;
+        let public_inputs = if num_public_inputs == 0 {
+            PublicInputs::new()
+        } else {
+            // TODO : Verify marvos prover's handling of public input
+            PublicInputs::from_vec(phase1.out_wit_pre_comm[1..=num_public_inputs].to_vec())
```
veljkovranic, line 270: This is a correct assumption.

Here is the old code (mroe explicit) of how the values where handled:

```   
 out_wit_pre_comm[0] = Field::ONE;
 let flat_inputs = flatten_param_vec(ordered_inputs);
 for i in 0..flat_inputs.len() {
     out_wit_pre_comm[1 + i] = flat_inputs[i];
 }
```

provekit/r1cs-compiler/src/whir_r1cs.rs
```diff
@@ -154,6 +155,7 @@ impl WhirR1CSSchemeBuilder for WhirR1CSScheme {
             w1_size,
             num_challenges,
             has_public_inputs,
+            r1cs_hash: [0u8; 32], // TODO: Mavros path needs r1cs_hash
```
veljkovranic: `new_from_mavros_r1cs` is being invoked  in this snippet of noir_proof_scheme generation

```        let whir_for_witness = WhirR1CSScheme::new_from_mavros_r1cs(
            &mavros_r1cs,
            mavros_r1cs.witness_layout.pre_commitment_size(),
            mavros_r1cs.witness_layout.challenges_size,
            num_public_inputs > 0,
            hash_config.engine_id(),
        );

        let r1cs = convert_mavros_r1cs_to_provekit(&mavros_r1cs);
```

You can just revert the order of these two instructions and that will give you the R1CS type you need to perform the hashing.

provekit/r1cs-compiler/src/whir_r1cs.rs
```diff
@@ -173,8 +154,8 @@ impl WhirR1CSSchemeBuilder for WhirR1CSScheme {
             whir_witness: Self::new_whir_zk_config_for_size(m, 1, hash_id),
             w1_size,
             num_challenges,
-            challenge_offsets,
             has_public_inputs,
+            r1cs_hash: [0u8; 32],
```
Bisht13: Any future caller of new_from_dimensions or new_from_mavros_r1cs silently gets a zero R1CS hash, the R1CS won't be bound to the transcript. Consider making r1cs_hash a required parameter, or adding debug_assert_ne!(self.r1cs_hash, [0u8; 32]) in create_domain_separator().

provekit/verifier/src/whir_r1cs.rs
```diff
@@ -45,7 +44,8 @@ impl WhirR1CSVerifier for WhirR1CSScheme {
         r1cs: &R1CS,
         hash_config: HashConfig,
     ) -> Result<()> {
-        let ds = self.create_domain_separator().instance(&Empty);
+        let instance = public_inputs.hash_bytes();
```
Bisht13, line 56: The r1cs_hash is bound via the protocol ID, but an explicit check here gives a clear error instead of a cryptic WHIR verification failure.

provekit/common/src/witness/mod.rs
```diff
@@ -95,6 +95,16 @@ impl PublicInputs {
             _ => self.0.iter().copied().reduce(compress).unwrap(),
         }
     }
+
+    #[must_use]
```
Bisht13, line 102: Nit: missing /// doc comment

provekit/common/src/whir_r1cs.rs
```diff
@@ -25,9 +25,9 @@ pub struct WhirR1CSScheme {
     pub m_0:               usize,
     pub a_num_terms:       usize,
     pub num_challenges:    usize,
-    pub challenge_offsets: Vec<usize>,
     pub has_public_inputs: bool,
     pub whir_witness:      WhirZkConfig,
+    pub r1cs_hash:         [u8; 32],
```
Bisht13: Consider a Hash newtype in common instead of raw [u8; 32]. This PR introduces three places using [u8; 32] as a hash (r1cs_hash field, R1CS::hash() return, PublicInputs::hash_bytes() return). A shared newtype makes the uninitialized [0u8; 32] in new_from_dimensions explicit (e.g. Hash::UNSET), prevents mixing up arbitrary byte arrays with hashes, and is reusable.

provekit/common/src/whir_r1cs.rs
```diff
@@ -18,6 +18,24 @@ pub type WhirZkConfig = GenericWhirZkConfig<FieldElement>;
 /// Type alias for the whir domain separator used in provekit's outer protocol.
 type WhirDomainSeparator = transcript::DomainSeparator<'static, ()>;
 
+/// SHA3-256 hash of a serialized R1CS instance, used to bind the Fiat-Shamir
+/// transcript to a concrete circuit.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+pub struct R1csHash([u8; 32]);
```
Bisht13, line 24: Can we have it as just Hash (and have it in the proper file)?

ashpect: https://github.com/worldfnd/provekit/pull/316#discussion_r3042207863
added r1cshash acc to this, keeping the name as just hash might be misleading and can be confusing as skyscraper hash can have diff type. R1CShash explictely uses sha256 rn.

Bisht13 (approved):

PR #321, opened 2026-03-12
fix: public input binding verification

ashpect:
## Problem

The verifier hashes `public_inputs` to derive the Fiat-Shamir challenge `x`, but the check for `public_eval`, the inner product `⟨[1, x, x², …], committed_poly⟩` matching those inputs was missing. The prover sends `public_eval` as a `prover_hint_ark` (not absorbed into the transcript), so a malicious prover can commit to one set of values and claim another as public output. Verification passes because WHIR only checks the evaluation claims it receives, not where they came from.

Two bugs:

1. **`public_eval` was not transcript-bound.** Sent via `prover_hint_ark` instead of `prover_message`, so it was never absorbed into Fiat-Shamir. A prover could swap it without affecting subsequent challenges.

2. **Off-by-one in public weight length.** `make_public_weight` and `compute_public_eval` received `public_inputs.len()` but the witness layout has an R1CS constant `1` at position 0 before the public inputs, so the weight vector was one element short.

## Fix

Changed `prover_hint_ark` to `prover_message` for `public_eval` in both prover and verifier so the value is absorbed into the Fiat-Shamir transcript.

`make_public_weight` / `compute_public_eval` now internally add `+1` to account for the R1CS constant at position 0. Parameter renamed to `num_public_inputs` to clarify semantics.

Added `verify_public_input_binding()` in the verifier that recomputes the expected evaluation from known public inputs and rejects if it doesn't match the prover's claim.

Added `test_public_input_binding_exploit` that proves honestly, then tampers `public_inputs` to a wrong value and asserts the verifier rejects.

### DIFF
--- a/provekit/verifier/src/whir_r1cs.rs
+++ b/provekit/verifier/src/whir_r1cs.rs
@@ -290,6 +290,9 @@ fn verify_public_input_binding(
         expected += x_pow * pi;
         x_pow *= x;
     }
-    ensure!(public_eval == expected, "Public input binding check failed");
+    ensure!(
+        public_eval == expected,
+        "Public input binding check failed: expected {expected:?}, got {public_eval:?}"
+    );
     Ok(())
 }


provekit/verifier/src/whir_r1cs.rs
```diff
@@ -273,3 +275,21 @@ pub fn run_sumcheck_verifier(
...
+    x: FieldElement,
+    public_inputs: &PublicInputs,
+) -> Result<()> {
+    let mut expected = FieldElement::one();
+    let mut x_pow = x;
+    for &pi in &public_inputs.0 {
+        expected += x_pow * pi;
+        x_pow *= x;
+    }
+    ensure!(public_eval == expected, "Public input binding check failed");
```
Bisht13: Suggest changing to:
```suggestion
ensure!(
      public_eval == expected,
      "Public input binding check failed: expected {expected:?}, got {public_eval:?}"
  );
```

PR #336, opened 2026-03-16
fix: handling public dead inputs

### LINKED ISSUE #335: oprf circuit prepare error
When running the OPRF circuit through `provekit-cli prepare`, compilation fails:

```
Error: while compiling Noir program

Caused by:
    no builder for public input ACIR index 10
```

ashpect:
fixes #335 

error : When running the OPRF circuit through `provekit-cli prepare`, compilation fails:

```
Error: while compiling Noir program

Caused by:
    no builder for public input ACIR index 10
```

root cause : Noir's ACIR `Circuit` struct stores public inputs in `public_parameters: PublicInputs` and constraints in `opcodes: Vec<Opcode>`. These are independent — a witness can be in `public_parameters` without appearing in any `Opcode` (e.g. inputs used only in Brillig). ProveKit's R1CS compiler walks `opcodes` to create witness builders but never reads `public_parameters`, so it misses any public input that no opcode references, hence never finding that public input builder later. This happens on unconstrained public inputs.

fix : After processing all opcodes in `add_circuit_with_breakdown`, call f`etch_r1cs_witness_index` for every witness in `circuit.public_inputs()`. This is idempotent, already-seen witnesses hit the map and return, unseen ones get a new `WitnessBuilder::Acir` entry.

### DIFF
--- a/provekit/r1cs-compiler/src/noir_to_r1cs.rs
+++ b/provekit/r1cs-compiler/src/noir_to_r1cs.rs
@@ -736,3 +736,37 @@ impl NoirToR1CSCompiler {
         Ok(breakdown)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use {
+        super::*,
+        acir::circuit::PublicInputs as AcirPublicInputs,
+        provekit_common::witness::WitnessBuilder,
+        std::collections::{BTreeSet, HashSet},
+    };
+
+    /// Regression: public inputs absent from all opcodes must still get
+    /// builders (case for unconstrained public inputs), otherwise
+    /// split_and_prepare_layers fails with NoBuilderForPublicInput.
+    #[test]
+    fn dead_public_input_compiles() {
+        let circuit = Circuit {
+            public_parameters: AcirPublicInputs(BTreeSet::from([NoirWitness(1)])),
+            opcodes: vec![],
+            ..Default::default()
+        };
+        let (r1cs, witness_map, witness_builders) =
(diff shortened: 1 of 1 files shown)


provekit/r1cs-compiler/src/noir_to_r1cs.rs
```diff
@@ -725,6 +725,14 @@ impl NoirToR1CSCompiler {
         breakdown.range_constraints = self.r1cs.num_constraints() - constraints_before_range;
         breakdown.range_witnesses = self.num_witnesses() - witnesses_before_range;
 
+        // Noir's Circuit.opcodes may not reference every witness declared in
+        // Circuit.public_parameters (e.g. public inputs declared but not constrained).
+        // Create builders for any public inputs that the opcode loop above
+        // missed.
+        for acir_idx in circuit.public_inputs().indices() {
```
Bisht13, line 732: please add a unit test that constructs a Circuit with a public input witness absent from all opcodes and asserts noir_to_r1cs() succeeds. Something like:
```rs
#[test]
fn dead_public_input_compiles() {
    // Circuit with public_parameters containing witness index 1,
    // but no Opcode references witness 1.
    let circuit = Circuit {
        public_parameters: PublicInputs(BTreeSet::from([Witness(1)])),
        opcodes: vec![], // or opcodes referencing only other witnesses
        ..Default::default()
    };
    noir_to_r1cs(&circuit).expect("dead public input should compile");
}
```

ashpect: added a suitable test

PR #345, opened 2026-03-18
Adds SPARK

batmendbar:
# SPARK

Reference for this implementation
- SPARK: https://eprint.iacr.org/2019/550
- Stronger security analysis of SPARK: https://people.cs.georgetown.edu/jthaler/Lasso-paper.pdf

## Proposed protototype workflow
1. Serve step 
    - One time
      - Starts the server
      - Compiles the circuit
      - Calculates the SPARK matrix data and commits to them
    - Ongoing
      - Listens to SPARK query requests and produces SPARK proofs using the pre-calculated commitments

2. Provekit prove step
    - Runs provekit prover and obtains a deferred evaluation
    - Sends a deferred evaluation request to the server

3. Provekit and SPARK verify step
    - Verifies Provekit and SPARK proofs

### Pack $A$, $B$, $C$ into one block matrix Z:
This is a result from Marcin (https://gist.github.com/kustosz/14b62de666f721ab855536e575891bd1)

**The trick:** 

$$Z = \begin{bmatrix} A & B \\ 0 & C \end{bmatrix}$$

Same total non-zeros, double the dimensions. Then for any $\beta$, $p$, and $q$:

$$A(p,q) + \beta B(p,q) + \beta^2 C(p,q) = (1+\beta)^2 \cdot Z\!\left(\tfrac{\beta}{1+\beta}, p,\ \tfrac{\beta}{1+\beta}, q\right)$$

One matrix, one commitment, one opening.

### Batching GPA and WHIR proofs

- Combining GPA
  - Products of hashes corresponding to read sets and write sets of row-wise and column-wise memory check are combined into one GPA
  - Products of hashes corresponding to init and final vectors are combined into one GPA (separate for row-wise and col-wise memory). Possible optimization - if number of rows and columns for the matrix are ensured to be equal, we can combine them into one GPA.

- WHIR Batching
| `num_terms_2batched` e-values are committed and opened together. Opened once in sumcheck and once in rs_ws GPA
| `num_terms_4batched` | Address/timestamp values for row-wise and col-wise memory checks are committed and opened together

### Temporary Sumcheck for split witness
The current ZK WHIR doesn't support batching which would enable easier handling of split witness commitment. The repo currently uses an additional sumcheck proposed by Marcin until batch ZK commitment is supported https://gist.github.com/kustosz/c7c3f756aaae77f37e035c30c4961ea3.

**The  trick:**

**Collapsing two claims into one:** With claims on $A(r,0,q_1)$ and $A(r,1,q_2)$. Draw random $\beta$ and note their RLC is just another sum over $A(r,\cdot)$, so run a **Sumcheck #2** to reduce to a single claim $A(r, \gamma)$.

## Full workflow for a Noir passport circuit:

```bash
# 1. Start the server (compiles circuits and pre-commits)
provekit-cli serve \
  --socket /tmp/spark.sock \
  --output-dir ./benchmark-inputs \
  --circuit t_attest:./target/t_attest.json &

# 2. Wait for server readiness
while [ ! -S /tmp/spark.sock ]; do sleep 1; done

# 3. Prove (generates Noir proof + SPARK proof)
provekit-cli prove \
  ./benchmark-inputs/t_attest.pkp \
  ./benchmark-inputs/tbs_720/t_attest.toml \
  -o ./benchmark-inputs/t_attest-proof.np \
  --socket /tmp/spark.sock \
  --circuit t_attest \
  --spark-out ./benchmark-inputs/t_attest-spark-proof.sp

# 4. Verify Noir + SPARK proof
provekit-cli verify \
  ./benchmark-inputs/t_attest.pkv \
  ./benchmark-inputs/t_attest-proof.np \
  --spark-proof ./benchmark-inputs/t_attest-spark-proof.sp
```
 
 # What is not included
 - Recursive SPARK 
 
# Benchmark
[TODO: Update with new benchmark]


provekit/verifier/src/whir_r1cs.rs
```diff
@@ -201,6 +201,9 @@ impl WhirR1CSVerifier for WhirR1CSScheme {
                 .verify(&mut arthur, &weight_refs_2, &evaluations_2, &commitment_2)
                 .map_err(|_| anyhow::anyhow!("WHIR verification failed for c2"))?;
 
+            // Temporary fix! Drain SPARK .
```
veljkovranic: is this supposed to stay like this?

tooling/cli/src/cmd/serve.rs
```diff
@@ -0,0 +1,240 @@
...
+    let num_nonzero = spark_data.matrix.coo.val.len();
+
+    info!("Proving ({num_constraints} constraints, {num_witnesses} witnesses)");
+    let scheme = SPARKProverScheme::new(num_constraints, num_witnesses, num_nonzero);
+    let proof = scheme
+        .prove(spark_data, &request.spark_query)
+        .context("generating SPARK proof")?;
+
+    info!("Writing proof to {:?}", request.output);
+    write(&proof, &request.output).context("writing spark proof")?;
```
veljkovranic: this seems risque, we shouldn't let them choose the path

provekit/spark/src/verifier.rs
```diff
@@ -0,0 +1,244 @@
...
+        let whir_proof = Proof {
+            narg_string: proof.narg_string,
+            hints: proof.hints,
+            #[cfg(debug_assertions)]
+            pattern: proof.pattern,
+        };
+        let mut arthur = VerifierState::new(&ds, &whir_proof, TranscriptSponge::default());
+
+        let claimed_value = (request.claimed_value
+            / (FieldElement::ONE + request.matrix_batching_randomness))
```
veljkovranic: can these numbers be zero?

batmendbar: Yes, but with extremely small probability. I can add a panic here if it does.

tooling/cli/src/cmd/serve.rs
batmendbar: The spark proof path?

provekit/verifier/src/whir_r1cs.rs
batmendbar: The problem here is that the current WHIR `verify` function doesn't expose its internal folding randomness which we need to run the split witness sumcheck. For debugging purposes, we can extract these folding randomnesses from the Prover, but that would be unsound, and I think would confuse others reading the code. Furthermore, this split witness is an only temporary solution until the ZK WHIR supports batch proving.

tooling/cli/src/cmd/serve.rs
veljkovranic: Yes, we should never let the client drive the server behavior. Imagine if I send '~/.ssh/authorized_keys' as the path I want my spark proof to be written in

provekit/verifier/src/whir_r1cs.rs
```diff
@@ -262,6 +265,32 @@ impl WhirR1CSVerifier for WhirR1CSScheme {
     }
 }
 
+/// TEMPORARY FIX: Drain the split witness for now. To properly check, we
+fn drain_split_witness_sumcheck(
```
Bisht13: This is transcript draining, not verification. We accept the split-witness path without checking the extra sumcheck relation, which makes this composition unsound.

tooling/cli/src/cmd/verify.rs
```diff
@@ -32,11 +37,26 @@ impl Command for Args {
...
         verifier
             .verify(&proof)
             .context("While verifying Noir proof")?;
 
+        // Verify the SPARK proof if provided
+        if let Some(spark_proof_path) = &self.spark_proof_path {
+            info!("Verifying SPARK proof from {spark_proof_path:?}");
+            let spark_proof: SPARKProof =
+                read(spark_proof_path).context("while reading SPARK proof")?;
+            let spark_statement = proof.r1cs_spark_query;
```
Bisht13: r1cs_spark_query is taken as plaintext from NoirProof, but the Noir verifier does not derive or authenticate it. That means a prover can swap in a different SPARK statement. We need cryptographic binding between the Noir proof and this query.

provekit/spark/src/verifier.rs
Bisht13: Can you add an `ensure!` on both prover and verifier paths?

tooling/cli/src/cmd/verify.rs
```diff
@@ -32,11 +37,26 @@ impl Command for Args {
...
+            let scheme = SPARKVerifierScheme::from_proof(&spark_proof);
```
Bisht13: The verifier scheme is reconstructed from untrusted proof data (whir_params, matrix_dimensions). That gives the prover control over verification parameters. These need to come from trusted verifier-side data or be cross-checked against it.

provekit/spark/src/verifier.rs
```diff
@@ -0,0 +1,244 @@
...
+            whir_configs:      proof.whir_params.clone(),
+            matrix_dimensions: proof.matrix_dimensions.clone(),
+        }
+    }
+}
+
+impl SPARKVerifier for SPARKScheme {
+    #[instrument(skip_all)]
+    fn verify(&self, proof: SPARKProof, request: &R1CSSparkQuery) -> Result<()> {
+        let ds = DomainSeparator::protocol(&self.whir_configs).instance(&Empty);
```
Bisht13: The SPARK transcript binds only protocol parameters, not circuit/R1CS identity. Two circuits with matching dimensions can share the same domain separator. We should bind circuit identity here (for example R1CS hash or equivalent trusted instance data).

tooling/cli/src/cmd/serve.rs
Bisht13: Agree with @veljkovranic

tooling/cli/src/cmd/spark_protocol.rs
```diff
@@ -0,0 +1,48 @@
...
+}
+
+pub fn read_message<T: DeserializeOwned>(stream: &mut impl Read) -> Result<T> {
+    let mut len_buf = [0u8; 4];
+    stream
+        .read_exact(&mut len_buf)
+        .context("reading message length")?;
+    let len = u32::from_le_bytes(len_buf) as usize;
+
+    let mut buf = vec![0u8; len];
```
Bisht13: The length prefix is fully trusted here, so a client can request an arbitrarily large allocation. Please cap message size before allocation and reject oversized requests.

provekit/common/src/noir_proof_scheme.rs
```diff
@@ -29,8 +30,9 @@ pub enum NoirProofScheme {
 
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub struct NoirProof {
-    pub public_inputs:   PublicInputs,
-    pub whir_r1cs_proof: WhirR1CSProof,
+    pub public_inputs:    PublicInputs,
+    pub whir_r1cs_proof:  WhirR1CSProof,
+    pub r1cs_spark_query: R1CSSparkQuery,
```
Bisht13: This changes the serialized NoirProof shape and makes SPARK query mandatory for every proof. Even with the version bump, this is a compatibility break with no migration/backward-compatible path. Please either make this optional with compatibility handling or document and coordinate the format break explicitly.

Bisht13 (changes requested):

provekit/verifier/src/whir_r1cs.rs
batmendbar: Removed SPARK verification from native verifier since the whir verifier checks the linear weight evaluation internally.

provekit/verifier/src/whir_r1cs.rs
batmendbar: Removed SPARK verification from native verifier since the whir verifier checks the linear weight evaluation internally.

provekit/common/src/noir_proof_scheme.rs
batmendbar: Separated spark query from noir proofs

tooling/cli/src/cmd/serve.rs
batmendbar: Disabled user specified path

tooling/cli/src/cmd/verify.rs
batmendbar: Removed native spark verification since direct MLE of linear weights are preferred

tooling/cli/src/cmd/spark_protocol.rs
batmendbar: Limited the length

provekit/spark/src/verifier.rs
batmendbar: Added ensure

tooling/cli/src/cmd/verify.rs
batmendbar: Removed the native verifier

provekit/spark/src/verifier.rs
batmendbar: The SPARK transcript now binds to the r1cs commitment and each requests.

tooling/cli/src/cmd/verify.rs
Bisht13: The combined Noir+SPARK binding is still not established: verify does not derive/output the expected query, and verify-spark accepts caller-supplied setup/query sidecars. For the recursive/combined verifier, the query should be recomputed or bound from the Noir verifier transcript, not trusted from JSON.

batmendbar: The current native verifier doesn't derive the expected query because ZKWHIR internally verifies the FinalClaim using mle_evaluate. On the other hand, the recursive verifier will extract the query from running  spartan-sumcheck and whir. Then it will verify the spark proof against this query. The current spark-verifier is a sandalone verifier where the query is a public input.

.github/workflows/end-to-end.yml.disabled
```diff
@@ -57,48 +57,65 @@ jobs:
...
-      - name: Verify proofs for all circuits
+      - name: Generate SPARK proofs for all circuits
         working-directory: noir-examples/noir-passport/merkle_age_check
         run: |
           for circuit in t_add_dsc_720 t_add_id_data_720 t_add_integrity_commit t_attest; do
-            echo "Verifying $circuit"
-            cargo run --release --bin provekit-cli verify \
-              ./benchmark-inputs/$circuit-verifier.pkv \
-              ./benchmark-inputs/$circuit-proof.np
-            echo "Verified $circuit"
```
veljkovranic, line 94: We should keep verifing provekits proof, SPARK verifier doesn't call it internally right?

provekit/spark/src/prover.rs
```diff
@@ -0,0 +1,622 @@
...
+            &DomainSeparator::protocol(&self.whir_configs)
+                .session(&spark_data.setup.transcript.narg_string)
+                .instance(&hash_query_set(requests)),
+            TranscriptSponge::default(),
+        );
+
+        let request = if requests.len() == 1 {
+            requests[0].clone()
+        } else {
+            let alphas = alphas_from_spark(
```
Bisht13, line 134: For multi-query batching we compute alphas from requests[0].row, but never ensure the rest of the requests use the same row. Please reject mixed-row batches before proving.

provekit/spark/src/verifier.rs
```diff
@@ -0,0 +1,325 @@
...
+                "parallel sumcheck final-equation check failed (b)"
+            );
+            ensure!(
+                final_claims[2] == h_at_fr * folded[2],
+                "parallel sumcheck final-equation check failed (c)"
+            );
+
+            R1CSSparkQuery {
+                point_to_evaluate: Point {
+                    row: requests[0].point_to_evaluate.row.clone(),
```
Bisht13: Same-row batching needs to be enforced verifier-side too. Otherwise verification can fold claims using requests[0].row even when later requests target different rows.

provekit/spark/src/setup.rs
```diff
@@ -0,0 +1,105 @@
...
+    matrix: &SparkMatrix,
+    hash_config: HashConfig,
+) -> (SPARKSetup, SparkWitnesses) {
+    let num_rows = matrix.timestamps.final_row.len();
+    let num_cols = matrix.timestamps.final_col.len();
+    let nonzero_terms = matrix.coo.val.len();
+    let scheme = SPARKProverScheme::new(num_rows, num_cols, nonzero_terms, hash_config);
+
+    let ds = DomainSeparator::protocol(&scheme.whir_configs).instance(&Empty);
+    let mut merlin = ProverState::new(&ds, TranscriptSponge::default());
```
Bisht13: This uses TranscriptSponge::default() even though setup receives hash_config. SPARK should use the configured Fiat-Shamir hash consistently, or store the config in SPARKSetup.

provekit/spark/src/verifier.rs
```diff
@@ -0,0 +1,325 @@
...
+            hints: proof.0.hints,
+            #[cfg(debug_assertions)]
+            pattern: proof.0.pattern,
+        };
+        let mut arthur = VerifierState::new(
+            &DomainSeparator::protocol(&setup.whir_configs)
+                .session(&setup.transcript.narg_string)
+                .instance(&hash_query_set(requests)),
+            &whir_proof,
+            TranscriptSponge::default(),
```
Bisht13: Verifier also uses the default transcript sponge. Please derive it from trusted setup/hash config so --hash does not affect commitments but not Fiat-Shamir.

tooling/cli/src/cmd/verify_spark.rs
```diff
@@ -0,0 +1,67 @@
...
+                            .collect::<Result<Vec<_>>>()
+                    },
+                )
+            },
+        );
+        let proof = proof?;
+        let setup = setup?;
+        let queries = queries?;
+
+        SPARKVerifierScheme
```
Bisht13: verify and verify-spark share no cryptographic state, a prover can pair an honest NoirProof for C₁ with a self-consistent .sp+.spc+query JSON set for a different circuit C₂ with matching dimensions, and both verifiers will accept. Derive the expected query from the Noir transcript and require verify-spark to consume it, ship a combined verifier.

tooling/cli/src/cmd/verify_spark.rs
```diff
@@ -0,0 +1,67 @@
...
+        anyhow::ensure!(
+            !self.query_paths.is_empty(),
+            "verify-spark needs at least one query path"
+        );
+
+        let (proof, (setup, queries)) = rayon::join(
+            || read::<SPARKProof>(&self.proof_path).context("while reading SPARK proof"),
+            || {
+                rayon::join(
+                    || read::<SPARKSetup>(&self.setup_path).context("while reading SPARK setup"),
```
Bisht13: SPARKSetup is loaded from a caller-supplied path with no trust link to .pkv. The verifier replays the transcript against itself but never compares to verifier-trusted state. Whoever controls .spc controls the commitments. Hash .spc into .pkv at prepare time and check at verify-spark start, or fold setup commitments into .pkv directly.

tooling/cli/src/cmd/prove_spark.rs
```diff
@@ -0,0 +1,119 @@
...
+
+        let prover: Prover = read(&self.prover_path).context("while reading Provekit Prover")?;
+
+        let queries = collect_queries(&self.spark_dir)?;
+        if queries.is_empty() {
+            info!("No SPARK queries found in {:?}", self.spark_dir);
+            return Ok(());
+        }
+
+        // TODO: preprocessing should be done once on the setup side, not per prove
```
Bisht13: This breaks SPARK's "commit once, prove many" promise, prepare already wrote .spc. The blocker is that preprocess_spark returns both SPARKSetup (serializable) and SparkWitnesses (not). Split it: preprocess_setup(matrix) -> SPARKSetup and recompute_witnesses(matrix, &setup) -> SparkWitnesses. prove-spark reads .spc and recomputes only the witnesses.

provekit/prover/src/lib.rs
```diff
@@ -40,14 +40,26 @@ pub use {ec_arith::ec_scalar_mul, r1cs::solve_witness_vec};
...
+///
+/// All methods return the `NoirProof` plus a `Vec<R1CSSparkQuery>` of SPARK
+/// queries produced as a side output. Callers that don't need the queries
+/// can discard with `let (proof, _) = ...`. SPARK query generation is
+/// off by default; enable it by calling
+/// [`Prover::set_produce_spark_query`](provekit_common::Prover::set_produce_spark_query).
 pub trait Prove {
     #[cfg(all(feature = "witness-generation", not(target_arch = "wasm32")))]
-    fn prove(self, input_map: InputMap) -> Result<NoirProof>;
+    fn prove(self, input_map: InputMap) -> Result<(NoirProof, Vec<R1CSSparkQuery>)>;
```
Bisht13: Prove now returns (NoirProof, Vec<R1CSSparkQuery>) for every caller, including FFI/wasm. Smaller break than the wire-format one, but still uncoordinated. Add prove_with_spark(...) for the tuple and leave prove(...) returning NoirProof; or wrap as pub struct ProveOutput { proof, spark_queries } so future fields are non-breaking.

provekit/common/src/prover.rs
```diff
@@ -18,6 +18,8 @@ pub struct NoirProver {
     pub split_witness_builders: SplitWitnessBuilders,
     pub witness_generator:      NoirWitnessGenerator,
     pub whir_for_witness:       WhirR1CSScheme,
+    #[serde(skip, default)]
```
Bisht13: Runtime config flag on a persisted .pkp schema, #[serde(skip, default)] masks the smell rather than fixing it. Tests/FFI silently get the default; every Prover constructor has to remember the field exists. Move to ProveOpts passed into prove(...); drop set_produce_spark_query.

provekit/spark/src/lib.rs
```diff
@@ -0,0 +1,19 @@
+pub mod gpa;
+pub mod memory;
+pub mod prover;
+pub mod setup;
+pub mod sumcheck;
+pub mod types;
+pub mod utils;
+pub mod verifier;
+
+pub use {
```
Bisht13, line 11: Two structs named SPARKScheme in sibling modules force the renames SPARKProverScheme/SPARKVerifierScheme at the crate root. The SPARKProver/SPARKVerifier traits each have one impl and one method. Drop the traits and rename the structs to Prover/Verifier (or ProverScheme/VerifierScheme).

provekit/spark/src/lib.rs
```diff
@@ -0,0 +1,19 @@
+pub mod gpa;
```
Bisht13: Every submodule is pub, so gpa::run_gpa4, sumcheck::run_spark_sumcheck, utils::calculate_adr etc. become public API. Make them pub(crate) and keep only the curated pub use façade, once external code starts importing internals, signatures freeze.

provekit/common/src/spark.rs
```diff
@@ -0,0 +1,35 @@
...
+}
+
+impl R1CSSparkQuery {
+    pub fn hash_bytes(&self) -> [u8; 32] {
+        hash_query_set(std::slice::from_ref(self))
+    }
+}
+
+pub fn hash_query_set(queries: &[R1CSSparkQuery]) -> [u8; 32] {
+    let bytes = postcard::to_allocvec(queries).expect("serializing R1CSSparkQuery set");
```
Bisht13: We shouldn't be using .expect, Postcard failure here panics both prover and verifier. Return Result<[u8; 32]>.

provekit/spark/src/verifier.rs
```diff
@@ -0,0 +1,325 @@
...
+
+    let (randomness, last_sumcheck_value) = run_sumcheck_verifier_spark(
+        arthur,
+        next_power_of_two(matrix_dimensions.nonzero_terms),
+        *claimed_value,
+    )
+    .context("While verifying SPARK sumcheck")?;
+    let eval_weight = MultilinearExtension::new(randomness);
+
+    let sumcheck_hints: [FieldElement; 3] = arthur
```
Bisht13: 14+ copies of .map_err(|_| anyhow::anyhow!("Failed to read X hint"))? Also discards the underlying WHIR error with |_|, making debugging hard. Helper fn read_hint<T>(arthur, label) -> Result<T> preserving {e}.

provekit/spark/src/gpa.rs
```diff
@@ -0,0 +1,431 @@
+use {
```
Bisht13, line 1: run_gpa2/run_gpa4 and gpa_sumcheck_verifier2/gpa_sumcheck_verifier4 are ~400 LOC of near-duplication, they differ only in base-layer shape. Parameterize over BaseLayer::{Line, Rect}.

provekit/spark/src/types.rs
```diff
@@ -0,0 +1,123 @@
...
+}
+
+#[derive(Debug, Clone)]
+pub struct SparkMatrix {
+    pub coo:        COOMatrix,
+    pub timestamps: TimeStamps,
+}
+
+#[derive(Debug, Clone)]
+pub struct COOMatrix {
```
Bisht13, line 77: COOMatrix stores row: Vec<usize> and row_field: Vec<FieldElement> (same data twice). Doubles memory for big circuits and the invariant row_field[i] == FieldElement::from(row[i] as u64) is unenforced. Drop one and derive on demand, or wrap construction in a constructor that enforces it.

tooling/cli/src/cmd/prove_spark.rs
```diff
@@ -0,0 +1,119 @@
...
+        let proof_path = self.spark_dir.join("spark_proof.sp");
+        write(&spark_proof, &proof_path)
+            .with_context(|| format!("writing SPARK proof to {proof_path:?}"))?;
+        info!("Wrote SPARK proof to {proof_path:?}");
+
+        Ok(())
+    }
+}
+
+fn collect_queries(dir: &Path) -> Result<Vec<R1CSSparkQuery>> {
```
Bisht13: collect_queries walks spark_query_<i>.json until first miss, if _3.json is missing but _4.json exists, _4 is silently skipped and the SPARK proof binds the wrong set. Use read_dir+sort, or an explicit manifest, and ensure! the discovered set is contiguous.

provekit/prover/src/whir_r1cs.rs
```diff
@@ -62,9 +63,11 @@ pub trait WhirR1CSProver {
         commitments: Vec<WhirR1CSCommitment>,
         full_witness: Vec<FieldElement>,
         public_inputs: &PublicInputs,
-    ) -> Result<WhirR1CSProof>;
+        produce_spark_query: bool,
+    ) -> Result<(WhirR1CSProof, Vec<R1CSSparkQuery>)>;
```
Bisht13: Here and :159, :269, #[allow(clippy::too_many_arguments)] mutes a workspace-level lint. Group into a ProveFromAlphasCtx, SparkProverContext already sets the precedent.

provekit/spark/src/prover.rs
```diff
@@ -0,0 +1,622 @@
...
+    tracing::instrument,
+    whir::{
+        algebra::{linear_form::MultilinearExtension, multilinear_extend},
+        engines::EngineId,
+        parameters::ProtocolParameters,
+        transcript::{DomainSeparator, ProverState, VerifierMessage},
+    },
+};
+
+pub trait SPARKProver {
```
Bisht13: SPARK* acronym casing violates Rust API Guidelines C-CASE and trips clippy::upper_case_acronyms. The codebase already uses Spark* for SparkMatrix/SparkWitnesses/SparkProverContext, internally inconsistent. Rename to Spark*.

provekit/spark/src/prover.rs
```diff
@@ -0,0 +1,622 @@
...
+            .in_scope(|| {
+                polys
+                    .par_iter()
+                    .map(|p| multilinear_extend(p, evaluation_randomness))
+                    .collect::<Vec<_>>()
+            })
+            .try_into()
+            .expect("4 polys");
+
+    merlin.prover_hint_ark(&row_address_eval);
```
Bisht13, line 411: One-line comment per hint call naming which WHIR check seals it would prevent a future refactor from breaking soundness.

Bisht13 (changes requested):

tooling/cli/src/cmd/verify_spark.rs
batmendbar: The verify in Provekit currently doesn't even need SPARK since it directly evaluates A, B, C matrices on their own. The current ZK-WHIR api doesn't allow us to change this. (e

veljkovranic (approved):

PR #349, opened 2026-03-18
bump Noir dependency to v1.0.0-beta.19

x-senpai-x:
## Summary

- Bumps all Noir language dependencies from `v1.0.0-beta.11` to `v1.0.0-beta.19`
- Adapts ProveKit's ACIR usage to the breaking API changes in beta.19
- Includes two separate compilation fixes 
- Align WASM/demo Noir JS stack with beta.19

## ACIR API changes (beta.11 → beta.19)

| Change | Old | New |
|--------|-----|-----|
| Input enum | `ConstantOrWitnessEnum` | `FunctionInput` |
| Value access | `.input()` method | direct dereference `*` |
| Bit width | `.num_bits()` on `FunctionInput` | separate `num_bits` field on AND/XOR/RANGE opcodes |
| `MemoryOp` | `predicate` field (always `None`) | field removed |
| `Poseidon2Permutation` | `len` field | removed; use `inputs.len()` |
| BigInt blackboxes | 6 variants | all removed from ACIR |

## Minor fixes

**`tracy` feature implies `profiling-allocator`** (`tooling/cli/Cargo.toml`): `ALLOCATOR.enable_tracy()` is gated on `profiling-allocator` but the `tracy` feature didn't enable it, causing a compile error with `--features tracy --no-default-features`.

**`passport-input-gen` witness-generation** (`playground/passport-input-gen/Cargo.toml`): workspace sets `provekit-prover` with `default-features = false`, so `witness-generation` was silently disabled. Now explicitly enabled.

## Dependency pin

Added `keccak = "=0.2.0-rc.2"` to workspace deps (anchored via `provekit-common`). `acvm_blackbox_solver` (beta.19) calls `keccak::f1600()` which was removed in the keccak `0.2.0` stable release. Without this pin, bare `cargo update` bumps the RC to stable and breaks the build.


tooling/provekit-ffi/tests/ffi_integration.rs
```diff
@@ -0,0 +1,1006 @@
...
+        unsafe { pk_free_verifier(self.0) };
+    }
+}
+
+struct ScopedBuf(PKBuf);
+impl Drop for ScopedBuf {
+    fn drop(&mut self) {
+        // pk_free_buf takes ownership by value; use ptr::read to avoid moving
+        // out of a field while the struct is being dropped.
+        unsafe {
```
ashpect: ptr::read copies the PKBuf struct value but doesn't zero the original. If ScopedBuf were ever double-dropped (e.g., panic during drop), the same would be freed twice.
 Since pk_free_buf takes PKBuf by value (not by pointer), a safer approach would be to std::mem::replace the fields to zero/null before passing:

tooling/provekit-ffi/tests/ffi_integration.rs
```diff
@@ -0,0 +1,1006 @@
...
+// K. Thread safety (release only)
+// ===========================================================================
+
+#[test]
+#[cfg_attr(debug_assertions, ignore)]
+fn k_concurrent_prove_same_handle() {
+    use std::thread;
+
+    let (pk, vk) = unsafe { prepare_basic_circuit() };
+    // Share the raw pointer across threads — PKProver is Send+Sync
```
ashpect: wrap the pointer in a newtype that implements Send + Sync (since PKProver is asserted Send + Sync):

  struct SendPtr(*const PKProver);
  unsafe impl Send for SendPtr {}
  unsafe impl Sync for SendPtr {}

tooling/provekit-ffi/tests/ffi_integration.rs
```diff
@@ -0,0 +1,1006 @@
...
+    assert_eq!(status, PK_INVALID_INPUT);
+}
+
+#[test]
+fn f_verify_null_proof_ptr_returns_invalid_input() {
+    init();
+    // We need a non-null verifier for this check. Allocate a junk pointer
+    // that passes the null check — but verify checks proof_ptr null before
+    // dereferencing verifier, so this is safe.
+    let status = unsafe { pk_verify(std::ptr::null(), std::ptr::null(), 16) };
```
ashpect: both verifier and proof_ptr are passed as null. mention in comment. 
Since
  pk_verify checks verifier.is_null() || proof_ptr.is_null() || proof_len == 0 in a single OR chain, this test actually triggers
  the verifier.is_null() branch, not the proof_ptr.is_null() branch. It passes by coincidence. To actually test the null-proof-ptr
  path:

  let (_pk, vk) = unsafe { prepare_basic_circuit() };
  let status = unsafe { pk_verify(vk.0, std::ptr::null(), 16) };
  assert_eq!(status, PK_INVALID_INPUT);

tooling/provekit-ffi/tests/ffi_integration.rs
```diff
@@ -0,0 +1,1006 @@
...
+    // Error should be set
+    let err1 = unsafe { last_error() };
+    assert!(
+        err1.as_slice().len() > 0,
+        "error should be set after failing call"
+    );
+
+    // Now a succeeding call (pk_init clears last error via catch_panic's
+    // clear_last_error)
+    let _ = pk_init();
```
ashpect: There's no clear_last_error() call here. The test passes because last_error() already consumed the error on line 838 (pk_get_last_error takes + clears). So the test is really just verifying "reading the error clears it", not "successful calls clear the error". The test name is misleading.

ashpect: Also, can you open a issue for provekit-ffi, so that i can take a look later :
 ```pk_init() doesn't call catch_panic(), so it neither clears errors nor catches panics. If register_ntt() panics, it unwinds across the FFI boundary to undefined behavior. Every other FFI function wraps in catch_panic, but pk_init doesn't.```

ashpect (commented):
Can you also add tests for pk_configure_memory, pk_prove_json with malformedjson, pk_save_prover / pk_save_verifier with null path, pk_load_verifier_bytes with corrupt data ?

provekit/r1cs-compiler/src/noir_to_r1cs.rs
```diff
@@ -654,20 +620,136 @@ impl NoirToR1CSCompiler {
...
+                        // The full MSM constraint set is always generated — GLV decomposition,
+                        // signed-bit decomposition, on-curve checks, and the EC addition chain.
+                        // R1CS is a static constraint system: every row of A·B=C is fixed at
+                        // compile time and must be satisfied for every valid witness. Unlike
+                        // plonkish systems (e.g. Barretenberg/UltraPlonk), which have gate
+                        // selectors that multiply a gate's contribution by 0 when inactive, R1CS
+                        // has no such mechanism. This is an inherent cost of R1CS for runtime
+                        // conditional operations.
+                        let predicate = self.fetch_constant_or_r1cs_witness(*predicate);
+                        let mut skip_msm = false;
```
ocdbytes: This adds the optimisation to ignore the branch when predicate is a 0 determined by noir compiler.

NIT : 

we are initialising the flag here but I would say we can just use `continue` on the loop something like this : 

```rust
  ConstantOrR1CSWitness::Constant(c) if c.is_zero() => {
      let out_x = self.fetch_r1cs_witness_index(outputs.0);
      let out_y = self.fetch_r1cs_witness_index(outputs.1);
      let out_inf = self.fetch_r1cs_witness_index(outputs.2);
      constrain_to_constant(self, out_x, FieldElement::zero());
      constrain_to_constant(self, out_y, FieldElement::zero());
      constrain_to_constant(self, out_inf, FieldElement::one());
      continue; 
  }
```

this looks cleaner according to my opinion. Let me know wyt

noir-examples/msm_conditional/src/main.nr
```diff
@@ -0,0 +1,28 @@
...
+///
+/// When `condition = true`, the MSM runs and the result must be non-infinite.
+/// When `condition = false`, the MSM is in an inactive branch (predicate=0 in
+/// ACIR), and ACVM outputs the identity point (0, 0, true). This exercises
+/// provekit's conditional MSM support.
+fn main(condition: bool, scalar_lo: Field, scalar_hi: Field, expected_infinite: bool) {
+    let scalar = EmbeddedCurveScalar { lo: scalar_lo, hi: scalar_hi };
+
+    // The if/else forces ACIR to emit MultiScalarMul with a Witness predicate.
+    let result: EmbeddedCurvePoint = if condition {
```
Bisht13, line 20: This only checks is_infinite, so it doesn’t fully lock the new predicate behavior. Since this example is meant to cover conditional MSM, please also assert the inactive-path output is exactly the identity point (0, 0, true) and add at least one active-path assertion on exact coordinates.

noir-examples/msm_conditional_nested/src/main.nr
```diff
@@ -0,0 +1,27 @@
...
+/// Exercises MSM with a compound predicate (outer AND inner).
+///
+/// When both `outer` and `inner` are true, the MSM runs and the result
+/// must be non-infinite. When either is false, the MSM is in an inactive
+/// branch and the circuit is still satisfiable (no assertion fires).
+///
+/// This exercises provekit's conditional MSM support with a Witness
+/// predicate that is the product of two boolean witnesses.
+fn main(outer: bool, inner: bool, scalar_lo: Field, scalar_hi: Field) {
+    if outer {
```
Bisht13: This covers satisfiability for nested predicates, but it still doesn’t assert what the false-path output must be. I’d prefer this example to materialize the result in all branches and assert the inactive case is the identity point, otherwise nested-predicate output semantics remain under-tested.

noir-examples/p256_bigcurve/src/main.nr
```diff
@@ -1,35 +1,39 @@
...
+    let r_x = Secp256r1_Fq::from_be_bytes(r);
+    let r = Secp256r1_Fr::from_be_bytes(r);
+    let s = Secp256r1_Fr::from_be_bytes(s);
 
     let s_g = Secp256r1Scalar::from_bignum(message / s);
     let s_p = Secp256r1Scalar::from_bignum(r / s);
-    let r_point = Secp256r1::msm([gen, public], [s_g, s_p]);
+    let r_point = Secp256r1::evaluate_linear_expression([gen, public], [s_g, s_p], []);
 
     assert(r_point.x == r_x);
```
Bisht13, line 45: This only checks the x-coordinate. For a crypto regression test, I’d also assert !r_point.is_infinity and the expected y-coordinate (or full point equality if the API allows it), otherwise a broken point can still pass with the right x.

Bisht13 (approved):

Bisht13:
This PR adds valuable FFI integration coverage, but the default CI path still runs debug cargo test, so the #[cfg_attr(debug_assertions, ignore)] cases in ffi_integration.rs don’t run here. Please add a targeted release job for cargo test -p provekit-ffi --test ffi_integration --release so the new verify/concurrency round-trips are exercised in CI.

PR #352, opened 2026-03-20
fix: canonicalize duplicate witness terms in R1CS constraints [LA - C]

Bisht13:
## Summary
- Canonicalize linear combinations inside `R1CS::add_constraint` before inserting into sparse matrices, merging duplicate witness indices (summing coefficients) and dropping zeros
- Add `debug_assert` in `SparseMatrix::set` to catch future regressions in debug builds

Addresses audit issue C (Least Authority, March 2026): duplicate witness terms in R1CS constraints silently overwrite coefficients via `SparseMatrix::set`, underconstraining the circuit.

## Test plan
- [x] 5 new regression tests in `r1cs::tests`

### DIFF
--- a/provekit/common/src/sparse_matrix.rs
+++ b/provekit/common/src/sparse_matrix.rs
@@ -355,9 +355,8 @@ impl SparseMatrix {
 
         // Find the column
         match cols.binary_search(&(col as u32)) {
-            Ok(i) => {
-                // Column already exists
-                self.values[row_range][i] = value;
+            Ok(_) => {
+                unreachable!("Duplicate column {col} in row {row}");
             }
             Err(i) => {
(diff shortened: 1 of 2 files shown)


provekit/common/src/sparse_matrix.rs
```diff
@@ -356,7 +356,8 @@ impl SparseMatrix {
         // Find the column
         match cols.binary_search(&(col as u32)) {
             Ok(i) => {
-                // Column already exists
+                // Column already exists, overwrite.
+                debug_assert!(false, "Duplicate column {col} in row {row}");
```
rose2221: replace `debug_assert!(false, ...)` with `assert!` (or `unreachable!`)

PR #361, opened 2026-03-21
Fix : Unbound LogUp Transcript Challenges Let the Prover Choose Lookup Randomness [LA-G]

ocdbytes:
## Summary

  - Backport of the soundness fix for unbound LogUp transcript challenges to the v1 branch.
  - The verifier samples Fiat-Shamir challenges for LogUp lookups but discards them (`let _logup_challenges`), never verifying that the committed
  w2 polynomial contains them. A malicious prover could substitute arbitrary challenge values in w2 without detection, breaking the LogUp
  argument's soundness.
  - Adds challenge binding that mirrors the existing public input binding pattern: prover sends `challenge_eval` as a transcript-bound
  `prover_message`, verifier independently recomputes from sampled challenges and checks equality, and a `SparseCovector` weight is included in the
   WHIR verification for w2.

  ## Changes

  **Core fix (prover + verifier transcript consistency):**
  - **`provekit/prover/src/whir_r1cs.rs`** — Compute `challenge_eval = Σ xⁱ · w2[offset_i]` and send as `prover_message` (transcript-bound). Add
  `SparseCovector` challenge weight to w2's WHIR prove call.
  - **`provekit/verifier/src/whir_r1cs.rs`** — Retain sampled challenges (was discarded via `let _logup_challenges`). Read `challenge_eval` from
  transcript, verify against independently computed expectation via `verify_challenge_binding()`. Add matching challenge weight to w2 WHIR verify
  call.

  **Supporting infrastructure:**
  - **`provekit/common/src/prefix_covector.rs`** — Add `SparseCovector` (implements `LinearForm` for non-contiguous positions),
  `make_challenge_weight()`, and `compute_challenge_eval()`. Includes 6 unit tests.
  - **`provekit/common/src/whir_r1cs.rs`** — Add `challenge_offsets: Vec<usize>` field to `WhirR1CSScheme` so the verifier knows where challenges
  sit in w2.
  - **`provekit/common/src/witness/witness_builder.rs`** — `split_and_prepare_layers` now returns `Vec<usize>` challenge offsets (was `usize`
  count) extracted from remapped Challenge builders.
  - **`provekit/r1cs-compiler/src/whir_r1cs.rs`** — `WhirR1CSSchemeBuilder::new_for_r1cs` accepts `challenge_offsets`. Added `debug_assert_eq!`
  enforcing `num_challenges == challenge_offsets.len()`.
  - **`provekit/r1cs-compiler/src/noir_proof_scheme.rs`** — Noir path destructures offsets from `split_and_prepare_layers`; Mavros path generates
  `(0..challenges_size).collect()`.

### DIFF
--- a/provekit/common/src/prefix_covector.rs
+++ b/provekit/common/src/prefix_covector.rs
@@ -247,6 +247,11 @@ impl SparseCovector {
     /// Create a new `SparseCovector` from `(position, weight)` pairs and a
     /// domain size.
     pub fn new(entries: Vec<(usize, FieldElement)>, domain_size: usize) -> Self {
+        assert!(domain_size.is_power_of_two());
+        assert!(
+            entries.iter().all(|&(pos, _)| pos < domain_size),
+            "SparseCovector: all entry positions must be < domain_size ({domain_size})"
+        );
         Self {
             entries,
             domain_size,
--- a/provekit/r1cs-compiler/src/whir_r1cs.rs
+++ b/provekit/r1cs-compiler/src/whir_r1cs.rs
@@ -26,7 +26,7 @@ impl WhirR1CSSchemeBuilder for WhirR1CSScheme {
         challenge_offsets: Vec<usize>,
         has_public_inputs: bool,
     ) -> Self {
-        debug_assert_eq!(
+        assert_eq!(
             num_challenges,
             challenge_offsets.len(),
             "num_challenges ({num_challenges}) must equal challenge_offsets.len() ({})",


provekit/common/src/prefix_covector.rs
```diff
@@ -235,6 +235,95 @@ pub fn compute_public_eval(
...
+    entries:     Vec<(usize, FieldElement)>,
+    domain_size: usize,
+}
+
+impl SparseCovector {
+    /// Create a new `SparseCovector` from `(position, weight)` pairs and a
+    /// domain size.
+    pub fn new(entries: Vec<(usize, FieldElement)>, domain_size: usize) -> Self {
+        Self {
+            entries,
```
ashpect, line 256: we should do a small domain size validation check as the prefixvector and offsetvector has.

```
debug_assert!(domain_size.is_power_of_two());
debug_assert!(entries.iter().all(|&(pos, _)| pos < domain_size));
```

ashpect: *assert

provekit/r1cs-compiler/src/whir_r1cs.rs
```diff
@@ -22,8 +23,15 @@ impl WhirR1CSSchemeBuilder for WhirR1CSScheme {
         r1cs: &R1CS,
         w1_size: usize,
         num_challenges: usize,
+        challenge_offsets: Vec<usize>,
         has_public_inputs: bool,
     ) -> Self {
+        debug_assert_eq!(
```
ashpect: use `assert_eq!`. any reasoning to use debug_assert ?

provekit/common/src/prefix_covector.rs
ocdbytes: prefixvector and offsetvector had debug_assert so I followed the pattern but added assert to SparseCovector. We can change prefixvector and offsetvector to have assert in another PR

PR #369, opened 2026-03-22
fix DSC Key Smuggling via Unsigned TBS Data [LA-A]

x-senpai-x:
## Summary

Adds a check_zero_padding(tbs_certificate, tbs_certificate_len) constraint in t_add_dsc_verify_1300 immediately after the length bounds check
Imports utils::check_zero_padding, which already exists and is used in the 720-byte path

## Background
The 1300-byte DSC registration circuit splits SHA-256 hashing across two sub-circuits (t_add_dsc_hash_1300 + t_add_dsc_verify_1300) using a pre-computed hash. Unlike the 720-byte path — which calls sig_check_rsa::verify_signature → sha256_and_check_data_to_sign → check_zero_padding automatically — the 1300-byte path passes a pre-computed msg_hash: [u8; 32] directly to fragmented_sig_check_rsa::verify_rsa_signature, which never sees the raw TBS buffer. This meant the zero-padding guard was never applied.


noir-examples/noir-passport/merkle_age_check/t_add_id_data_720/src/main.nr
ashpect, line 1: No explicit bounds assert after get_asn1_element_length(tbs_certificate). If a malformed DER header claims length > buffer size, there's no guard before verify_rsa_pubkey_in_tbs. The Noir prover would trap on OOB access, but an explicit assert(tbs_certificate_len <= TBS_CERT_SIZE) would be cleaner and match what t_add_dsc_verify_1300 does.

ashpect (approved):

noir-examples/noir-passport/merkle_age_check/t_add_id_data_720/src/main.nr
x-senpai-x: we have bound check in t_add_dsc_720, and via commitment we verify that the same tbs_certificate is used again so isn't it redundant to add another assertion?

noir-examples/noir-passport/utils/data-check/tbs-pubkey/src/lib.nr
```diff
@@ -84,7 +142,7 @@ fn test_verify_rsa_pubkey_in_tbs_with_allowed_exponent_65537() {
     tbs[5] = 0x00;
     tbs[6] = 0x01;
 
-    verify_rsa_pubkey_in_tbs(modulus, tbs, 0, 65537, 4);
+    verify_rsa_pubkey_in_tbs(modulus, tbs, 7, 0, 65537, 4);
```
rose2221: the five old exponent tests (lines 131-203) are stale, they were written for the pre-refactor function signature and now pass invalid tbs_certificate_len = 0 and pubkey_offset values too small for the new DER validation. They need to be updated, similar to the new build_test_tbs-based tests below.

noir-examples/noir-passport/utils/data-check/tbs-pubkey/src/lib.nr
```diff
@@ -1,57 +1,115 @@
...
-    //   109729 (0x01AC61) -> [0x01, 0xAC, 0x61]
-    //   122125 (0x01DC0D) -> [0x01, 0xDC, 0x0D]
-    //   130689 (0x01FF01) -> [0x01, 0xFF, 0x01]
+    //   109729 (0x01ACA1) -> [0x01, 0xAC, 0xA1]
+    //   122125 (0x01DD0D) -> [0x01, 0xDD, 0x0D]
+    //   130689 (0x01FE81) -> [0x01, 0xFE, 0x81]
     //
     // For the 3-byte case, u32 >> 16 naturally gives 0x00 for exponents needing a sign byte,
     // and the expected 3-byte sequence matches the DER value encoding directly.
     if exponent < 0x100 {
```
rose2221, line 110: TODO(security): Validate `tbs_certificate[exponent_offset - 2] == 0x02` (INTEGER tag) and `tbs_certificate[exponent_offset - 1] == exponent_byte_count` (length byte).
Without this, exponent_offset can point at coincidental byte matches elsewhere in TBS.

x-senpai-x: right this makes sense, adding it

PR #370, opened 2026-03-22
fix: binop width selection to preserve byte semantics [LA-D]

rose2221:
## Summary
- Restricted `get_optimal_binop_width` to only return widths that divide 8 (`{2, 4, 8}`) instead of the full range `[2, 8]`. Non-dividing widths (e.g. 3, 5, 6, 7) cause digit decompositions that overallocate bit capacity beyond `[0, 255]`, breaking byte semantics.
- Added unit tests for the binop module:
  - `optimal_binop_width_always_divides_8`: verifies the width invariant holds for all input sizes up to 1024.
  - `non_canonical_byte_rejected_by_recomposition`: regression test confirming that the recomposition constraint rejects non-canonical byte values (> 255) when using `w ∈ {2, 4}`.


provekit/r1cs-compiler/src/binops.rs
Bisht13: Convert this debug_assert! to assert! and update the logic.

PR #374, opened 2026-03-22
Reject malformed 32-bit constants in AND/XOR and SHA256 compilation [LASuggestion3]

rose2221:
- Add 32-bit width assertions before as u32 truncation in process_binop_opcode (3 sites) to reject constants exceeding 32 bits at compile time instead of silently truncating
- Add matching assertion in cow_to_digit in binops.rs for self-contained validation independent of callers
- Add 32-bit width assertion in sha256_compression.rs for hash value constants used in the optimization path
- Refactor all assertion sites to bind into_bigint() once and reuse, avoiding redundant Montgomery-to-integer conversions


Bisht13 (approved):

provekit/r1cs-compiler/src/noir_to_r1cs.rs
```diff
@@ -350,12 +350,23 @@ impl NoirToR1CSCompiler {
             (fe.into_bigint().0[bit_offset / 64] >> (bit_offset % 64)) as u8 as u64
         }
 
+        /// Assert that a field element fits in 32 bits.
+        fn assert_fits_32_bits(fe: &FieldElement) {
```
Bisht13: Bug: this hardcodes 32 bits, but process_binop_opcode is width-parameterized by num_bits and supports up to 128 (see commit cfcd5daa and the bin-opcode-u128 test). Upstream FunctionInput::constant(value, max_bits) accepts wide constants, and bit_and/bit_xor solvers have u128 property tests. A Noir program like a & 0xFFFF_FFFF_FF on a u64 would be incorrectly rejected. The check should validate against num_bits, not 32.

provekit/r1cs-compiler/src/binops.rs
```diff
@@ -84,7 +84,12 @@ fn cow_to_digit(
 ) -> ConstantOrR1CSWitness {
     match cow {
         ConstantOrR1CSWitness::Constant(c) => {
-            let val = c.into_bigint().0[0];
+            let bigint = c.into_bigint();
+            assert!(
```
Bisht13: Redundant: constants reaching cow_to_digit have already been byte-decomposed by process_binop_opcode via const_byte, so values here are ≤ 255 in normal flow. This 32-bit check is unreachable. Either drop it, or tighten to <= u8::MAX with a comment documenting the byte-decomposition invariant.

provekit/r1cs-compiler/src/binops.rs
```diff
@@ -586,4 +591,26 @@ mod tests {
             );
         }
     }
+
+    /// Verify that `cow_to_digit` panics when given a constant exceeding 32
+    /// bits. Regression test for Least Authority audit Suggestion 3.
+    #[test]
+    #[should_panic(expected = "binop constant exceeds 32 bits")]
```
Bisht13: This test panics on a helper with input that the real compiler flow would never produce. It doesn't exercise the user-facing regression. Replace with an end-to-end test: compile a Noir program doing u64/u128 AND/XOR with a constant > u32::MAX but within declared width, it should compile successfully (will catch the bug in comment https://github.com/worldfnd/provekit/pull/374#discussion_r3061299567).

Bisht13 (changes requested):

provekit/r1cs-compiler/src/noir_to_r1cs.rs
rose2221: fixed :- now assert_fits_num_bits(num_bits)

provekit/r1cs-compiler/src/binops.rs
rose2221: Dropped the unreachable <= u32::MAX assertion in cow_to_digit

provekit/r1cs-compiler/src/binops.rs
rose2221: fixed. replaced the helper-level cow_to_digit_rejects_oversized_constant test w an e2e regression that compiles a noir program through the full pipeline.

Bisht13 (approved):

PR #392, opened 2026-03-31
test(msm): add negative/adversarial tests for MSM constraint satisfaction

### LINKED ISSUE #348: Test: Add negative/adversarial tests for MSM constraint satisfaction
## Problem

The MSM implementation (PR #310) has thorough **positive tests** — 7 R1CS satisfaction tests in `msm_witness_solving.rs` plus 5 Noir e2e prove/verify tests — but **zero negative tests** verifying that invalid witnesses are rejected.

For a zero-knowledge proof system where soundness is critical, testing that the constraint system **rejects** malformed witnesses is as important as testing that it **accepts** valid ones. A constraint bug that silently accepts wrong witnesses would allow proof forgery.

## Proposed Tests

Add tests in `tooling/provekit-bench/tests/msm_witness_solving.rs` that compile a circuit, solve witnesses, **corrupt a specific witness value**, and assert that `check_r1cs_satisfaction` **fails**. Each test targets a different constraint layer:

### 1. Wrong output coordinates
```
Compile single-point MSM, solve with correct scalar, then replace output x-limb with a different value.
Expected: R1CS satisfaction fails (output constraining catches the mismatch).
```

### 2. Off-curve input point
```
Compile single-point MSM, set input point to (Gx, Gy + 1) — a point NOT on the curve.
Expected: R1CS satisfaction fails (on-curve check y² ≠ x³ + ax + b).
```

### 3. Wrong scalar (output doesn't match)
```
Compile single-point MSM with scalar=7, but set expected output to [5]*G instead of [7]*G.
Expected: R1CS satisfaction fails (identity check against accumulated offset).
```

### 4. s2 = 0 forgery attempt
```
Compile single-point MSM, solve normally, then zero out the s2 witness from FakeGLV.
Expected: R1CS satisfaction fails (s2 ≠ 0 check in scalar_relation.rs:87-88).
```

### 5. Flipped sign bit
```
Compile single-point MSM, solve normally, then flip neg1 from 1→0 (or 0→1).
Expected: R1CS satisfaction fails (scalar relation no longer holds).
```

### Implementation approach

Each test follows the existing `run_single_point_msm_test_limbed` pattern but adds a corruption step between `solve_witnesses` and `check_r1cs_satisfaction`:

```rust
let witness = solve_witnesses(...);

// Corrupt a specific witness
let mut corrupted = witness.clone();
corrupted[target_idx] = FieldElement::from(42u64);  // wrong value

// Verify the original passes
check_r1cs_satisfaction(&compiler.r1cs, &witness).expect("valid witness should pass");

// Verify the corrupted version fails
assert!(
    check_r1cs_satisfaction(&compiler.r1cs, &corrupted).is_err(),
    "corrupted witness should fail R1CS satisfaction"
);
```

Finding the right `target_idx` for each corruption requires inspecting the witness builder layout — this is the main implementation effort.

## Priority

Medium — the positive tests and the identity check against hardcoded accumulated offsets provide strong coverage already, but negative tests are standard practice for cryptographic code and would catch regressions in constraint completeness.

PranavPipariya:
## Summary

Adds the negative/adversarial MSM constraint-satisfaction tests proposed in #348.

This PR is test-only and does not change MSM implementation behavior.

Closes #348.

## Included tests

- wrong output coordinate
- off-curve input point
- wrong scalar/output pairing
- `s2 = 0` forgery attempt
- flipped `neg1` sign bit

## Validation

- `cargo +nightly-2026-03-04 fmt --all --check`
- `cargo +nightly-2026-03-04 clippy --all-targets --all-features --verbose`
- `cargo +nightly-2026-03-04 test -p provekit-bench --test msm_witness_solving`
- `cargo +nightly-2026-03-04 test --no-fail-fast --all-features --verbose --lib --tests --bins`

### DIFF
--- a/tooling/provekit-bench/tests/msm_witness_solving.rs
+++ b/tooling/provekit-bench/tests/msm_witness_solving.rs
@@ -505,6 +505,33 @@ fn test_single_point_rejects_wrong_output_coordinate() {
     );
 }
 
+/// Corrupting an expected output y-limb must also violate the output equality
+/// constraints.
+#[test]
+fn test_single_point_rejects_wrong_output_y_coordinate() {
+    let curve = Secp256r1;
+    let gx = curve.generator().0;
+    let gy = curve.generator().1;
+    let scalar: [u64; 4] = [7, 0, 0, 0];
+    let (ex, ey) = ec_scalar_mul(
+        &gx,
+        &gy,
+        &scalar,
+        &curve.curve_a(),
+        &curve.field_modulus_p(),
+    );
+    let fixture = build_single_point_msm_fixture(&gx, &gy, false, &scalar, &ex, &ey, false);
+
+    assert_single_point_corruption_is_rejected(
+        fixture,
+        "wrong output y coordinate",
+        |layout, corrupted| {
+            let idx = layout.out_y_limbs[0];
+            corrupted[idx] = different_field_element(corrupted[idx]);
+        },
+    );
+}
+
 /// Corrupting the input y-coordinate must violate the curve-membership and
 /// consistency constraints.
 #[test]
@@ -563,6 +590,34 @@ fn test_single_point_rejects_wrong_scalar_output_pairing() {
         |layout, corrupted| {
             overwrite_witness_values(corrupted, &layout.out_x_limbs, &wrong_ex_fes);
             overwrite_witness_values(corrupted, &layout.out_y_limbs, &wrong_ey_fes);
+        },
+    );
+}
+
+/// Replacing the scalar input while keeping the original output must fail.
+#[test]
+fn test_single_point_rejects_scalar_corruption() {
+    let curve = Secp256r1;
+    let gx = curve.generator().0;
+    let gy = curve.generator().1;
+    let scalar: [u64; 4] = [7, 0, 0, 0];
+    let wrong_scalar: [u64; 4] = [5, 0, 0, 0];
+    let (wrong_lo, wrong_hi) = split_scalar(&wrong_scalar);
+    let (ex, ey) = ec_scalar_mul(
+        &gx,
+        &gy,
+        &scalar,
+        &curve.curve_a(),
+        &curve.field_modulus_p(),
+    );
+    let fixture = build_single_point_msm_fixture(&gx, &gy, false, &scalar, &ex, &ey, false);
+
+    assert_single_point_corruption_is_rejected(
+        fixture,
+        "scalar corruption",
+        |layout, corrupted| {
+            corrupted[layout.scalar_lo] = u256_to_fe(&wrong_lo);
+            corrupted[layout.scalar_hi] = u256_to_fe(&wrong_hi);
         },
     );
 }
--- a/tooling/provekit-bench/tests/msm_witness_solving.rs
+++ b/tooling/provekit-bench/tests/msm_witness_solving.rs
@@ -124,15 +124,98 @@ fn u256_to_limb_fes(v: &[u64; 4], limb_bits: u32, num_limbs: usize) -> Vec<Field
     decompose_to_limbs(v, limb_bits, num_limbs)
 }
 
-// ---------------------------------------------------------------------------
-// Single-point limbed MSM test runner
-// ---------------------------------------------------------------------------
+/// Increment a 256-bit little-endian integer by one.
+fn increment_u256(v: &[u64; 4]) -> [u64; 4] {
+    let mut out = *v;
+    for limb in &mut out {
+        let (next, carry) = limb.overflowing_add(1);
+        *limb = next;
+        if !carry {
+            break;
+        }
+    }
+    out
+}
 
-/// Compile and solve a single-point MSM circuit using the limbed API.
-///
-/// When `expected_inf` is true, the expected output is point at infinity
-/// (all output limbs zero, out_inf = 1).
-fn run_single_point_msm_test_limbed(
+/// Pick a different field element without relying on arithmetic traits.
+fn different_field_element(value: FieldElement) -> FieldElement {
+    if value == FieldElement::zero() {
+        FieldElement::from(1u64)
+    } else {
+        FieldElement::zero()
+    }
+}
+
+/// Overwrite specific witness slots with the supplied values.
+fn overwrite_witness_values(
+    witness: &mut [FieldElement],
+    indices: &[usize],
+    values: &[FieldElement],
+) {
+    assert_eq!(
+        indices.len(),
+        values.len(),
+        "indices and values must have the same length"
+    );
+    for (&idx, &value) in indices.iter().zip(values.iter()) {
+        witness[idx] = value;
+    }
+}
+
+#[derive(Clone, Copy, Debug)]
+struct FakeGlvWitnessIndices {
+    s1:   usize,
+    s2:   usize,
+    neg1: usize,
+    neg2: usize,
+}
+
+#[derive(Clone, Debug)]
+struct SinglePointMsmLayout {
+    point_x_limbs: Vec<usize>,
+    point_y_limbs: Vec<usize>,
+    point_inf:     usize,
+    scalar_lo:     usize,
+    scalar_hi:     usize,
+    out_x_limbs:   Vec<usize>,
+    out_y_limbs:   Vec<usize>,
+    out_inf:       usize,
+    fake_glv:      FakeGlvWitnessIndices,
+}
+
+struct SinglePointMsmFixture {
+    compiler:       NoirToR1CSCompiler,
+    num_witnesses:  usize,
+    initial_values: Vec<(usize, FieldElement)>,
+    layout:         SinglePointMsmLayout,
+}
+
+/// Locate the single FakeGLV hint used by the single-point MSM circuit.
+fn locate_single_point_fake_glv(builders: &[WitnessBuilder]) -> FakeGlvWitnessIndices {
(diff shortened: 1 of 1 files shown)


ocdbytes (changes requested):
@PranavPipariya 

Can you also add two more test cases here :
- out_y limb corruption (same as written for out_x).
- scalar corruption (replace the scalar but keep output computed from the original scalar). this is to check end to end scalar output binding.

After this will give one more review for the code.

tooling/provekit-bench/tests/msm_witness_solving.rs
```diff
@@ -325,6 +478,213 @@ fn test_arbitrary_point_and_scalar() {
...
+    let gx = curve.generator().0;
+    let gy = curve.generator().1;
+    let scalar: [u64; 4] = [7, 0, 0, 0];
+    let (ex, ey) = ec_scalar_mul(
+        &gx,
+        &gy,
+        &scalar,
+        &curve.curve_a(),
+        &curve.field_modulus_p(),
+    );
```
ocdbytes: Ig we can de duplicate this among the tests

tooling/provekit-bench/tests/msm_witness_solving.rs
```diff
@@ -124,23 +124,106 @@ fn u256_to_limb_fes(v: &[u64; 4], limb_bits: u32, num_limbs: usize) -> Vec<Field
...
-/// (all output limbs zero, out_inf = 1).
-fn run_single_point_msm_test_limbed(
+/// Pick a different field element without relying on arithmetic traits.
+fn different_field_element(value: FieldElement) -> FieldElement {
+    if value == FieldElement::zero() {
+        FieldElement::from(1u64)
+    } else {
+        FieldElement::zero()
+    }
+}
```
ocdbytes: we can use a better approach here rather than toggling between 0 and 1

ocdbytes (changes requested):
Also add a test for 
- out_inf corruption
- neg2 flip
- multi point MSM (this will catch any error in the process multi point path)

NIT : for tests which dont require much changes and are mostly duplicated logic like out_x and out_y or neg_1 and neg_2. We can have a single test and add it as a case rather than having two different tests with mostly same logic.

PranavPipariya:
Made the updates:

- consolidated the duplicated `out_x`/`out_y` and `neg1`/`neg2` cases into case-driven tests
- added coverage for `out_inf`, `neg2`, and the multi-point MSM negative path
- updated the `neg1`/`neg2` case to use an actual boolean flip

Also re-ran the validation commands locally.

ocdbytes:
@PranavPipariya Please address the above comments regarding the code and then will give a final review.

tooling/provekit-bench/tests/msm_witness_solving.rs
PranavPipariya: Consolidated the duplicated `out_x`/`out_y` and `neg1`/`neg2` cases into case-driven tests that share the setup. Please take a look at `73f41e68`.

tooling/provekit-bench/tests/msm_witness_solving.rs
PranavPipariya: Replaced `different_field_element`; the general corruption path now uses `corrupt_field_element`, and the `neg1`/`neg2` case uses an actual boolean flip. Please take a look.

tooling/provekit-bench/tests/msm_witness_solving.rs
ocdbytes: This logic : 

```rust
let curve = Secp256r1;
    let gx = curve.generator().0;
    let gy = curve.generator().1;
    let scalar: [u64; 4] = [7, 0, 0, 0];
    let (ex, ey) = ec_scalar_mul(
        &gx,
        &gy,
        &scalar,
        &curve.curve_a(),
        &curve.field_modulus_p(),
    );
```

appears too many times which is just a repeating code. Ig this can be delegated to a single function

tooling/provekit-bench/tests/msm_witness_solving.rs
```diff
@@ -525,3 +860,108 @@ fn test_two_point_one_zero_scalar() {
     check_r1cs_satisfaction(&compiler.r1cs, &witness)
         .expect("R1CS satisfaction check failed for two-point MSM with one zero scalar");
 }
+
+/// Two-point MSM: corrupting the output must be rejected, exercising the
+/// multi-point accumulation and output-constraining path.
+#[test]
+fn test_two_point_msm_rejects_wrong_output() {
```
ocdbytes, line 850: this new test it just copy of the test `build_single_point_msm_fixture.` We can de duplicate this.

tooling/provekit-bench/tests/msm_witness_solving.rs
ocdbytes: Pre existing test should also be using the the function for params

tooling/provekit-bench/tests/msm_witness_solving.rs
PranavPipariya: Yeah, you're totally right. should've been more DRY in the first place. have extracted the repeated single-point generator setup into a shared helper and updated the affected tests to use it. please take a look.

tooling/provekit-bench/tests/msm_witness_solving.rs
ocdbytes: remove these print statements these were left in previous review

tooling/provekit-bench/tests/msm_witness_solving.rs
```diff
@@ -325,6 +529,137 @@ fn test_arbitrary_point_and_scalar() {
...
+            },
+        );
+    }
+}
+
+/// Corrupting the output infinity flag must violate the output constraints.
+#[test]
+fn test_single_point_rejects_wrong_output_inf() {
+    let case = secp256r1_generator_case([7, 0, 0, 0]);
+    let fixture = build_generator_single_point_fixture(case);
```
ocdbytes: NIT : 

The pattern here of calling `secp256r1_generator_case` and `build_generator_single_point_fixture` looks redundant use something like this : 

```rust
 fn generator_fixture(scalar: [u64; 4]) -> SinglePointMsmFixture {
      build_generator_single_point_fixture(secp256r1_generator_case(scalar))
  } 
```

so the tests which require their own case internally can call otherwise 

```rust
  fn test_single_point_rejects_wrong_output_inf() {
      let fixture = generator_fixture([7, 0, 0, 0]);            // ← one call
......
  }
```

tooling/provekit-bench/tests/msm_witness_solving.rs
```diff
@@ -124,23 +124,157 @@ fn u256_to_limb_fes(v: &[u64; 4], limb_bits: u32, num_limbs: usize) -> Vec<Field
...
+        })
+        .collect();
+
+    assert_eq!(
+        fake_glv_indices.len(),
+        1,
+        "single-point MSM circuit should contain exactly one FakeGLV hint"
+    );
+    fake_glv_indices[0]
+}
```
ocdbytes: NIT : 

ik this is a test code but here collect will take each iter struct result and then allocate a vec for that. Insted we only want one result which has one FakeGLV hint so we can just iterate over it without allocating a vector for this : 

```suggestion
  fn locate_single_point_fake_glv(builders: &[WitnessBuilder]) -> FakeGlvWitnessIndices {
      let mut iter = builders.iter().filter_map(|builder| match builder {
          WitnessBuilder::FakeGLVHint { output_start, .. } => Some(FakeGlvWitnessIndices {
              s1:   *output_start,
              s2:   *output_start + 1,
              neg1: *output_start + 2,
              neg2: *output_start + 3,
          }),
          _ => None,
      });

      let result = iter.next().expect("should have at least one FakeGLV hint");
      assert!(iter.next().is_none(), "should have exactly one FakeGLV hint");
      result
  }
```

first next() checks that there is a element 
second next() confirms only one element is there

ocdbytes (changes requested):
Some NITS. 
Other than this things look good to me !!

tooling/provekit-bench/tests/msm_witness_solving.rs
PranavPipariya: removed the print statements

tooling/provekit-bench/tests/msm_witness_solving.rs
PranavPipariya: done. have extracted the shared setup into `build_default_two_point_msm_fixture`. Both `test_two_point_msm` and `test_two_point_msm_rejects_wrong_output` use it now. please take a look.

tooling/provekit-bench/tests/msm_witness_solving.rs
PranavPipariya: done. the pre-existing positive tests now use `secp256r1_generator_params` (or `secp256r1_generator` where curve params aren't needed).

tooling/provekit-bench/tests/msm_witness_solving.rs
PranavPipariya: Added generator_fixture as mentioned. simple tests now use the one-liner. `test_single_point_rejects_off_curve_input` and `test_single_point_rejects_wrong_scalar_output_pairing` keep the two-step form because they read case fields before building the fixture.

ocdbytes (approved):

ocdbytes:
@PranavPipariya fix the lint and then we can merge the PR

Bisht13 (approved):

Bisht13:
Thanks for the contribution @PranavPipariya!

PR #412, opened 2026-04-16
Added poseidon2 as a hashconfig 

x-senpai-x:


provekit/common/src/poseidon2/whir.rs
```diff
@@ -0,0 +1,177 @@
...
+            let msg = &input[i * size..(i + 1) * size];
+            out_hash.0 = hash_message(msg, chunks_per_msg);
+        }
+    }
+}
+
+/// Sponge hash of a single `num_fes`-element message. The capacity lane is
+/// pre-seeded with `IV = num_fes * 2^64` to prevent zero-padding collisions.
+/// Output is `state[0]` after the final permutation.
+fn hash_message(msg: &[u8], num_fes: usize) -> [u8; 32] {
```
Bisht13: This sponge logic duplicates poseidon2::poseidon2_hash’s IV/absorb construction.

provekit/common/src/poseidon2/whir.rs
```diff
@@ -0,0 +1,177 @@
...
+/// Derived as `SHA3-256("whir::hash" || "poseidon2")`.
+pub const POSEIDON2: EngineId = EngineId::new([
+    0xab, 0x4c, 0x81, 0x0f, 0x79, 0xdd, 0xf2, 0xa5, 0x4e, 0x81, 0x73, 0x1f, 0x97, 0x08, 0x23, 0x6a,
+    0xcf, 0x5c, 0xc2, 0xcc, 0x35, 0xe9, 0xc4, 0x57, 0xa4, 0x35, 0x37, 0xdc, 0x01, 0x23, 0x6b, 0xb7,
+]);
+
+/// `2^64` as a BN254 field element; capacity-lane IV multiplier. Matches the
+/// construction in [`poseidon2::poseidon2_hash`] so this engine and the
+/// one-shot hash stay semantically aligned.
+static TWO_POW_64: LazyLock<Fr> = LazyLock::new(|| Fr::from(1u64 << 32) * Fr::from(1u64 << 32));
```
Bisht13: TWO_POW_64 is duplicated from poseidon2/src/hash.rs.

provekit/common/src/utils/mod.rs
```diff
@@ -7,9 +7,26 @@ pub mod serde_jsonify;
 pub mod sumcheck;
 
 pub use self::print_abi::PrintAbi;
+
+/// Deserializes a BN254 field element from up to 32 little-endian bytes.
```
Bisht13, line 11: This adds new field/byte helpers, but we already have overlapping serializers elsewhere (hash_config, skyscraper::sponge).

provekit/common/src/hash_config.rs
```diff
@@ -208,6 +217,16 @@ where
     FieldElement::from_le_bytes_mod_order(&hasher.finalize())
 }
 
+/// Poseidon2 one-shot hash over `elements` (including empty input).
+///
+/// Does NOT use [`PUBLIC_INPUTS_DST`] — length domain-separation is handled
```
Bisht13: Poseidon2 skips PUBLIC_INPUTS_DST while the digest-based variants include it.

provekit/common/src/poseidon2/whir.rs
```diff
@@ -0,0 +1,177 @@
...
+//!
+//! Messages are multiples of 32 bytes; each 32-byte chunk is a BN254 field
+//! element absorbed into the rate portion of a 4-lane state seeded with
+//! `IV = num_fes * 2^64` in the capacity lane. `state[0]` after the final
+//! permutation is the 32-byte output.
+
+use {
+    crate::utils::{bytes_to_field, field_to_bytes_le},
+    ark_bn254::Fr,
+    ark_ff::Zero,
```
Bisht13: Nit: the sibling poseidon2 crate uses ark_std::Zero throughout.

provekit/common/src/utils/mod.rs
```diff
@@ -7,9 +7,26 @@ pub mod serde_jsonify;
...
+#[inline]
+pub fn bytes_to_field(bytes: &[u8]) -> FieldElement {
```
Bisht13, line 13: These only have two call sites (the two new Poseidon2 files), yet hash_config::fe_to_bytes_le and skyscraper::sponge::{bytes_to_fr, fr_to_bytes} still exist. We now have three serializers and two deserializers with different semantics (utils::bytes_to_field reduces; skyscraper::bytes_to_fr assumes canonical input). Either route everyone through utils::* and retire the rest, or keep the helpers local to the Poseidon2 module.

provekit/common/src/hash_config.rs
```diff
@@ -133,6 +141,7 @@ impl HashConfig {
             Self::Sha256 => hash_digest::<sha2::Sha256>(PUBLIC_INPUTS_DST, elements),
             Self::Keccak => hash_digest::<sha3::Keccak256>(PUBLIC_INPUTS_DST, elements),
             Self::Blake3 => hash_blake3(PUBLIC_INPUTS_DST, elements),
+            Self::Poseidon2 => hash_poseidon2(elements),
```
Bisht13, line 149: Every non-Skyscraper public-input binding here is explicitly domain-separated via PUBLIC_INPUTS_DST, but Poseidon2 uses raw poseidon2_hash(elements). Is that intentional? I think Poseidon2 should get explicit public-input domain separation too, otherwise this becomes a protocol-format decision we may regret later.

provekit/common/src/poseidon2/sponge.rs
```diff
@@ -0,0 +1,98 @@
+//! Poseidon2 duplex sponge for Fiat-Shamir.
```
Bisht13, line 1: Please pin the canonical-lane invariant in a //! note. bytes_to_field = from_le_bytes_mod_order reduces each 32-byte lane mod p, so this adapter is only byte-injective when every absorbed lane is already < p. Spongefish writes raw bytes straight into permutation_state (duplex_sponge.rs:197-213), and the domain separator absorbs arbitrary as_bytes(), we rely on the protocol to only ever absorb canonical field-element encodings. Currently true (all absorbs are field elements or engine outputs, which are canonical), but nothing enforces it. SkyscraperSponge has the same issue via Fr::new(BigInt(_)), codebase-wide, not introduced here. Worth a follow-up issue for a byte-injective adapter.

provekit/common/src/poseidon2/whir.rs
Bisht13: This is poseidon2::poseidon2_hash reimplemented over bytes. I walked num_fes ∈ {1, 3, 4, 6}, state evolution and permutation counts match exactly (num_fes = 0 diverges but is unreachable via supports_size). Please either expose a byte-oriented variant from the poseidon2 crate, or at minimum add a cross-consistency test:
```rs
#[test]
fn whir_engine_matches_poseidon2_hash() {
    for n in 1..=10usize {
        let fs: Vec<Fr> = (1..=n as u64).map(Fr::from).collect();
        let msg: Vec<u8> = fs.iter().flat_map(|f| field_to_bytes_le(*f)).collect();
        let mut out = [Hash::default()];
        Poseidon2HashEngine.hash_many(n * 32, &msg, &mut out);
        assert_eq!(out[0].0, field_to_bytes_le(poseidon2::poseidon2_hash(&fs)));
    }
}
```

Bisht13 (changes requested):

Bisht13:
Can we include HashConfig::Poseidon2 in provekit-bench/tests/compiler.rs loop as well? This PR wires Poseidon2 through the runtime hash-config path, so it would be good to cover it in the same regression test.

Bisht13:
Near the invalid hash-config tests in ffi_integration.rs, we validate bad hash-config values, but I don’t see a happy-path test for the new 4 = Poseidon2 case. Please add one FFI round-trip using the new value.

provekit/common/src/utils/mod.rs
x-senpai-x: yeah the objective was to move them to common mod.rs, I think there has been a problem with rebase because initially I had moved the logic out of `skyscraper/ `and inside `common/src/utils `instead, I'll remove it from skyscaper

provekit/common/src/hash_config.rs
x-senpai-x: Yes I think we should add domain separator here, current version only prevents collision b/w different length inputs but doesn't help in separating public inputs hashing with other hashes being used.

provekit/common/src/poseidon2/sponge.rs
x-senpai-x: Raised an [issue](https://github.com/worldfnd/provekit/issues/419) for this

provekit/common/src/hash_config.rs
Bisht13: This helper is now a duplicate of utils::field_to_bytes_le (plus a debug_assert!).

provekit/common/src/poseidon2/whir.rs
```diff
@@ -0,0 +1,149 @@
...
+impl HashEngine for Poseidon2HashEngine {
+    fn name(&self) -> Cow<'_, str> {
+        "poseidon2".into()
+    }
+
+    fn supports_size(&self, size: usize) -> bool {
+        size > 0 && size % 32 == 0
+    }
+
+    fn preferred_batch_size(&self) -> usize {
```
Bisht13: hash_many is sequential, no SIMD, no batch path. Trait default is 1; DigestEngine returns 1. Returning 4 misleads the caller's work-distribution heuristic. Delete the override.

provekit/common/src/hash_config.rs
```diff
@@ -208,6 +222,22 @@ where
     FieldElement::from_le_bytes_mod_order(&hasher.finalize())
 }
 
+/// Poseidon2 one-shot hash over `elements` (including empty input).
+///
+/// Prepends [`PUBLIC_INPUTS_DST_FE`] as the first absorbed field element
```
Bisht13, line 213: This still doesn’t add explicit public-input DST for Poseidon2 yet, it still delegates to plain poseidon2_hash(elements).

provekit/common/src/witness/mod.rs
```diff
@@ -351,6 +366,20 @@ mod tests {
         0x4e, 0x47, 0x7d, 0x1f, 0xf9, 0xf5, 0x79, 0xc1, 0x46, 0xb4, 0x28, 0x84, 0xa5, 0x6b, 0xc5,
         0xa5, 0x25,
     ];
+    // Poseidon2([]) = poseidon2_hash([PUBLIC_INPUTS_DST_FE]) in LE bytes —
```
Bisht13, line 368: This comment now describes Poseidon2 as hashing a DST-derived field element first, but PublicInputs::hash() still delegates to hash_field_elements(&self.0), and the HashConfig::Poseidon2 path still looks undomain-separated.

Bisht13 (approved):

PR #414, opened 2026-04-20
Use Verity internally in the wasm demo

Bisht13:
## Summary
- keep the existing ProveKit wasm demo UX while routing prove/verify through the published `@atheonxyz/verity` package internally
- remove the demo-local noir-init and direct `provekit_wasm` browser wiring
- update setup to install the npm package and regenerate demo artifacts without depending on the local `../atheon/verity` checkout

## Testing
- `node --check playground/wasm-demo/src/demo-web.mjs`
- `node --check playground/wasm-demo/scripts/setup.mjs`
- `node --check playground/wasm-demo/scripts/serve.mjs`
- `npm run setup` (in `playground/wasm-demo`)
- node smoke using `@atheonxyz/verity` to prove + verify against demo artifacts
- headless browser: built-in SHA256 prove + verify
- headless browser: custom upload prove + verify
- curl header checks for COOP/COEP and Verity wasm asset serving

## Notes
- The demo stays visibly branded as ProveKit; Verity is kept as an internal runtime detail.
- This version uses the published npm package only, per request.


Bisht13:
Updated this branch to fix the metrics bug: constraints and witnesses are now read directly from the PKP in-browser instead of coming from metadata/opcode counts. Verified locally in the browser: SHA256 now shows 14,038 constraints and 25,220 witnesses.

PR #442, opened 2026-05-14
Add macro based NTT

zkfriendly:
## Summary

Ports the ProveKit NTT path from main:
- [main RS integration](https://github.com/worldfnd/provekit/blob/cc391c8cc72766cc47f8243402e7f51e16c6d7cd/provekit/common/src/ntt.rs)
- [main NTT crate](https://github.com/worldfnd/provekit/tree/cc391c8cc72766cc47f8243402e7f51e16c6d7cd/ntt)

The RS interface on this branch is slightly different from main WHIR, so this PR keeps the port minimal and adapts the interleaved encoding layout locally to match the older interface.

## Notes

Main also has a [B51 backend for wasm](https://github.com/worldfnd/provekit/blob/cc391c8cc72766cc47f8243402e7f51e16c6d7cd/ntt/src/b51_interleaved.rs).

That backend is not ported here because it depends on the `bn254` implementation from Skyscraper, which is also behind main on this branch. Pulling it in looks like a larger sync/refactor. Should we do that as part of this PR, or leave it for a follow-up?

## Results

On `complete_age_check`, 10 runs each (`~/before` vs `~/after`):

| metric | before | after|
| --- | ---: | ---: |
| allocator peak | `816 MB` | `774 MB` |
| max RSS avg | `1358.1 MiB` | `1025.2 MiB` |
| macOS peak footprint avg | `1328.1 MiB` | `994.9 MiB` |
| `prove_with_witness` avg | `1.100 s` | `0.993 s` |
| full `run` avg | `1.831 s` | `1.743 s` |
| wall-clock `real` avg | `1.838 s` | `1.752 s` |

The current `after` build keeps the memory win (`42 MB` allocator peak, about `333 MiB`
RSS/footprint) and is also faster in this run: about `9.7%` faster in
`prove_with_witness` and about `4.8%` faster for the full `prove` command.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3493,6 +3493,8 @@ dependencies = [
  "rand_chacha 0.9.0",
  "rand_xorshift 0.4.0",
  "regex-syntax",
+ "rusty-fork",
+ "tempfile",
  "unarray",
 ]
 
@@ -3614,7 +3616,7 @@ checksum = "95067976aca6421a523e491fce939a3e65249bac4b977adee0ee9771568e8aa3"
 
 [[package]]
 name = "provekit-bench"
-version = "0.1.0"
+version = "1.0.0"
 dependencies = [
  "anyhow",
  "divan",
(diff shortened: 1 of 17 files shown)


zkfriendly:
should we publish ntt as a package (similar to whir for example) and use it like that?

zkfriendly:
closing this. Changes are merged for v2

PR #455, opened 2026-06-16
refactor: make the protocol spine field-agnostic; extract provekit-field-bn254 + provekit-noir

BornPsych:
## Summary

Makes the proving spine (`provekit-common`, `provekit-prover`, `provekit-verifier`) field-agnostic and moves all bn254-specific code into a new `provekit-field-bn254` crate. The Noir/mavros frontend moves into a new `provekit-noir` crate. **bn254 output is unchanged from v2, byte for byte.**

The field is now chosen by which crate a binary depends on, not by a Cargo feature. The spine names no concrete field; a binary links the one field crate it registers. Adding a field is a new crate plus a `register()` call.

This PR is a **pure decomposition** — it moves code and severs dependencies. Making the spine *generic* over the field (so a second field like Goldilocks can plug in) is a deliberate follow-up; see "Why no generics yet" below and the tracking issue #456.

## Why a runtime registry

The orphan rule blocks `impl SomeTrait for ark_bn254::Fr` from living in `provekit-field-bn254` (it can only live in the trait's crate or in arkworks). So instead of a per-field trait impl, `common` exposes a `FieldHashProvider` registry — the same mechanism whir uses for its `ENGINES`/`NTT` registries. `provekit_field_bn254::register()` installs the provider and engines at startup, and the spine looks them up at runtime.

This registry, and the `DynFieldSponge` boxing it needs, is **temporary scaffolding** — it exists only because the spine is field-free but not yet generic. The follow-up replaces it with a compile-time field type parameter. Everything transitional is listed in #456 so a reviewer can see exactly what stays and what goes.

## Why a separate `provekit-noir` crate

The Noir/mavros frontend pulls in the whole Noir toolchain (`acir`, `noirc_abi`, `mavros-vm`/`-artifacts`). With those types in `common`, the entire spine inherits the toolchain transitively — so the verifier and the field-agnostic prover engine compile it even though neither parses Noir. Pulling the frontend into `provekit-noir` severs that: the spine's `cargo tree` carries no acir/nargo, the verifier links only the `NoirProof` type, and a non-Noir build (e.g. the future Goldilocks backend) never compiles the toolchain at all.

## Why no generics yet

Threading a `<P: ProofField>` type parameter through the spine is the real end goal, but doing it in this PR would put the byte-for-byte compatibility claim at risk. Generic type parameters perturb `serde` output — derive macros add type bounds, and type names can leak into the serialized scheme that seeds the Fiat-Shamir domain separator. That is exactly the surface that must stay identical for already-deployed verifiers and mobile clients.

So this PR keeps a single `Identity<Fr>` instantiation point and stays provably byte-identical to v2 (empty diff on the R1CS-construction path + a pinned regression test). The generics work then lands as a separate, separately-verifiable change. The split keeps each PR small and each compatibility claim easy to check.

## What changed

- `common`: field-agnostic spine; new `FieldHashProvider` registry and `DynFieldSponge`; `FieldElement` derived from a single `Identity<Fr>` instantiation point. Drops the ntt/poseidon2/skyscraper deps **and** the acir/noirc_abi/mavros frontend deps (the Noir frontend types moved to `provekit-noir`).
- `provekit-prover`: field-agnostic WHIR engine; no field/acir/nargo deps. It still carries the non-WASM mavros proving path, so it isn't fully frontend-agnostic yet — that isolation is a follow-up.
- `provekit-verifier`: field-agnostic; the caller registers the backend.
- `provekit-field-bn254` (new): Skyscraper/Poseidon2/NTT, the EC/bigint/witness-hint code, the `FieldHashProvider` impl, and `register()`. The bn254 public-input KATs live here.
- `provekit-noir` (new): the Noir/mavros frontend — scheme types (`NoirProofScheme`, `NoirProof`, the prover/verifier structs), the `Prove` impls, witness solving, and input glue. The verifier depends on it for `NoirProof` only.
- Tooling (cli/ffi/bench/wasm) rewired to `provekit-noir`; standalone-verify paths and `pk_init` call `register()`.

## Verification

- e2e compiler suite: 73/73 pass.
- Field KATs, relocated unit tests, and MSM witness-solving tests pass.
- `cargo tree`: common, prover, and verifier link no field crate.
- bn254 is unchanged from v2: the R1CS-construction path has an empty diff against v2, and a regression test in `provekit-bench` pins basic-4's R1CS hash and proof-structure params.

WHIR ZK proofs use `thread_rng` blinding and are non-deterministic by design, so the regression test pins **deterministic** artifacts (serialized scheme, domain separator, `r1cs.hash()`, public-input binding) and checks proofs functionally rather than byte-comparing them.

## Follow-ups (tracked in #456)

- Make the spine generic over the field and add a Goldilocks backend. This removes the runtime registry and `DynFieldSponge`, relocates the residual 256-bit helpers still in `common`, and genericizes the monomorphic serde helpers. #456 has the full temporary-vs-permanent list.
- Make the prover engine fully frontend-agnostic by lifting the mavros proving path (`prove_mavros`) out of the WHIR engine — the last frontend coupling left in `provekit-prover`.
- The relocated bigint/EC/NTT code carries pre-existing clippy warnings (same as on v2); a focused cleanup is a separate task.


shreyas-londhe:
Solid carve. Verified bn254 is byte-identical to v2: the boxed sponge forwards to the same `DuplexSpongeInterface` calls, and the public-input KATs were moved verbatim, not regenerated. Hash-via-registry is the right call too. whir dispatches its own Merkle hashes through `ENGINES`/`EngineId`, and hash config is runtime data, so a registry fits.

One architectural gap holds this back as a base for the Goldilocks work, plus a few panics to clean up.

**Make the algebra generic over `Embedding`. Field-agnostic hashing isn't enough.**

The spine still pins one concrete `FieldElement` (`common/src/lib.rs`: `ProvekitEmbedding = Identity<Fr>`, `FieldElement = ::Source`) and uses it for both committed data (coeffs/w1/w2 in `interner.rs`, `r1cs.rs`, `sparse_matrix.rs`) and challenges (`verifier/src/whir_r1cs.rs`). Goldilocks needs those to be different types: committed data in `M::Source` (base, `Field64`), challenges in `M::Target` (ext, `Field64_3`). A single alias can't carry both. So the flip to `Basefield<Field64_3>` re-threads the whole spine anyway, and the `Identity`-vs-`Basefield` differential test is impossible, since one global alias plus a one-slot provider can't hold two fields in one binary.

The fix is additive:

- Keep the hash registry. whir does the same.
- Make the math (sumcheck, covectors, sparse-matrix) generic over `<M: Embedding>`. Committed data becomes `M::Source`, challenges `M::Target`, base×ext products go through `mixed_mul`/`mixed_dot`. Instantiate `Identity<Fr>` for bn254; output stays byte-identical.
- `whir::Config<M: Embedding>` is already generic, so this matches the dep. Our wrappers calling whir want the same `M`.

This is how the references layer it. whir: `Config<M>` for algebra, separate `ENGINES` registry for hashing. Plonky3: `StarkGenericConfig` bundle at the `prove`/`verify` entry, AIR and field math generic over `F: Field` underneath. None of arkworks, whir, or Plonky3 put a hash method on the field trait.

The earlier `impl ProofField for Fr` hit the orphan rule (E0117: both the trait and `Fr` are foreign to `field/bn254`), which is why the registry replaced it. Fine. If a per-field trait comes back later, impl it for a local marker (`struct Bn254; impl ProofField for Bn254 { type Embedding = Identity<Fr>; }`), same as arkworks `FrConfig`, Plonky3 `BabyBearParameters`, whir's own `FConfig64`. The math doesn't need that trait though. Generic over `Embedding` is leaner.

**Panics in library paths. Fix before merge.**

- `common/src/field/mod.rs`: `provider()` is `.expect()`, so every Skyscraper/Poseidon2 path panics if `register()` wasn't called. Return a `Result` and propagate.
- `verifier/src/lib.rs`: `verify()` dropped self-registration. A verify-only caller that skips `register()` panics. Tests survive only because the prover registers first in-process. Re-register or return `Err`.
- `field/bn254/src/lib.rs`: `field_sponge()` is `unreachable!()` on non-native configs, but it's a public trait method. Narrow the input type or return `Option`.
- `common/src/field/mod.rs`: `register_field_hash_provider` does `let _ = OnceLock::set(...)`, silently dropping a conflicting second registration. Once a second field crate exists that's a silent wrong-field bug. `debug_assert!` or panic on conflict.

**Scope.** Ship this as the modularization plus hash registry. That part's good. Track the `Embedding` genericization as the next PR. That's the piece that carries base/ext.

shreyas-londhe:
Closing in favor of #458.

This PR landed the decomposition: field-free spine plus hash registry, with bn254 verified byte-identical to main. The review here concluded that the registry and `DynFieldSponge` are transitional, and that the real blocker for Goldilocks is making the algebra generic over `Embedding`, with committed data in `M::Source`, challenges in `M::Target`, and _base×ext_ products through `mixed_mul`. Tracked in #456.

#458 folds that genericization into the decomposition in one pass, so the registry never gets built.

Thanks @BornPsych.

PR #458, opened 2026-06-23
refactor: make the proving spine generic over the field (`<P: ProofField>`); extract provekit-backend-bn254

### LINKED ISSUE #456: Make the proving spine generic over the field (retire the runtime registry from #455)
## What this tracks

#455 made the proving spine (`provekit-common`/`-prover`/`-verifier`) **field-agnostic** by moving all bn254 code into `provekit-field-bn254` and wiring the field in through a **runtime registry** (`FieldHashProvider`) plus a boxed sponge (`DynFieldSponge`). That was done on purpose, to keep bn254 **byte-for-byte identical to v2**.

The next step is the real goal: make the spine **generic over the field** — one type parameter `P` that names a base field and an extension field — so a second field (Goldilocks) can be plugged in. Doing that lets us **delete** the temporary registry/boxing from #455 and resolve the field at compile time instead of at runtime. This issue lists what's temporary, what stays, and why we split it out.

## Why we didn't do generics in #455

Threading a `<P>` type parameter through the spine perturbs `serde` output: derive macros add type bounds, and type names can leak into the serialized scheme that seeds the Fiat-Shamir domain separator. That's exactly the surface that must stay identical for already-deployed verifiers and mobile clients. So #455 keeps a single `Identity<Fr>` instantiation point and stays provably byte-identical to v2 (empty diff on the R1CS-construction path + a pinned regression test). The generics change then lands separately, where its compatibility can be verified on its own.

## What's temporary (introduced by #455, removed/replaced by the generics work)

| Temporary thing | Where | Replaced by |
|---|---|---|
| Runtime `FieldHashProvider` registry — the `OnceLock`, `register_field_hash_provider`, `provider()` (panics if unregistered), and the "must call `register()` first" precondition at every entry point | `common/src/field/mod.rs` (+ `register()` call sites in prover/verifier/r1cs-compiler/ffi) | Compile-time `P::method()` dispatch — no global, no precondition |
| `DynFieldSponge` — the boxed `dyn` sponge in `TranscriptSponge`, needed only because the spine no longer names a concrete sponge type | `common/src/field/mod.rs`, `common/src/transcript_sponge.rs` | The sponge resolves through the field type parameter; no boxing / type erasure |
| The single `ProvekitEmbedding = Identity<Fr>` / `FieldElement` alias (the one bn254 instantiation point) | `common/src/lib.rs` | `<P: ProofField>` with `Base<P>` (committed data) and `Ext<P>` (challenges) |
| Monomorphic serde helpers hardcoded to `Vec<FieldElement>` / `Option<FieldElement>` | `common/src/utils/serde_ark_vec.rs`, `serde_ark_option.rs` | Generic over `<T>` under `#[serde(bound = "")]` (must stay byte-identical for `T = Fr`) |
| Residual 256-bit / `PrimeField`-width code still in the spine (field-free at the crate level, but still assumes bn254's field width) | `common/src/witness/digits.rs` (`BitIteratorLE<BigInt<4>>`), `common/src/utils/mod.rs` (`HALF`, `uint_to_field`, byte bridges), `common/src/hash_config.rs` (digest→field reduction, `PUBLIC_INPUTS_DST_FE`) | Move into the field crate / behind a `FieldHash` trait method |
| WHIR scheme aliases hardcode `Identity<Fr>` | `common/src/whir_r1cs.rs` (`WhirConfig`, `WhirZkConfig`, `WhirR1CSScheme`) | `WhirR1CSScheme<P>` with `WhirConfig<P>`/`WhirZkConfig<P>` (parameterize this first) |
| Dev-dependency cycle so the spine's own tests can use a concrete field | `common/Cargo.toml` (dev-dep on `provekit-field-bn254`) | Spine tests instantiate `P` directly once the field is a type parameter |

## What stays (permanent — #455's lasting value)

- The crate split: `provekit-field-bn254` and `provekit-noir`.
- The field-free spine dependency graph (`common`/`prover`/`verifier` link no field crate).
- The relocated bn254 code: EC/bigint/digit-decomposition solvers, NTT, Skyscraper/Poseidon2 — all in the field crate.
- The relocated Noir/mavros frontend in `provekit-noir`.
- The byte-identical regression gate — becomes the standing compatibility gate.
- Each backend's `register()` keeps its NTT + ENGINES inserts: WHIR's `NTT`/`ENGINES` are pre-existing global registries populated at startup; generics makes ProveKit's **field choice** compile-time but does not remove WHIR's own startup registration.

## Related

- #419 (byte-injective transcript sponge adapter) touches the same sponge but is a separate concern.

Introduced by #455.

shreyas-londhe:
## Summary

ProveKit is currently welded to a single field. Every layer of the proving spine, `provekit-common`, `provekit-prover`, and `provekit-verifier`, refers to one concrete `FieldElement` (BN254's `Fr`). This PR makes the spine generic over the field and moves all BN254-specific code into a new `provekit-backend-bn254` crate. The field a binary proves over is now decided by which backend crate it links, not by a type baked into the spine. BN254 output is unchanged, byte for byte.

## Why

The next milestone is a Goldilocks backend, and Goldilocks cannot be expressed as a drop-in replacement for `Fr`. It is a 64-bit prime, too small to draw Fiat-Shamir challenges from safely, so a sound protocol has to keep committed data (witness, matrix coefficients) in the 64-bit base field while drawing challenges from a degree-3 extension. BN254 never forced this distinction because its base and extension are the same type, so the codebase collapsed the two into one `FieldElement` alias and used it everywhere: committed data in `interner`, `r1cs`, `sparse_matrix`, and challenges in the verifier and sumcheck.

A single alias cannot carry two different types. So supporting any small field means teaching the spine the difference between "the field we commit in" and "the field we sample challenges in." That is the core of this refactor. It is also why a narrower change (just making hashing field-agnostic) would not have been enough: the algebra itself has to distinguish base from extension.

## What changed

The base/extension split is modeled with WHIR's `Embedding` (`Source` is the base field, `Target` the extension). Two traits in `common` sit on top of it:
- `ProofField` names the embedding for a backend. `Base<P>` and `Ext<P>` resolve to the committed and challenge fields.
- `FieldHash: ProofField` supplies the transcript sponge, public-input hashing, and the byte bridges, kept off the algebra surface so the math traits stay clean.

The algebra is then generic over the embedding. `R1CS<F>`, `Interner<F>`, the sparse-matrix types, the covectors, and the sumcheck utilities carry the type parameter; committed data is `M::Source`, challenges are `M::Target`, and the mixed base×ext products that show up in binding and sumcheck go through the embedding's `mixed_mul`. The prover and verifier are threaded with `<P: ProofField>`.

All BN254-specific code moves out of the spine into `provekit-backend-bn254`: the field definition, the Skyscraper/Poseidon2/NTT WHIR hash adapters, the witness solver, the Noir/mavros frontend and proving path, and public-input hashing (with its known-answer tests). The crate implements `ProofField`/`FieldHash` for a local `Bn254Field` marker with `Embedding = Identity<Fr>`, where base and extension coincide. Tooling (cli, ffi, bench, wasm, gnark, verifier-server) is rewired to the new layout.

## How fields plug in now

The dependency arrow points one way: a backend crate depends on the spine and supplies the concrete field; the spine names no field and pulls in no field crate or Noir toolchain. `cargo tree` for `common`, `prover`, and `verifier` carries no `ark-bn254`, no `acir`, no `nargo`.

Adding a field is additive. Write a backend crate, define a marker type, and implement `ProofField`/`FieldHash` for it. No spine edits. Field selection is resolved at compile time through the type parameter rather than a runtime registry, so there is no dispatch cost on the hot path, and mixing two fields in one proof is a type error rather than a runtime surprise. For BN254 the embedding is the identity, the generic code monomorphizes to exactly what shipped before, and the output is provably unchanged.

## Compatibility

BN254 stays byte-identical:
- Public-input binding loop unchanged: `expected` starts at the position-0 constant `1`, folds inputs at `1..=N`, so `n = num_public_inputs + 1` holds. `mixed_mul` is ordinary multiplication for `Identity<Fr>`.
- Fiat-Shamir tag and sponge unchanged: the BN254 `FieldHash` impl delegates to the existing `hash_field_elements`, `field_to_bytes_le`, and `TranscriptSponge::from_config`. The byte-exact KATs are retained as a regression gate.
- Every prover message stays transcript-bound (`prover_message`, `ProverState<P::Sponge>`); none moved to a hint.
- Serialized formats are wire-compatible: generic types use `#[serde(bound = "")]` and add no data fields, and the one `PhantomData` added lives in a serde deserializer visitor with no wire footprint. For BN254 `Ext<P> = Base<P> = FieldElement`, so postcard and JSON output match.

No Goldilocks code is wired into the proof path. `Field64` appears only in `#[cfg(test)]` modules, where it exercises the now-generic code over a second field and pins that the abstraction holds.

## Verification

`fmt`, `clippy`, `build`, `test` (425 passed, 0 failed), `--doc`, and `cargo doc -D warnings` all pass with `--all-features`. The Go recursive-verifier is untouched.

## Follow-ups

Add the Goldilocks backend: a marker with `Embedding = Basefield<Field64_3>`, committed data in `Field64`, challenges in `Field64_3`. The seam this PR installs is the whole enabling piece, so the backend is a new crate plus an instantiation, with no further spine re-threading.

Closes #456.


veljkovranic (changes requested):
This command is failing on this branch:

`cargo check -p provekit-wasm --target wasm32-unknown-unknown`

Seems like the imports have been mixed up a bit.

Also, I am not sure how we are going to version the docs that live here, but this example
`docs/src/content/docs/e2e/rust.mdx`
would become stale if we merge right now.

I know that website targets v1, but we should probably think about the docs versioning right now. 

Tagging @dcbuild3r for input

shreyas-londhe:
@veljkovranic pushed a fix for the wasm command.

Regarding the docs - `rust.mdx` is pinned to crates.io `1.0.0` and this PR doesn't edit it, so merging changes nothing about what the docs show — the example still resolves against `1.0.0`. So I don't think docs staleness should block this PR. 

The real issue: `Prover`/`Verifier` move out of `provekit_common` into `provekit-backend-bn254`, which is breaking — it needs a `2.0.0` bump, and the docs rewrite rides with that release.

provekit/prover/src/whir_r1cs.rs
```diff
@@ -571,18 +517,19 @@ pub fn run_zk_sumcheck_prover(
...
 
-        let mut combined_hhat_i_coeffs = [FieldElement::zero(); 4];
+        let mut combined_hhat_i_coeffs = [F::zero(); 4];
 
         combined_hhat_i_coeffs[0] = hhat_i_at_0 + rho * g_poly[0];
 
         let g_at_minus_one = g_poly[0] - g_poly[1] + g_poly[2] - g_poly[3];
         let combined_at_em1 = hhat_i_at_em1 + rho * g_at_minus_one;
 
-        combined_hhat_i_coeffs[2] = HALF
```
veljkovranic, line 581: Is there a chance to make these new generic fields expose the value of HALF, so that we can avoid division?

veljkovranic: If not, we can just compute it ourselves, outside of the for loop

provekit/common/src/field.rs
```diff
@@ -0,0 +1,56 @@
...
+
+/// Base (committed) field of a [`ProofField`].
+pub type Base<P> = <<P as ProofField>::Embedding as Embedding>::Source;
+
+/// Extension (challenge) field of a [`ProofField`].
+pub type Ext<P> = <<P as ProofField>::Embedding as Embedding>::Target;
+
+/// Hash and byte-bridge glue, kept out of [`ProofField`]'s algebra surface and
+/// composed as a supertrait so [`Ext<Self>`] is nameable.
+pub trait FieldHash: ProofField {
```
zkfriendly, line 30: `default_hash()`, `ext_from_bytes()`, and `from_digest()` have no call sites anywhere in the spine or tooling — only the Bn254Field impl and its own roundtrip test. Wondering why we need them now?

provekit/common/src/compress.rs
```diff
@@ -0,0 +1,61 @@
...
+/// then decompresses when the sumcheck begins.
+pub struct CompressedR1CS {
+    num_constraints: usize,
+    num_witnesses:   usize,
+    num_virtual:     usize,
+    blob:            Vec<u8>,
+}
+
+impl CompressedR1CS {
+    pub fn compress<F: Field>(r1cs: R1CS<F>) -> Result<Self> {
```
zkfriendly: `compress<F>` and `decompress<F>` are now independently typed on a raw byte blob so `compress::<Fr>(...)` followed by `decompress::<Field64>()` type-checks and only fails (or worse, misparses) at runtime deserialization. Making the struct `CompressedR1CS<F>` with a `PhantomData<F>` would make the round-trip a compile-time guarantee for free.

zkfriendly (commented):
Great job! I like how lean is the prover crate is now.

provekit/prover/src/whir_r1cs.rs
shreyas-londhe: `ark_ff::Field` doesn't expose a HALF constant, so went with the second option.

provekit/common/src/field.rs
shreyas-londhe: Fair catch — these were added anticipating the goldilocks backend, but even the follow-up branches only implement them without calling them from the spine. Removed all three in 6e96519c; will re-add alongside their first real call site.

provekit/common/src/compress.rs
shreyas-londhe: Good idea — done in 4b3b5c1c. `CompressedR1CS` now carries the field as a phantom type parameter, so a mismatched round-trip fails to type-check.

zkfriendly: btw, for v1 we merged https://github.com/worldfnd/provekit/pull/438 where compression/decompression was explicitly removed. So I think we might even be able to remove this completely?

shreyas-londhe: Fair argument, but the scope of this PR is making the proving spine generic, and deleting a component would be scope creep. I'd suggest we table this for now and remove when we get into profiling phase.

veljkovranic (approved):

PR #460, opened 2026-06-23
feat: add Goldilocks backend + two-field differential fixtures

BornPsych:
## Summary

Adds `provekit-backend-goldilocks`, a second backend that instantiates the field-generic spine over the ~192-bit Goldilocks cubic extension `Field64_3`, and restructures `provekit-fixtures` so the prove→verify and soundness suites run over **both** bn254 and goldilocks from shared, field-generic test bodies. **bn254 is unchanged.**

This is the backend half of the field-generic work: #458/#459 made the spine generic over `<P: ProofField>` and added the scheme builder + fixtures; this PR shows a second, structurally different field plugs in with no spine edits.

## Why

The spine was made field-generic precisely so a non-bn254 field could be added as a backend crate plus an instantiation, with no changes to `common` / `prover` / `verifier`. Goldilocks is the validation field: a 64-bit prime whose cubic extension `Field64_3` is large enough (~192 bits) to draw Fiat-Shamir challenges from. This PR is the proof that the seam holds — the backend is a marker type implementing `ProofField` / `FieldHash` plus a `register()` call, nothing more. (Pre-EF/BF the embedding is `Identity<Field64_3>`, so base == ext for now; the base≠ext split lands once zkWHIR supports it.)

## What changed

- **`provekit-backend-goldilocks`** (new): `GoldilocksField` with `Embedding = Identity<Field64_3>`; a `FieldHash` impl with a `Sha256` default hash, 24-byte little-endian canonical bytes, digest-spread hashing, and a transcript sponge; `register()` installs the `Field64_3` NTT engine in whir's global registry. Mirrors the `provekit-backend-bn254` layout.
- **`provekit-fixtures`**: the bn254-only `roundtrip.rs` / `soundness.rs` are replaced by `tests/shared/mod.rs` — field-generic `<P: FieldHash>` test bodies plus `roundtrip_suite!` / `soundness_suite!` macros — and four thin per-field instantiation files (`bn254_*`, `goldilocks_*`). The same assertions now run over both fields.
- **Workspace**: registers the new crate as a member + workspace dep, and adds it as a fixtures dev-dep.

## Verification

- `cargo build --workspace` green.
- `provekit-fixtures`: bn254 6 roundtrip + 4 soundness, goldilocks 6 roundtrip + 4 soundness — all pass; 2 benches ignored.
- `provekit-backend-goldilocks` unit tests (byte roundtrip) pass.

---
Stacks on #459 (open); until that merges, the commit list above includes #459's commits — the Goldilocks-specific changes are the final 2 commits.


provekit/backend/goldilocks/src/field_hash.rs
```diff
@@ -0,0 +1,198 @@
...
+    whir::algebra::fields::{Field64, Field64_3},
+};
+
+/// Domain-separation tag for public-input instance binding.
+///
+/// **Protocol-visible constant.** Absorbed into the SHA-256, Keccak, and BLAKE3
+/// hashes used for public-input commitments; changing it invalidates every
+/// proof generated under those configurations. The `V1` suffix reserves an
+/// unambiguous upgrade path (`_V2`, …) for any future construction change.
+const PUBLIC_INPUTS_DST: &[u8] = b"PROVEKIT_PUBLIC_INPUTS_V1";
```
shreyas-londhe: **nit**: This should be `_V2`

tooling/provekit-fixtures/tests/shared/mod.rs
```diff
@@ -0,0 +1,419 @@
...
+
+    // A structurally different instance (more constraints/witnesses).
+    let (r1cs_b, _) = squaring_chain::<Base<P>>(3, 12);
+    assert!(scheme.verify(&proof, &public_inputs, &r1cs_b).is_err());
+}
+
+/// Public-input binding: a verify against wrong public inputs must reject at
+/// both `N = 1` (trivial binding) and `N = 2` (non-trivial binding loop), even
+/// though the R1CS itself is satisfied.
+pub fn public_input_binding_mismatch_is_rejected<P>()
```
shreyas-londhe, line 240: **question (blocking)**: this test proves with the correct inputs and verifies with the wrong ones, so the verifier rejects at the "Public inputs hash mismatch" guard before `verify_public_input_binding` ever runs. It covers instance binding, but the binding covector's reject path is now untested. The deleted `soundness.rs` proved and verified with the same wrong inputs, which got past the hash guard and exercised the covector math (the PR #321 path). Can we restore that shape here, at both N=1 and N=2?

provekit/backend/goldilocks/src/bytes.rs
```diff
@@ -0,0 +1,37 @@
...
+/// Deserializes an extension element from up to 24 little-endian bytes,
+/// reducing each 8-byte coefficient mod the base prime (total — never fails,
+/// matching the prime-field `from_le_bytes_mod_order` contract). A short input
+/// is zero-padded; bytes beyond the 24th are ignored.
+#[inline]
+pub(crate) fn bytes_to_field(bytes: &[u8]) -> Field64_3 {
+    let mut buf = [0u8; EXT_BYTES];
+    let n = bytes.len().min(EXT_BYTES);
+    buf[..n].copy_from_slice(&bytes[..n]);
+    let c0 = Field64::from_le_bytes_mod_order(&buf[0..8]);
```
shreyas-londhe: **nit**: each coord decodes via `from_le_bytes_mod_order`, so the 2^32 − 1 encodings in [p, 2^64) silently alias to canonical values. Harmless while this is test-only, but it needs a `< p` rejection check before it ever sits on a real deserialization path.

shreyas-londhe (changes requested):

tooling/provekit-fixtures/tests/shared/mod.rs
BornPsych: Yes, we can do that here. I don't see any blocker with goldilocks introduction, Will add that.

zkfriendly (approved):

provekit/backend/goldilocks/src/field_hash.rs
BornPsych: Done

provekit/backend/goldilocks/src/bytes.rs
BornPsych: Done

tooling/provekit-fixtures/tests/shared/mod.rs
BornPsych: Added those test back, now we test both `hash mismatch` and `binding covector` in rejected path

shreyas-londhe (approved):
LGTM

PR #462, opened 2026-06-26
chore: gate Vercel deployments to docs-only changes

BornPsych:
## Problem

Vercel fires on every PR and push regardless of whether `docs/` changed. On non-docs PRs produces a spurious `Vercel: FAILURE` status check — confusing reviewers and wasting build minutes on commits that have nothing to do with the docs site.

## Solution

Two inert files — no `vercel.json`, no jobs, no secrets needed:

- **`.vercelignore`** — ignore everything except `docs/**`
- **`.github/workflows/docs-gate.yml`** — path-filter workflow; GitHub
  only dispatches the Vercel webhook when `docs/**` files change

The workflow has no jobs; it exists purely as a filter gate.


BornPsych:
Closing this — after investigation, the deploy-on-every-PR behavior is a Vercel **project setting** (dashboard) and cannot be controlled from repo files: GitHub Actions `paths:` filters do not gate the Vercel GitHub App, and a root `.vercelignore` either no-ops or breaks the docs build (Root Directory is `docs/`). Tracking the real fix — a Vercel "Ignored Build Step" — in #463.

PR #469, opened 2026-07-08
ci: field-separation guard + per-backend standalone builds

BornPsych:
## Summary

Adds three CI jobs that enforce two invariants the existing `--all-features` build
silently masks: that the 64-bit Goldilocks backend never links a bn254/256-bit
dependency, and that each backend (plus its welded tooling and fixtures) still builds
and tests in isolation. No production code changes.

## Why

#458–#460 made the proving spine field-generic and added a second backend
(`provekit-backend-goldilocks`) whose entire point is a pure 64-bit dependency graph.
Two things now need machine enforcement:

- **Field separation** — a single accidental dependency on a bn254/256-bit crate (a
  curve stack, or the Noir/Mavros frontend) would defeat the separation the backend split
  was built for, and a monolithic `--all-features` build would never catch it.
- **Standalone buildability** — a backend that only compiles with every feature enabled is
  not actually standalone. These jobs catch a backend or its tooling that stops building on
  its own.

## What changed

- **`scripts/check-field-separation.sh`** (new, fail-closed): captures each backend's
  `cargo tree` and (1) allowlists the workspace-local crates the Goldilocks graph may
  contain (`provekit-common`, `-prover`, `-verifier`, and the goldilocks backend itself),
  (2) denylists external bn254/256-bit crate families (`ark-bn254`, `ark-ec`, `k256`,
  `acir`, `mavros-*`, …) as defense-in-depth, and (3) aborts if `cargo tree` errors or
  returns an empty graph — a guard that cannot inspect the graph must not pass.
- **`.github/workflows/ci.yml`**: three new jobs —
  - `field_separation` runs the guard script.
  - `goldilocks_standalone` builds + tests `provekit-backend-goldilocks` with no
    `--all-features`, plus its fixtures.
  - `bn254_standalone` builds + tests `provekit-backend-bn254` and its welded tooling
    (`r1cs-compiler`, `cli`, `ffi`, `gnark`, `bench`, `verifier-server`,
    `passport-input-gen`), plus its fixtures.

## Verification

- `./scripts/check-field-separation.sh` → `OK: field-backend dependency separation holds.`
  (exit 0) against current `main`.
- `ci.yml` parses as valid YAML; all three jobs registered.

## Commits

- ci: per-backend standalone builds + field-separation guard

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,3 +96,60 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: crate-ci/typos@v1.44.0
+
+  field_separation:
+    name: Field-backend separation guard
+    runs-on: ubuntu-24.04-arm
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup rust toolchain, cache and bins
+        uses: moonrepo/setup-rust@v1
+        with:
+          channel: nightly-2026-03-04
+          cache-base: main
+      - name: Check field-backend dependency separation
+        run: ./scripts/check-field-separation.sh
+
+  goldilocks_standalone:
+    name: Goldilocks backend (standalone, no --all-features)
+    runs-on: ubuntu-24.04-arm
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup rust toolchain, cache and bins
+        uses: moonrepo/setup-rust@v1
+        with:
+          channel: nightly-2026-03-04
+          cache-base: main
+      - name: Build goldilocks backend in isolation
+        run: cargo build --all-targets -p provekit-backend-goldilocks
+      - name: Test goldilocks backend in isolation
+        run: cargo test --no-fail-fast --lib --tests --bins -p provekit-backend-goldilocks
+      - name: Test goldilocks fixtures
+        run: cargo test -p provekit-fixtures --test goldilocks_roundtrip --test goldilocks_soundness
+
+  bn254_standalone:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -96,3 +96,60 @@ jobs:
...
+      - uses: actions/checkout@v4
+      - name: Setup rust toolchain, cache and bins
+        uses: moonrepo/setup-rust@v1
+        with:
+          channel: nightly-2026-03-04
+          cache-base: main
+      - name: Check field-backend dependency separation
+        run: ./scripts/check-field-separation.sh
+
+  goldilocks_standalone:
```
zkfriendly, line 113: `provekit-fixtures` has `provekit-backend-bn254` as a dev-dependency, so `cargo test -p provekit-fixtures --test goldilocks_roundtrip` ... compiles the entire `bn254` stack (acir, noirc, mavros, skyscraper, ntt) inside the "goldilocks standalone" job. The two earlier steps are still valid isolation checks since they're separate cargo invocations, but this makes the job much slower than its name suggests and means a green run here doesn't imply the goldilocks fixtures work without the `bn254` stack present in the build.

Suggestion: move both fixture-test steps (goldilocks and bn254) into one shared fixtures job, and keep the two standalone jobs to pure backend build+test. Same coverage, and the standalone jobs get much cheaper.

BornPsych: good suggestion, really appreciated

shreyas-londhe (approved):
LGTM.

PR #470, opened 2026-07-08
feat: commit the R1CS witness over the base field (proof format 2.0)

BornPsych:
## Summary

Commits the R1CS witness over the **base** field and moves the Spartan blinding polynomial `g` to a separate **extension**-field commitment, instead of committing the whole witness (padded with the blinding coefficients) in the extension field. The proof format is bumped to **2.0** and tagged per field so one field's proof is rejected by another field's verifier. bn254 and Goldilocks both ride the new uniform non-ZK base-commit path.

Builds on the field-generic spine + two-backend work (#458/#459/#460). Independent of the CI guard PR (#469).

## Why

Goldilocks is a 64-bit prime: its whole value is that committed data (witness, matrix coefficients) stays in the base field `Field64`, with only challenges/sumcheck in the degree-3 extension `Field64_3`. The old scheme committed the witness in the extension and padded `w1` with room for the blinding polynomial — that forces base data through extension-width commitments and couples witness sizing to blinding room. Splitting the two commitments lets the witness commit natively in the base field and sizes the blinding commitment to just its `4 * m_0` cubic coefficients.

## What changed

- **`provekit-common`** — `WhirR1CSScheme` gains `whir_witness: GenericWhirConfig<P::Embedding>`   (base-field, non-ZK) and a separate `whir_blinding: GenericWhirConfig<Identity<Ext<P>>>`; new `new_witness_config_for_size` / `new_blinding_config_for_size` / `whir_protocol_params`; the old `4 * m_0` `w1`-padding is gone. `.np` format magic is derived from `ProofField::FIELD_ID` (`np_format`); all format versions bump to `2.0` (major must match on read). Covector eval helpers are embedding-aware (base × ext).
- **`provekit-prover`** — commits the base-field witness with `whir_witness` (no base→ext lift) and commits `g` with the separate `whir_blinding`; `BlindingState` carries `g`'s own vector + WHIR witness; `g` is opened by a dedicated `whir_blinding.prove` after the witness opens.
- **`provekit-verifier`** — receives the ext blinding commitment right after the witness commitment (matching prover transcript order), verifies the witness with the two-step `whir::Config` API, and verifies `g` by a dedicated open at the end.
- **Backends** — bn254 drops the zkWHIR config for the same non-ZK path (`Identity<Fr>`, `FIELD_ID 0`, `.np` magic byte-identical); Goldilocks switches to `Basefield<Field64_3>`  (`FIELD_ID 1`). Adds `GoldilocksEfField` (`Identity<Field64_3>`, `FIELD_ID 2`) as a stand-in for challenge-bearing fixtures.
- **Tooling / tests** — cli + verifier-server pass `whir_witness` + `whir_blinding` to the gnark writer; r1cs-compiler security tests check the two new configs (dimension test drops to `m = 13`); fixtures route base vs challenge suites to the right field.

## Caveats

- **`GoldilocksEfField` is a stand-in.** A single `Field64` challenge is only ~64-bit sound, so challenge-bearing fixtures run on the ext-leaf field until a base transcript codec + k-fold repetition land (see the `TODO` in `goldilocks/src/field.rs`).
- **Witness commitment is non-hiding** — WHIR openings leak witness values, so this provides sumcheck ZK only, not witness ZK (`whir_witness` `TODO`).
- **Format 2.0 is a breaking change** — proofs are not compatible with 1.x verifiers.

## Out of scope (tracked separately)

This PR changes only the Rust WHIR proving/verifying path. It **intentionally does not** touch:

- the Go `recursive-verifier/` (which still parses the old fused zkWHIR commitment layout), or
- the `provekit-gnark` bridge / Groth16 prover path.

Consequence: the **WHIR → `recursive-verifier` (Go) → Groth16** on-chain route will not verify 2.0 proofs until a paired Go update lands, tracked as follow-up. The **direct Groth16+BSB22 backend (#427 / #447)** is a separate crate and is unaffected by this change (the two only share a few tooling files). This keeps the PR focused on the Rust proof-system change; the recursive-verifier port is deliberately left for a coordinated follow-up.

## Verification

- `cargo fmt --all --check`, `cargo clippy --all-targets --all-features` (no new errors), `cargo check --all-targets --all-features` green.
- Fixtures over **both** fields: bn254 7 roundtrip + 6 soundness, goldilocks 7 roundtrip + 6 soundness — all pass.
- `provekit-r1cs-compiler` 41 tests pass (security + dimension, split config).
- Core libs: common 45, prover 8, verifier 83, bn254 45, goldilocks 8; doc tests pass.

## Commits

9 authored commits (base-field split across common / prover / verifier / backends / tooling / fixtures) plus one commit reconciling the base-sampling bound and rustfmt onto current `main`.

### DIFF
--- a/provekit/common/src/whir_r1cs.rs
+++ b/provekit/common/src/whir_r1cs.rs
@@ -259,13 +259,45 @@ pub struct ProvekitProof<P: ProofField> {
     pub whir_r1cs_proof: WhirR1CSProof,
 }
 
+/// Formats whose magic `.np` must never alias.
+#[cfg(not(target_arch = "wasm32"))]
+const NON_NP_FORMATS: [[u8; 8]; 3] = [
+    binary_format::PROVER_FORMAT,
+    binary_format::VERIFIER_FORMAT,
+    binary_format::NOIR_PROOF_SCHEME_FORMAT,
+];
+
+#[cfg(not(target_arch = "wasm32"))]
+const fn formats_eq(a: [u8; 8], b: [u8; 8]) -> bool {
+    let mut i = 0;
+    while i < 8 {
+        if a[i] != b[i] {
+            return false;
+        }
+        i += 1;
+    }
+    true
+}
+
 /// Derive the `.np` format magic from [`ProofField::FIELD_ID`] by offsetting
 /// the final magic byte: bn254 (id 0) keeps the historical magic, other fields
 /// a distinct one.
+///
+/// The offset is a bijection on `u8`, so distinct ids always yield distinct
+/// magics — at the cost of reserving all 256 values of the final byte.
 #[cfg(not(target_arch = "wasm32"))]
 const fn np_format(field_id: u8) -> [u8; 8] {
     let mut f = binary_format::NOIR_PROOF_FORMAT;
     f[7] = f[7].wrapping_add(field_id);
+
+    let mut i = 0;
+    while i < NON_NP_FORMATS.len() {
+        assert!(
+            !formats_eq(f, NON_NP_FORMATS[i]),
+            "FIELD_ID makes the `.np` magic collide with another ProveKit file format"
+        );
+        i += 1;
+    }
     f
 }
 
@@ -286,13 +318,26 @@ impl<P: ProofField> MaybeHashAware for ProvekitProof<P> {
 
 #[cfg(all(test, not(target_arch = "wasm32")))]
 mod tests {
-    use super::{binary_format, np_format};
+    use {
+        super::{binary_format, np_format, NON_NP_FORMATS},
+        std::collections::HashSet,
+    };
 
     #[test]
-    fn np_format_preserves_bn254_and_distinguishes_fields() {
+    fn np_format_preserves_bn254() {
         assert_eq!(np_format(0), binary_format::NOIR_PROOF_FORMAT);
-        assert_ne!(np_format(1), np_format(0));
-        assert_ne!(np_format(2), np_format(0));
-        assert_ne!(np_format(2), np_format(1));
+    }
+
+    #[test]
+    fn np_format_is_collision_free_over_every_field_id() {
+        let magics: HashSet<[u8; 8]> = (0..=u8::MAX).map(np_format).collect();
+        assert_eq!(magics.len(), 256, "FIELD_ID -> magic must be injective");
+        for other in NON_NP_FORMATS {
+            assert!(
+                !magics.contains(&other),
+                "`.np` magic space collides with {:?}",
+                std::str::from_utf8(&other)
+            );
+        }
     }
 }
--- a/provekit/common/src/whir_r1cs.rs
+++ b/provekit/common/src/whir_r1cs.rs
@@ -60,6 +60,16 @@ impl R1csHash {
     }
 }
 
+/// WHIR proving/verifying parameters for one R1CS instance, generic over the
+/// proof field `P`.
+///
+/// # Zero-knowledge
+///
+/// The ZK posture is fixed, not configurable:
+/// - Sumcheck ZK is always on: the Spartan sumcheck rounds are masked by a
+///   blinding polynomial `g`, committed separately in `whir_blinding`.
+/// - Witness ZK is off: the witness is committed non-hiding in `whir_witness`,
+///   whose WHIR openings leak witness values. It will be enabled by zkWHIR 3.0.
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 #[serde(bound = "")]
 pub struct WhirR1CSScheme<P: ProofField> {
--- a/provekit/common/src/whir_r1cs.rs
+++ b/provekit/common/src/whir_r1cs.rs
@@ -216,8 +216,7 @@ impl<P: FieldHash> WhirR1CSScheme<P> {
         m_0: usize,
         hash_id: EngineId,
     ) -> GenericWhirConfig<Identity<Ext<P>>> {
-        let nv_blind = ((4 * m_0).next_power_of_two().trailing_zeros() as usize)
-            .max(MIN_BLINDING_NUM_VARIABLES);
+        let nv_blind = next_power_of_two(4 * m_0).max(MIN_BLINDING_NUM_VARIABLES);
         GenericWhirConfig::<Identity<Ext<P>>>::new(
             1 << nv_blind,
             &Self::whir_protocol_params(hash_id),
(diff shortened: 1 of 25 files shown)


provekit/r1cs-compiler/src/whir_r1cs.rs
```diff
@@ -103,49 +103,31 @@ mod tests {
...
         );
         assert!(
             sec_blinding >= 128.0,
-            "Blinding commitment security {sec_blinding:.2} < 128 bits"
+            "Blinding commitment security {sec_blinding:.2} < 128 bits at size {size}"
         );
     }
 
+    #[test]
+    fn verify_security_level() {
```
zkfriendly, line 126: Security-level tests only cover bn254, whir's security formulas depend on field size, and Goldilocks' extension is ~192 bits vs bn254's 254 — there's no test asserting the Goldilocks witness/blinding configs hit 128 bit. Add a GoldilocksField instantiation of assert_configs_secure (it's one line now that the helper exists? )

tooling/cli/src/cmd/generate_gnark_inputs.rs
```diff
@@ -56,8 +56,8 @@ impl Command for Args {
 
         write_gnark_parameters_to_file(
```
zkfriendly, line 57: The gnark/Go tooling will produce garbage without warning. The PR does mention the Go recursive-verifier out of scope, but tooling/cli `generate-gnark-inputs` and `verifier-server` still happily write gnark parameter files from the new configs. Maybe consider a hard error (or at min a loud warning)

zkfriendly (changes requested):
PR looks great! 

One thing to bring up is the fact that bn254 path is no longer zk (I understand this is because zook not landed yet). But I was wondering, even with zook, it's worth considering making zk optional (there are use cases that don't need it, and this can make the prover faster) and being very explicit regarding whether zk is enabled or not. Currently, it is a bit odd that proof version 2 is NOT zk but 1* is.

provekit/r1cs-compiler/src/whir_r1cs.rs
BornPsych: Done, added the test for goldilock, since security is derived from the extension field the commitment opens over extension field, not the base field it commits in, so Goldilocks extension field is exercised alongside bn254

tooling/cli/src/cmd/generate_gnark_inputs.rs
BornPsych: Added the warning as well

tooling/provekit-fixtures/tests/shared/mod.rs
```diff
@@ -441,6 +452,14 @@ macro_rules! soundness_suite {
...
+    };
+}
+
+/// Challenge-bearing soundness suite (see [`challenge_roundtrip_suite!`]):
+/// requires an `Identity` field where the base and extension fields coincide.
+#[macro_export]
+macro_rules! challenge_soundness_suite {
+    ($field:ty, $register:path) => {
         #[test]
         fn tampered_public_input_is_rejected() {
```
shreyas-londhe: **minor:** `tampered_public_input_is_rejected` moved off the base-commit path

This test now lives in `challenge_soundness_suite!`, so it only runs against `GoldilocksEfField` (the ext leaf). The base-commit witness path, which is the new code this PR adds, no longer runs the dedicated public-input tamper test. `public_input_binding_mismatch_is_rejected` still covers the base suite, but only partially. Move this back into `soundness_suite!` so the base-commit path gets the same coverage.

provekit/common/src/whir_r1cs.rs
```diff
@@ -221,9 +249,19 @@ pub struct ProvekitProof<P: ProofField> {
     pub whir_r1cs_proof: WhirR1CSProof,
 }
 
+/// Derive the `.np` format magic from [`ProofField::FIELD_ID`] by offsetting
+/// the final magic byte: bn254 (id 0) keeps the historical magic, other fields
+/// a distinct one.
+#[cfg(not(target_arch = "wasm32"))]
+const fn np_format(field_id: u8) -> [u8; 8] {
+    let mut f = binary_format::NOIR_PROOF_FORMAT;
+    f[7] = f[7].wrapping_add(field_id);
```
shreyas-londhe, line 291: **nit:** `np_format` magic is collision-fragile

`np_format` derives the magic as `magic[7].wrapping_add(field_id)`. That's fine for the current ids (0, 1, 2), but it will collide silently once ids grow or wrap past 256, and it could land inside another format's byte space. A `debug_assert!(field_id < N)` or a per-field magic table would be sturdier than relying on small ids.

shreyas-londhe:
**minor:** Field-tag / version check is binary-only

The PR description says proofs are "tagged per field so one field's proof is rejected by another field's verifier," and that the 2.0 bump "fails fast." That's true on the binary path: `io/bin.rs` checks `FORMAT` and `major ==` when it parses the header. The JSON path doesn't. `read_json` (`io/json.rs:35`) is a bare `serde_json::from_reader` with no magic, version, or field check, and the `ensure!` calls in `verify()` are hash-based (`r1cs_hash`, `public_inputs_hash`), not version-based.

A cross-field or downgraded `.json` proof is caught only by the downstream transcript mismatch. That's still sound, since the domain separator serializes the whole scheme, but it isn't the explicit tag gate the description implies. Either thread the version and field check into `read_json`, or scope the description to the binary format.

zkfriendly (approved):
Follow-ups would be to add optional zk. WHIR is already making this optional with an upcoming PR.

BornPsych:
> **minor:** Field-tag / version check is binary-only
> 
> The PR description says proofs are "tagged per field so one field's proof is rejected by another field's verifier," and that the 2.0 bump "fails fast." That's true on the binary path: `io/bin.rs` checks `FORMAT` and `major ==` when it parses the header. The JSON path doesn't. `read_json` (`io/json.rs:35`) is a bare `serde_json::from_reader` with no magic, version, or field check, and the `ensure!` calls in `verify()` are hash-based (`r1cs_hash`, `public_inputs_hash`), not version-based.
> 
> A cross-field or downgraded `.json` proof is caught only by the downstream transcript mismatch. That's still sound, since the domain separator serializes the whole scheme, but it isn't the explicit tag gate the description implies. Either thread the version and field check into `read_json`, or scope the description to the binary format.

Confirmed, and it applies to all four `FileFormat` types, not only the proof: `read_json` is a bare `serde_json::from_reader`.

Two things I checked. A 1.x JSON proof deserializes cleanly into 2.0, because `ProvekitProof`'s struct shape never changed between the versions. Only the bytes inside `narg_string` did, and that field is opaque hex, so any length decodes. Cross-field does get caught, but by accident: field elements hex-encode to different widths (32B for `Fr`, 8B for `Field64`, 24B for `Field64_3`) and the deserializer rejects short reads and trailing bytes. Hand it a circuit with zero public inputs and there is nothing left to catch it on.

Both still fail in transcript replay, so nothing unsound gets through. It just fails after the full verification run, with an error that points somewhere else.

Added a `warn!` on the JSON read path for now. since this pr is already reviewed and this fix would add a lot of changes in this PR. Tracked in #472.

tooling/provekit-fixtures/tests/shared/mod.rs
BornPsych: Good catch, Thanks, relocated to the `soundness_suite!`

provekit/common/src/whir_r1cs.rs
BornPsych: fixed this as well

shreyas-londhe (approved):
LGTM

PR #471, opened 2026-07-09
perf(prover): run the first sumcheck round in the base field

BornPsych:
## Summary

Runs the **first Spartan sumcheck round over the base field**. After the witness became base-committed (#470), the opening round still lifted the `a, b, c` mles into the extension *before* folding. This computes round 0's cubic evaluations directly on the base-field mles (one mixed base×ext multiply per point against the extension `eq`), then lifts into the extension only at the round-0 challenge. It also extracts the per-round message construction into a shared `combined_round_message` helper and adds the base-field sumcheck primitives it needs. A pair of `#[ignore]`d profiling probes are included so the win is reproducible in-tree.

Stacks on **#470** (base-field witness commitment). Independent of the CI guard (#469).

## Why

On the Goldilocks base-field path the witness mles are 64-bit `Field64`, but the sumcheck `eq` and challenges live in the degree-3 extension `Field64_3`. Lifting `a, b, c` to the extension before the first fold pays extension-width multiplies for data that is still base-valued. Doing round 0 in the base field removes one round's worth of extension multiplies — ~5% prove-time at 2^20 on the goldilocks base-field path. bn254 rides the identity embedding, so its path is a no-op (base == ext) and its proofs are unchanged.

## What changed

- **`provekit-common`** — `common/src/utils/sumcheck.rs` gains `mixed_sumcheck_map_reduce` and `mixed_fold`: reduce/fold the leading variable of base-field mles against an extension-field `eq`/challenge. Because the base→ext embedding `map` is a ring homomorphism, fold-then-lift is bit-identical to the old lift-then-fold — pinned by `test_mixed_map_reduce_matches_lifted` and `test_explicit_mixed_fold_matches_fused_fold`.
- **`provekit-prover`** — `prover/src/whir_r1cs.rs`: round 0 is hoisted out of the sumcheck loop and computed on the base-field mles (the loop now runs `1..m_0`); the per-round coefficient combination (`hhat + rho·g_poly`, saved-value equality) is extracted verbatim into `combined_round_message` and shared by round 0 and the loop.
- **`provekit-backend-bn254`** — `mavros_prove.rs` passes the embedding (`&Identity::new()`) and the base-field mles into the prover; identity embedding ⇒ no-op, transcript unchanged.
- **Profiling (`#[ignore]d`)** — `tooling/provekit-fixtures/tests/profile.rs`: a size sweep (BF vs EF vs bn254, 2^14..2^20) and a 2^20 goldilocks-BF flamegraph + stage breakdown. Timing excludes scheme construction and harness clones. Dev-deps `inferno` / `tracing-flame` are scoped to `provekit-fixtures` only.

## Transcript / soundness

The prover round messages, challenge draw order, and the separate ext blinding commitment are unchanged from #470 — the verifier is untouched and replays the identical Fiat-Shamir transcript. Round 0's base-field evaluations equal the old extension-lifted ones because `map` is a ring homomorphism (`map(a·b − c) = map(a)·map(b) − map(c)`, `map(0) = 0` for zero-padding). Both bn254 and goldilocks fixtures still prove→verify.

## Out of scope (tracked separately)

Same as #470: this does not touch the Go `recursive-verifier/` or the `provekit-gnark` bridge. The reported ~5% is from the local flamegraph probe, not a CI-enforced benchmark.

## Verification

- `cargo fmt --all --check`, `cargo clippy --all-targets --all-features`, `cargo build --all-targets --all-features` — all green, no new warnings.
- `cargo build --locked -p provekit-fixtures` — merged `Cargo.lock` self-consistent with the new `inferno`/`tracing-flame` pins.
- `provekit-common` 87 (incl. the base/ext parity tests), `provekit-r1cs-compiler` 41.
- Fixtures over **both** fields: bn254 7 roundtrip + 6 soundness, goldilocks 7 + 6 — all pass. Profiling probes are `#[ignore]d`.

## Commits

5 commits: profiling probes (ignored) · doc-comment rewrap · base-field sumcheck round helpers (`common`) · first sumcheck round in the base field (`prover` + bn254 call-site) · review follow-ups (hoist `1/2` inverse out of the per-round loop, move the profiling dev-deps to `[workspace.dependencies]`, dedup the single-commit timed prove into the fixtures harness).

### DIFF
--- a/provekit/prover/src/whir_r1cs.rs
+++ b/provekit/prover/src/whir_r1cs.rs
@@ -1,7 +1,7 @@
 use {
     ::tracing::instrument,
     anyhow::{ensure, Result},
-    ark_ff::Field,
+    ark_ff::{Field, One},
     ark_std::{
         rand::distributions::{Distribution, Standard},
         Zero,
@@ -525,6 +525,10 @@ where
     // Prove that sum of F + ρ·G over the boolean hypercube equals ρ·Σ(G).
     let mut saved_val_for_sumcheck_equality_assertion = rho * sum_g_reduce;
 
+    // 2 is invertible in any field of odd characteristic; precompute 1/2 once
+    // and reuse it for each round's cubic-coefficient solve.
+    let half = M::Target::one() / (M::Target::one() + M::Target::one());
+
     // First round: a, b, c are base-field, so the constraint products stay in
     // the base field with one mixed mul per point against the ext eq.
     let hhat = mixed_sumcheck_map_reduce([&a[..], &b[..], &c[..]], &eq, |[a, b, c], eq| {
@@ -543,6 +547,7 @@ where
         hhat,
         g_poly,
         rho,
+        half,
         saved_val_for_sumcheck_equality_assertion,
     );
     saved_val_for_sumcheck_equality_assertion = saved_val;
@@ -582,6 +587,7 @@ where
             hhat,
             g_poly,
             rho,
+            half,
             saved_val_for_sumcheck_equality_assertion,
         );
         saved_val_for_sumcheck_equality_assertion = saved_val;
@@ -606,6 +612,7 @@ fn combined_round_message<F: Field + Codec, S: DuplexSpongeInterface<U = u8>>(
     hhat: [F; 3],
     g_poly: [F; 4],
     rho: F,
+    half: F,
     saved_val_for_sumcheck_equality_assertion: F,
 ) -> (F, F) {
     let [hhat_i_at_0, hhat_i_at_em1, hhat_i_at_inf_over_x_cube] = hhat;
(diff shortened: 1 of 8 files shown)


shreyas-londhe (approved):
LGTM

provekit/prover/src/whir_r1cs.rs
```diff
@@ -495,44 +488,83 @@ pub fn pad_to_pow2_len_min2<F: Field>(v: &mut Vec<F>) {
...
+        g_poly,
+        rho,
+        half,
+        saved_val_for_sumcheck_equality_assertion,
+    );
+    saved_val_for_sumcheck_equality_assertion = saved_val;
+    alpha.push(alpha_0);
+
+    // Fold at the first challenge, lifting the mles into the extension.
+    let folded_a = mixed_fold(embedding, &a, alpha_0);
```
zkfriendly, line 557: `mixed_fold` breaks the prior fused fold/reduce path, adding four allocations and another scan. This is not a no-op for BN254 Identity. We can address this later since the focus is Goldilocks

zkfriendly (approved):
LGTM!

PR #485, opened 2026-08-20
feat: commit the R1CS witness with zook

BornPsych:
Switches the R1CS witness commitment from plain WHIR to zook (`RateSchedule::Adaptive`, `Mode::Standard`), and bumps whir to `78e9460`.

On the passport circuit this is ~37% off the R1CS proving core and ~17% off total prove time, for ~26% larger proofs.

SPARK query production works under zook: worldfnd/whir#274 makes zook's `prove` return its final evaluation point and RLC coefficients (`ProverClaim`), and the SPARK queries are assembled from it exactly as before the switch. Verified end-to-end (`prepare --spark` → `prove --produce-spark-query` → `prove-spark` → `verify-spark`) on a single-commitment circuit and a dual-commitment (range-check) circuit.

Notes:
- The whir pin is the head of worldfnd/whir#274 (still open); swap to a whir `main` rev once it merges.
- gnark parameter export and the Go recursive verifier are intentionally left disabled; they need a paired update.

### DIFF
--- a/provekit/common/src/whir_r1cs.rs
+++ b/provekit/common/src/whir_r1cs.rs
@@ -18,10 +18,39 @@ use {
     },
     serde::{Deserialize, Serialize},
     whir::{
-        algebra::embedding::Identity, engines::EngineId, parameters::ProtocolParameters,
-        protocols::whir::Config as GenericWhirConfig, transcript,
+        algebra::embedding::Identity,
+        engines::EngineId,
+        parameters::ProtocolParameters,
+        protocols::{
+            params::{
+                DecodingRegime, FoldingFactor, KneeWeight, Mode, PowBudget, RateSchedule,
+                SecuritySpec, TuningSpec,
+            },
+            whir::Config as GenericWhirConfig,
+            zook::ProtocolConfig as ZookConfig,
+        },
+        transcript,
     },
 };
+
+/// Zook mode for the witness commitment; [`Mode::ZeroKnowledge`] makes it
+/// hiding.
+const ZOOK_MODE: Mode = Mode::Standard;
+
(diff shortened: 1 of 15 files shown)


provekit/common/src/whir_r1cs.rs
```diff
@@ -18,11 +18,40 @@ use {
...
+            whir::Config as GenericWhirConfig,
+            zook::ProtocolConfig as ZookConfig,
+        },
+        transcript,
     },
 };
 
+/// Zook mode for the witness commitment; [`Mode::ZeroKnowledge`] makes it
+/// hiding.
+const ZOOK_MODE: Mode = Mode::Standard;
```
zkfriendly: Should we have `ZOOK_MODE` as a const or a dynamic as a cli flag or prover config? my question is do we need any other protocol change if we set `ZOOK_MODE = ZeroKnowledge`?

zkfriendly: Same question for the rate and security bits and other protocol params that whir takes as config

zkfriendly (changes requested):
Changes look good and correct. One thing I'd like to discuss briefly before merging is the required Zook configs. Right now, this PR hardcodes the config as consts, but Zook takes it as user-provided input; should we also take it as input? Maybe via the cli arguments, and we could have the default return the currently hardcoded values?

provekit/common/src/whir_r1cs.rs
BornPsych: good suggestion, I’ve made ZK the default, with `--witness-mode standard` as the opt-out for smaller non-hiding proofs. The config is already serialized into the scheme/domain separator, so the verifier follows whatever mode prepare used.

zkfriendly (approved):

PR #489, opened 2026-08-26
Bump Noir toolchain to v1.0.0-beta.26

BornPsych:
Incremental Noir toolchain upgrade from v1.0.0-beta.20 to v1.0.0-beta.26, one audited bump per version, using the published provekit_* beta.26 crates.


dcbuild3r (changes requested):
I found a compatibility issue: this PR removes `provekit-cli prepare --force` from `tooling/cli/src/cmd/prepare.rs`, but the same head still documents `--force` as a supported prepare option in README.md. Existing callers and users would fail argument parsing.

dcbuild3r (approved):
The compatibility issue is fixed by restoring `prepare --force`, and the refreshed stack includes the v1 Vercel-deployment suppression. Re-reviewed the latest head; approved pending fresh CI.

